PR #21, opened 2022-08-11
support heartbeat daemon over grpc based on inline proto

andymck:
TODO: 
1. grpc auth token support ( what is there atm doesnt work )
2. fix timestamp field
3. pull mobile proto from lib, remove inline copy


andymck:
legacy

PR #25, opened 2022-08-15
Add initial support for lora ingestor and verifier

andymck:
very much wip...both this code and my understanding of much of it


vihu:
This needs to add and handle location for hotspots, i.e. query the follower, get gateway_info and when outputting the files to the S3 bucket, attach location of the verified hotspot in `lora_witness_report_req_v1` and `lora_beacon_report_req_v1`. The protos for those also need to be updated accordingly.

store/src/file_store.rs
```diff
@@ -95,7 +95,7 @@ impl FileStore {
                                 None
                             }
                         })
-                        .filter(move |info| after.map_or(true, |v| info.timestamp >= v))
+                        .filter(move |info| after.map_or(true, |v| info.timestamp > v))
```
vihu, line 98: iirc, doing this messed up the mobile rewards file fetch at the time, is there a particular reason the `>=` is being removed here?

madninja: Yeah please do _not_ change semantics like that

vihu: I have reverted this back to `>=` in 749fc0e. We can get back to it at later if we need to.

vihu (approved):
LFG

PR #49, opened 2022-09-02
Mobile verifier 

maplant:
This is the initial draft for the mobile verifier. There is still quite a bit that needs to get done. Here are some of the items off the top of my head:  

- [x] Rip out everything in the rewards server that was moved into the verifier 
- [x] Have the rewards server pull from S3 
- [x] Merge in proto definitions for various structs into the proto repo 
- [x] Fix the server run loop which is broken right now 
- [x] Refactor shoddy error handling conventions - for example, using strings as the type for `invalid_reason`
- [x] Rebase onto master
- [ ] Add more tests


mobile_verifier/src/bin/verifier_server.rs
```diff
@@ -0,0 +1,23 @@
+use mobile_verifier::{Result, Server};
```
jeffgrunewald, line 1: rust "package layout" isn't clear to me on this: will this result in two separate executable binaries, each named `verifier_server` and `verifier_generate`?

maplant: Yeah. Any module in the `src/bin/` directory gets generated as a binary with the name of the module

madninja: please see my previous comment.. add to the main verifier as a command instead of making a new binary please. Same way the reward server does it

maplant: Ah, I misunderstood. Fixed.

mobile_verifier/src/reward_speed_share.rs
```diff
@@ -0,0 +1,327 @@
...
+            parse_dt("2022-08-01 0:00:00 +0000"),
+            0,
+            0,
+            0,
+            Validity::Valid,
+        );
+        let s1 = SpeedShare::new(
+            gw_public_key.clone(),
+            parse_dt("2022-08-01 6:00:00 +0000"),
+            bytes_per_s(25),
```
samgutentag, line 223:
```suggestion
            bytes_per_s(20),
```

to match the example in the comment above

mobile_verifier/src/reward_share.rs
```diff
@@ -0,0 +1,335 @@
...
+            timestamp: st_timestamp,
+            upload_speed: st_upload_speed,
+            download_speed: st_download_speed,
+            latency: st_latency,
+            ..
+        } = speedtest_req_v1;
+
+        if st_timestamp < after_utc || st_timestamp >= before_utc {
+            self.invalid_speed_shares.push(SpeedShare {
+                pub_key: PublicKey::try_from(st_pub_key.as_ref()).unwrap(),
```
madninja, line 244: why not create `From<SpeedtestReqV1> for SpeedtestShare`? That not possible?

maplant: Not really - we also need validity, so we'd have to have something like `fn from_speedtest_req_v1(req: SpeedtestReqV1, validity: Validity) -> Self`, which is probably cleaner

madninja: you could have a From that makes validity false by default and then set it in above function you define there?

maplant: I'm not a huge fan of that since it would lead to having code like: 
```rust
let mut share = SpeedShare::from(share);
share.validity = Validity::Invalid;
```
I'd like to come back and address this with a better, functional pattern.

maplant: I'll address this in the `FileStore` refactor I'm working on

maplant: Nevermind. Addressed

mobile_verifier/src/lib.rs
```diff
@@ -0,0 +1,54 @@
...
+
+pub use error::{Error, Result};
+pub use server::Server;
+
+use poc_store::FileStore;
+use rust_decimal::prelude::*;
+use rust_decimal_macros::dec;
+use std::io;
+
+pub async fn write_json<T: ?Sized + serde::Serialize>(
```
madninja, line 19: I do _not_ think pushing json to buckets is the right thing to do here. We're not going to support this json format publicly. We need proto messages to operate on instead of this approach

mobile_verifier/src/reward_share.rs
```diff
@@ -0,0 +1,297 @@
...
+                    timestamp: hb_timestamp,
+                    pub_key: hb_pub_key.to_vec(),
+                    weight: crate::bones_to_u64(cell_type.reward_weight()),
+                    cell_type: cell_type as i32,
+                    validity: ShareValidity::HeartbeatOutsideRange as i32,
+                });
+                return;
+            }
+
+            if let Ok(gw_pubkey) = PublicKey::try_from(hb_pub_key.as_ref()) {
```
madninja, line 261:
```suggestion
            if let Ok(gw_pubkey) = PublicKey::try_from(hb_pub_key.as_slice()) {
```

I've started using this

mobile_verifier/src/lib.rs
bbalser: where are we even using this?  I don't see it being called.

madninja: it's called a bunch of times in subnetwork_rewards afaict

bbalser: nevermind me, my language server was apparently very tired this morning and couldn't find the references

mobile_verifier/src/cli/generate.rs
```diff
@@ -0,0 +1,68 @@
...
+            input_bucket,
+            output_bucket,
+            dump_to_stdout,
+        } = self;
+
+        tracing::info!(
+            "Verifying shares from bucket {input_bucket} within the following time range: {after} to {before}"
+        );
+
+        let input_store = FileStore::new(None, "us-west-2", input_bucket).await?;
```
jeffgrunewald, line 40: that region feels like it should be the default value for an optional environment variable

jeffgrunewald: can the `FileStore` hide the implementation details of that so we only need to pass it the bucket name perhaps?

mobile_verifier/src/reward_share.rs
```diff
@@ -0,0 +1,297 @@
...
+                    ShareValidity::Valid,
+                );
+
+                if self
+                    .shares
+                    .get(&hb_cbsd_id)
+                    .map_or(false, |found_share| found_share.timestamp > hb_timestamp)
+                {
+                    return;
+                }
```
jeffgrunewald, line 276: do we want to continue to _only_ care about the most recent timestamp here if we're now requiring 2-3 heartbeats within the 24 hour period? do we need a count of valid heartbeats or is this just for reference?

mobile_verifier/src/reward_speed_share.rs
```diff
@@ -0,0 +1,342 @@
+//use crate::PublicKey;
```
jeffgrunewald, line 1:
```suggestion
```

mobile_verifier/src/server.rs
```diff
@@ -0,0 +1,76 @@
...
+
+        let mut last_reward_end_time =
+            DateTime::<Utc>::from_utc(NaiveDateTime::from_timestamp(last_reward_end_time, 0), Utc);
+        loop {
+            let (start, stop) = get_time_range(last_reward_end_time, lookup_delay);
+            SubnetworkRewards::from_period(&input_store, follower_client.clone(), start, stop)
+                .await?
+                .write(&output_store)
+                .await?;
+            last_reward_end_time = stop;
```
jeffgrunewald, line 57: this doesn't appear to be persisting anywhere, does that mean that if we want to preserve the time the server left off processing we need to track it outside of the server somewhere and inject it into the restart by way of the `LAST_REWARD_END_TIME` environment variable or else it'll restart processing from 1, 1, 1970?

mobile_verifier/src/lib.rs
maplant: Fixed

mobile_verifier/src/server.rs
maplant: Yes. I can add this value to the Meta db but I wasn't sure what the preferred way to track this is

mobile_verifier/src/subnetwork_rewards.rs
```diff
@@ -0,0 +1,687 @@
...
+
+        let cell_shares = cell_shares(&gathered_shares.shares);
+
+        let hotspot_shares = hotspot_shares(
+            &gathered_shares.shares,
+            &gathered_shares.speed_shares_moving_avg,
+        );
+
+        let (owner_shares, missing_owner_shares) = follower_service
+            .owner_shares(hotspot_shares.clone())
```
jeffgrunewald, line 79: why do hotspot shares need to be cloned here instead of just passing a reference to them for the duration of this calculation?

maplant: we need to write hotspot shares to s3 later

mobile_verifier/src/reward_share.rs
```diff
@@ -0,0 +1,297 @@
...
+    fn gather_speedtest(&mut self, speedtest_req: SpeedtestReqV1, after_utc: u64, before_utc: u64) {
+        // TODO: Better error handling for bad public keys
+        let mut speed_share = match SpeedShare::try_from(speedtest_req) {
+            Ok(share) => share,
+            Err(_) => return,
+        };
+
+        if speed_share.timestamp < after_utc || speed_share.timestamp >= before_utc {
+            speed_share.validity = ShareValidity::HeartbeatOutsideRange;
+            self.invalid_speed_shares.push(speed_share);
```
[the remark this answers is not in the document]
maplant: They're written as internal structs if further computation based on them is needed. Anything that's immediately written to a proto struct is not needed any more. 

I don't think this is ideal - for hygiene I think it would be better to write them both to internal structs and then convert them to the serializable formats in the write method. I did this because it was faster to write, but I'd be happy to clean it up

mobile_verifier/src/reward_share.rs
maplant: I don't think we _will_, but I think the plan was to address that issue after merging this PR. We will need a count of valid heartbeats

mobile_verifier/src/subnetwork_rewards.rs
jeffgrunewald: still learning the borrow checker but even though this is an async function (can't figure out where the `owner_shares` method is implemented on that follower client) but we're immediately awaiting on it and the _usage_ is effectively synchronous so is this just by virtue of the call being async that the compiler can't be sure a reference to `hotspot_shares` passed to the `owner_shares` method will return before the main function ends and we need to write it out to s3?
or for that matter the value's not being mutated in either case so couldn't we safely pass an immutable reference to both `owner_shares` and the s3 writer?
not saying this needs to change so much as embiggening my understanding

store/src/meta.rs
```diff
@@ -0,0 +1,127 @@
...
+            Some(str_val) => {
+                let value = str_val.parse().map_err(|_| MetaError::DecodeError)?;
+                Ok(Self {
+                    key: key.to_string(),
+                    value,
+                })
+            }
+            None => {
+                let value = default_fn();
+                sqlx::query(
```
bbalser, line 89: Can you use the MetaValue.insert function defined above to avoid have the query in a second place?

bbalser (approved):

jeffgrunewald (commented):
So we’re putting the meta store stuff in the `store` crate instead of its own? Is that mixing concerns since the store is supposed to be for files in s3?

maplant:
> So we’re putting the meta store stuff in the `store` crate instead of its own? Is that mixing concerns since the store is supposed to be for files in s3?

I think that it could go either way. Store just seems to imply online storage to me, so I thought it was a good place to put it. Otherwise two crates - file_store and meta_store would be appropriate

mobile_verifier/src/cli/generate.rs
```diff
@@ -0,0 +1,65 @@
...
+
+        /*
+        if let Some(output_bucket) = output_bucket {
+            todo!();
+            /*
+            let output_store = FileStore::new(None, "us-west-2", output_bucket).await?;
+            rewards.write(&output_store).await?;
+            */
+         }
+         */
```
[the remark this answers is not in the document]
maplant: nope

mobile_verifier/src/reward_share.rs
```diff
@@ -0,0 +1,312 @@
...
+                let emissions_per_share = actual_emissions / total_shares;
+                for (owner, share) in owner_shares {
+                    owner_emissions.insert(owner, Mobile::from(share * emissions_per_share));
+                }
+            }
+        }
+        OwnerEmissions(owner_emissions)
+    }
+
+    #[allow(dead_code)]
```
jeffgrunewald, line 72: This still dead?

maplant: It is not, but the function is only used in tests so this still needs to be here :-/

jeffgrunewald: Would it be better to just make it a public function or move it to a test-only function if it’s being called in multiple modules’ tests?

maplant: It is public, we'd have to make OwnerEmissions pub and reachable from the root of the module as well

PR #61, opened 2022-09-12
Simple env use, doc add

madninja:


status/src/server.rs
```diff
@@ -22,14 +23,10 @@ pub struct Server {
 
 impl Server {
     pub async fn from_env() -> Result<Self> {
-        let socket_addr = dotenv::var("API_SOCKET_ADDR").and_then(|v| {
```
jeffgrunewald, line 25: Still need that dotenv error in the errors module?

madninja: ah no we do not.. good catch.

madninja:
> Switching out dotenv because it doesn’t support a simple `unwrap_or_else` default value?

It's not great and using dotenv didn't really add anything

status/src/server.rs
madninja: hmm not quite.. `dotenv::dotenv()` still needs the Error represented in the crates

madninja: > per the docs, what does "load the .env file" mean? is that like sourcing the file into the runtime environment from within the application in case your terminal session or systemd unit or whatever doesn't do so or is it specific to the dotenv crate such that we don't need that function anymore either?

true.. though it's useful on the current server(s) since the env variables are _not_ set outside of the `.env` file which this function loads into the environment

store/src/file_upload.rs
```diff
@@ -38,17 +34,12 @@ impl FileUpload {
     pub async fn run(self, shutdown: &triggered::Listener) -> Result {
         tracing::info!("starting file uploader");
 
-        let enabled = self.enabled;
```
jeffgrunewald, line 41: some sort of testing gate i assume

jeffgrunewald (approved):

store/src/file_upload.rs
madninja: yeah, never really used it and just use an invalid bucket name to avoid uploads

PR #62, opened 2022-09-12
Added test to file_sink.rs

bbalser:


store/src/file_sink.rs
```diff
@@ -295,3 +295,64 @@ impl FileSink {
...
+    async fn writes_a_framed_gzip_encoded_file() {
+        let tmp_dir = tempdir::TempDir::new("entropy_tests").unwrap();
+        let (shutdown_trigger, shutdown_listener) = triggered::trigger();
+        let (sender, receiver) = message_channel(10);
+
+        let mut file_sink = FileSinkBuilder::new(FileType::Entropy, &tmp_dir.path(), receiver)
+            .roll_time(chrono::Duration::milliseconds(100))
+            .create()
+            .await
+            .unwrap();
```
jeffgrunewald, line 317:
```suggestion
            .await?;
```

store/src/file_sink.rs
```diff
@@ -295,3 +295,64 @@ impl FileSink {
...
+
+        let sink_thread = tokio::spawn(async move {
+            let _ = file_sink.run(&shutdown_listener).await;
+        });
+
+        let _ = sender
+            .send(String::into_bytes("hello".to_string()))
+            .await
+            .unwrap();
```
jeffgrunewald, line 326:
```suggestion
            .await?;
```

store/src/file_sink.rs
```diff
@@ -295,3 +295,64 @@ impl FileSink {
...
+
+        std::thread::sleep(time::Duration::from_millis(200));
+
+        shutdown_trigger.trigger();
+        let _ = sink_thread.await;
+
+        let entropy_file = get_entropy_file(&tmp_dir).await.unwrap();
```
jeffgrunewald, line 333:
```suggestion
        let entropy_file = get_entropy_file(&tmp_dir).await?;
```

store/src/file_sink.rs
```diff
@@ -295,3 +295,64 @@ impl FileSink {
...
+
+        let g = GzipDecoder::new(BufReader::new(
+            File::open(entropy_file.path()).await.unwrap(),
```
jeffgrunewald, line 336:
```suggestion
            File::open(entropy_file.path()).await?,
```

store/src/file_sink.rs
```diff
@@ -295,3 +295,64 @@ impl FileSink {
...
+        ));
+        let mut fr = FramedRead::new(g, LengthDelimitedCodec::new());
+
+        assert_eq!("hello", fr.next().await.unwrap().unwrap());
+    }
+
+    async fn get_entropy_file(tmp_dir: &tempdir::TempDir) -> Option<DirEntry> {
+        let mut entries = fs::read_dir(tmp_dir.path()).await.unwrap();
```
jeffgrunewald, line 344:
```suggestion
        let mut entries = fs::read_dir(tmp_dir.path()).await?;
```

store/src/file_sink.rs
```diff
@@ -295,3 +295,64 @@ impl FileSink {
...
+
+        while let Some(entry) = entries.next_entry().await.unwrap() {
```
jeffgrunewald, line 346:
```suggestion
        while let Some(entry) = entries.next_entry().await? {
```

store/src/file_sink.rs
```diff
@@ -295,3 +295,64 @@ impl FileSink {
...
+            if is_entropy_file(&entry) {
+                return Some(entry);
+            }
+        }
+
+        None
+    }
+
+    fn is_entropy_file(entry: &DirEntry) -> bool {
+        entry.file_name().to_str().unwrap().starts_with("entropy.")
```
jeffgrunewald, line 356:
```suggestion
        entry.file_name().to_str()?.starts_with("entropy.")
```

jeffgrunewald (commented):
i think in the test case you can safely use the shorthand `?` anywhere you would otherwise use unwrap

store/src/file_sink.rs
madninja: You can't do the question mark bit here since a test does not return a result. The convention is to use `.expect("could not create sink")` to make the failure clear

store/src/file_sink.rs
```diff
@@ -295,3 +295,64 @@ impl FileSink {
...
+        let sink_thread = tokio::spawn(async move {
+            let _ = file_sink.run(&shutdown_listener).await;
+        });
+
+        let _ = sender
+            .send(String::into_bytes("hello".to_string()))
+            .await
+            .unwrap();
+
+        std::thread::sleep(time::Duration::from_millis(200));
```
madninja, line 328: Do not std::thread::sleep but use `tokio::time::sleep` with `.await`. async systems don't like you blocking the working thread

store/src/file_sink.rs
jeffgrunewald: we worked out that a test is similar to `main` in that it doesn't return a result by default; i sort of understand how that works in the sense of a test being a standalone execution of the code in question but on the other hand it seems so natural to me that a test should return a `Result`

store/src/file_sink.rs
```diff
@@ -295,3 +295,64 @@ impl FileSink {
...
+        let (sender, receiver) = message_channel(10);
+
+        let mut file_sink = FileSinkBuilder::new(FileType::Entropy, &tmp_dir.path(), receiver)
+            .roll_time(chrono::Duration::milliseconds(100))
+            .create()
+            .await
+            .unwrap();
+
+        let sink_thread = tokio::spawn(async move {
+            let _ = file_sink.run(&shutdown_listener).await;
```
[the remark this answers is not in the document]
bbalser: Addressed in dc8b27e

store/src/file_sink.rs
bbalser: Addressed in dc8b27e

store/src/file_sink.rs
bbalser: Fixed in dc8b27e

jeffgrunewald (approved):

store/src/file_sink.rs
```diff
@@ -295,3 +295,70 @@ impl FileSink {
...
+            if is_entropy_file(&entry) {
+                return entry;
+            }
+        }
+
+        panic!("no entropy file available")
+    }
+
+    fn is_entropy_file(entry: &DirEntry) -> bool {
+        entry.file_name().to_str().unwrap().starts_with("entropy.")
```
madninja, line 362: use `FileType` to parse and then compare with `FileType::Entropy`

store/Cargo.toml
```diff
@@ -37,4 +37,5 @@ strum_macros = "0"
 
 [dev-dependencies]
 base64 = "0"
+tempdir = "0.3.7"
```
madninja, line 40: we usually don't nail down this specifically.. but to the major version in this case so.. change to "0"

store/src/file_sink.rs
```diff
@@ -295,3 +295,70 @@ impl FileSink {
...
+
+        shutdown_trigger.trigger();
+        sink_thread.await.expect("file sink did not complete");
+
+        let entropy_file = get_entropy_file(&tmp_dir).await;
+        assert_eq!("hello", read_file(&entropy_file).await);
+    }
+
+    async fn read_file(entry: &DirEntry) -> bytes::BytesMut {
+        let gzip_decoder =
```
madninja, line 340: you can't use file_source here? you're doing the same thing

store/src/file_sink.rs
bbalser: Addressed in 598a7d7

store/Cargo.toml
bbalser: Address in 598a7d7

store/src/file_sink.rs
bbalser: Fixed in 598a7d7

much easier, thank you

madninja (approved):

PR #64, opened 2022-09-13
Convert mobile_rewards to use mobile_verifier

maplant:


mobile_rewards/src/server.rs
```diff
@@ -371,7 +397,92 @@ impl Server {
...
+        reward_server_signature: vec![],
+    };
+    txn.reward_server_signature = sign_txn(&txn, keypair)?;
+    let hash = hash_txn_b64_url(&txn);
+    Ok((txn, hash))
+}
+
+use sha2::{Digest, Sha256};
+
+fn hash_txn_b64_url(txn: &BlockchainTxnSubnetworkRewardsV1) -> String {
```
vihu, line 477: There's a trait in `txn_hash.rs` to do this

mobile_rewards/src/server.rs
```diff
@@ -371,7 +397,92 @@ impl Server {
...
+    let mut txn = txn.clone();
+    txn.reward_server_signature = vec![];
+    let digest = Sha256::digest(&txn.encode_to_vec()).to_vec();
+    base64::encode_config(&digest, base64::URL_SAFE_NO_PAD)
+}
+
+fn sign_txn(txn: &BlockchainTxnSubnetworkRewardsV1, keypair: &Keypair) -> Result<Vec<u8>> {
```
vihu, line 484: Same thing for this, should be able to call `txn.sign()` once getting the `TxnSign` trait.

mobile_rewards/src/server.rs
maplant: Yeah, I got rid of that and signed and combined them into this function. The reasoning being that a Trait that is only implemented for one type should just be a function. 

Additionally, we don't have any function that take `impl Sign` or whatever, so they're not great traits to have around

vihu: I see. Well since at least injector will also be signing transactions (and at some point so will iot-rewards server), we should probably move those traits to poc-common (or some such thing) imo.

Or perhaps, maybe slightly better would be to use the trait directly from `helium-crypto` (I believe the traits we had were pretty much doing the same thing that was happening in helium-crypto with some limited functionality)

mobile_rewards/src/server.rs
```diff
@@ -371,7 +397,92 @@ impl Server {
...
+        txn_hash: txn_hash.to_vec(),
+        txn_types: txn_types.iter().map(|e| e.to_string()).collect(),
+    };
+    let res = client.txn_stream(req).await?.into_inner();
+    Ok(res)
+}
+
+async fn reward_period(client: &mut follower::Client<Channel>) -> Result<Range<u64>> {
+    let req = FollowerSubnetworkLastRewardHeightReqV1 {
+        token_type: BlockchainTokenTypeV1::from(ProtoTokenType::Mobile).into(),
```
jeffgrunewald, line 440:
```suggestion
        token_type: ProtoTokenType::Mobile,
```
i'm assuming this is what clippy is trying to tell us not to do here unless i'm/it's getting confused about what we're trying to `into()` into?

mobile_rewards/src/server.rs
```diff
@@ -371,7 +397,92 @@ impl Server {
...
+
+pub fn construct_txn(
+    keypair: &Keypair,
+    rewards: Vec<SubnetworkReward>,
+    period: Range<u64>,
+) -> Result<(BlockchainTxnSubnetworkRewardsV1, String)> {
+    let mut txn = BlockchainTxnSubnetworkRewardsV1 {
+        rewards: rewards,
+        token_type: BlockchainTokenTypeV1::from(ProtoTokenType::Mobile).into(),
+        start_epoch: period.start,
```
jeffgrunewald, line 466: interesting; didn't know you could treat a range like this

mobile_rewards/src/server.rs
```diff
@@ -371,7 +397,92 @@ impl Server {
...
+        end_epoch: period.end,
+        reward_server_signature: vec![],
+    };
+    txn.reward_server_signature = sign_txn(&txn, keypair)?;
+    let hash = hash_txn_b64_url(&txn);
+    Ok((txn, hash))
+}
+
+use sha2::{Digest, Sha256};
```
jeffgrunewald, line 475: since this isn't scoped within a submodule or a function is there any conventional guideline around keeping all the use statements grouped together at the top of the file?

maplant: I would not recommend doing this, I just through rustfmt would put it at the top of the file. 
If this sort of thing makes sense then I think that it probably makes more sense to split up into another module

mobile_rewards/src/server.rs
```diff
@@ -1,34 +1,42 @@
...
+        follower::{
+            self, FollowerSubnetworkLastRewardHeightReqV1, FollowerTxnStreamReqV1,
+            FollowerTxnStreamRespV1,
+        },
+        poc_mobile::OwnerEmissions,
         transaction::{TxnQueryRespV1, TxnStatus as ProtoTxnStatus},
+        Channel, Endpoint,
     },
-    BlockchainTokenTypeV1, BlockchainTxn, BlockchainTxnSubnetworkRewardsV1, Message,
+    BlockchainTokenTypeV1, BlockchainTokenTypeV1 as ProtoTokenType, BlockchainTxn,
```
jeffgrunewald, line 25:
```suggestion
   BlockchainTokenTypeV1 as ProtoTokenType, BlockchainTxn,
```
looks like a redundant import of token type

mobile_rewards/src/server.rs
```diff
@@ -1,34 +1,42 @@
...
+use poc_store::{FileStore, FileType};
 use sqlx::{Pool, Postgres};
 use std::env;
+use std::{ops::Range, time::Duration as StdDuration};
 use tokio::time;
 use tonic::Streaming;
 
 pub const DEFAULT_START_REWARD_BLOCK: i64 = 1477650;
+/// default minutes to delay lookup from now
+pub const DEFAULT_LOOKUP_DELAY: i64 = 30;
```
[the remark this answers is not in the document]
madninja: Ideally we'd just use `Duration`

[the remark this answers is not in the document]
jeffgrunewald: is a duration something you can derive at compile time or do you need something like lazy_static to handle that?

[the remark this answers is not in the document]
madninja: > is a duration something you can derive at compile time or do you need something like lazy_static to handle that?

You sure can. For example: https://github.com/helium/oracles/blob/main/mobile_rewards/src/follower.rs#L20

mobile_rewards/src/server.rs
```diff
@@ -283,23 +293,39 @@ impl Server {
 
     async fn handle_rewards(
         &mut self,
-        reward_period: RewardPeriod,
-        last_reward_time: i64,
-        end_utc: i64,
+        reward_period: Range<u64>,
+        start_utc: DateTime<Utc>,
+        end_utc: DateTime<Utc>,
```
jeffgrunewald, line 298: this is going to introduce a bug when you persist this value back to the db by updating the Meta `last_reward_end_time`; you're calling `end_utc.to_string()` to update that value at the end of this function which is going to be something like `"2022-09-15 15:04:33.074973664 UTC"` so the next loop of the clock tick function it'll pass this string to your `get_time_range` function which is only accepting an i64 unix timestamp

mobile_rewards/src/server.rs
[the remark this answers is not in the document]
madninja: True. which prolly why it was simpler to do it this way. I don't mind either way but the name should indeed include the unit if it's a bare number

mobile_rewards/src/server.rs
```diff
@@ -283,23 +293,39 @@ impl Server {
...
+            .verifier_store
+            .source(stream::iter(file_list).map(Ok).boxed());
+        let mut rewards = vec![];
+        while let Some(Ok(msg)) = stream.next().await {
+            match OwnerEmissions::decode(msg) {
+                Ok(OwnerEmissions { emissions }) => {
+                    for emission in emissions.into_iter() {
+                        rewards.push(SubnetworkReward {
+                            account: emission.pub_key,
+                            amount: emission.weight,
```
jeffgrunewald, line 318: is this the correct value that should be pushed into the reward? either the name `weight` is misleading or this is the weighted share of a given amount of rewards, not the actual rewards themselves. i believe this needs to have maths done on it to determine how many mobile bones of the total for the reward period should be allocated to each owner pubkey based on the weight amount. the mobile bones integer is what should be setting the reward amount value

maplant: I think I need to remove owner emissions and just use subnetwork rewards.  I probably just named the field incorrectly 

```rust
impl From<OwnerEmissions> for SubnetworkRewards {
    fn from(owner_emissions: OwnerEmissions) -> SubnetworkRewards {
        let unsorted_rewards = owner_emissions
            .0
            .into_iter()
            .map(|(owner, amt)| ProtoSubnetworkReward {
                account: owner.to_vec(),
                amount: u64::from(amt),
            })
            .collect();
        SubnetworkRewards(unsorted_rewards)
    }
}
```

maplant: https://github.com/helium/proto/pull/178

mobile_rewards/src/server.rs
maplant: oooooooh nice catch, I'm going to re-work meta to fix issues like this

madninja (commented):
Very nice! Just a nit-question

In addition: should meta live as a sibling of the mobile_verifier now that it's generic enough to be used by the other crates that use metadata (like injectors and verifiers)?

mobile_rewards/src/server.rs
```diff
@@ -324,10 +321,13 @@ impl Server {
         }
 
         if !rewards.is_empty() {
-            let pool = &self.pool.clone();
-            let issue_rewards = self.issue_rewards(rewards, reward_period);
-            let update_end_time = Meta::update(pool, "last_reward_end_time", end_utc.to_string());
-            try_join!(issue_rewards, update_end_time)?;
+            let pool = self.pool.clone();
+            let (issue_rewards, update_end_time) = join!(
```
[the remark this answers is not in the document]
maplant: Because it uses a separate, more specific error enum join can't resolve them this way :-/

maplant:
> Very nice! Just a nit-question
> 
> In addition: should meta live as a sibling of the mobile_verifier now that it's generic enough to be used by the other crates that use metadata (like injectors and verifiers)?

I think that is a good idea, maybe give it a better name than `Meta` too (although to be honest, a better name escapes me)

madninja:
> > Very nice! Just a nit-question
> > In addition: should meta live as a sibling of the mobile_verifier now that it's generic enough to be used by the other crates that use metadata (like injectors and verifiers)?
> 
> I think that is a good idea, maybe give it a better name than `Meta` too (although to be honest, a better name escapes me)



> > Very nice! Just a nit-question
> > In addition: should meta live as a sibling of the mobile_verifier now that it's generic enough to be used by the other crates that use metadata (like injectors and verifiers)?
> 
> I think that is a good idea, maybe give it a better name than `Meta` too (although to be honest, a better name escapes me)

`meta_store`?

mobile_rewards/src/server.rs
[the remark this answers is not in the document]
madninja: so why not two separate serial calls then?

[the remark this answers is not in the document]
maplant: That's a good point. I was going to change it to serial but figured that this way was intentional and left it. I'll remove the join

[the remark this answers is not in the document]
jeffgrunewald: when it was `try_join!` and then `?` the result it wouldn't succeed unless they both did, correct?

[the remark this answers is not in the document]
maplant: The function would return an error, but the update_end_time would still get pushed to the database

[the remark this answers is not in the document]
maplant: (unless the future got cancelled before)

mobile_rewards/src/server.rs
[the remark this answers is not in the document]
jeffgrunewald: looks like you _can_ compute a `std::time::Duration` in a const but _not_ a `chrono::Duration` fwiw

mobile_rewards/src/server.rs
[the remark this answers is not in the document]
vihu: Well, I suppose then it is incorrect right now and should probably happen in sequence, i.e. only update last_reward_end_time if we were able to issue rewards.

mobile_rewards/src/server.rs
```diff
@@ -371,7 +398,93 @@ impl Server {
...
+        txn_hash: txn_hash.to_vec(),
+        txn_types: txn_types.iter().map(|e| e.to_string()).collect(),
+    };
+    let res = client.txn_stream(req).await?.into_inner();
+    Ok(res)
+}
+
+async fn reward_period(client: &mut follower::Client<Channel>) -> Result<Range<u64>> {
+    let req = FollowerSubnetworkLastRewardHeightReqV1 {
+        token_type: ProtoTokenType::Mobile,
```
jeffgrunewald, line 444:
```suggestion
        token_type: ProtoTokenType::Mobile.into(),
```

mobile_rewards/src/server.rs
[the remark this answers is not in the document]
madninja: concur.. ok @maplant let's get that fixed up then. Glad we caught it

file_store/src/cli/bucket.rs
```diff
@@ -217,7 +217,6 @@ fn locate(
         FileType::LoraValidPoc => {
             LoraValidPoc::decode(buf).and_then(|event| event.to_value_if(gateway))
         }
-        FileType::EntropyReport => Ok(None),
```
jeffgrunewald, line 220: do we no longer need to report these for validation?

madninja: Fair enough.. I was trying to be explicit, but it's superfluous

mobile_rewards/src/server.rs
```diff
@@ -371,7 +391,93 @@ impl Server {
...
 }
 
+const CONNECT_TIMEOUT: StdDuration = StdDuration::from_secs(5);
+const RPC_TIMEOUT: StdDuration = StdDuration::from_secs(5);
+const DEFAULT_URI: &str = "http://127.0.0.1:8080";
+
 async fn update_last_follower(pool: &Pool<Postgres>, last_height: u64) -> Result {
     metrics::gauge!("reward_server_last_reward_height", last_height as f64);
-    Meta::update(pool, "last_follower_height", last_height.to_string()).await
+    MetaValue::new("last_follower_height", last_height)
```
vihu, line 400: Is this call correct? Or should it be calling `MetaValue::update(last_height)`?

maplant: We call insert right after, so yes this is correct

vihu: I mean logically, are we inserting a new k:v pair for `last_follower_height:value` or are we updating an existing `last_follower_height` value here. The function name seems to imply that it should be an update?

maplant: This inserts if the value does not exist and updates when it does, so this is what we want.

maplant: I've fixed this so that `last_follower_height` is always a `MetaValue`, so when we want to change its value we just call `.update`

vihu: Thanks, I think that's much cleaner :)

bbalser (approved):

jeffgrunewald (approved):

PR #67, opened 2022-09-14
Add boilerplate for poc-injector

vihu:
# Summary
- This just adds boilerplate code for poc-injector server.
- The main thing to look at is the `server.rs` module.

# Notes:
- Start server, operating at some `POC_TICK_TIME` (set to 30 mins as a default constant).
- Add deb packaging stuff
- This doesn't really do any business logic, that will be in a separate PR, but I and @bbalser think this is harmless to actually land on main.
- Once we agree that this is indeed the kind of behavior we want for the poc-injector server to have, I'll start working on the business logic, in a hopefully smaller PR.

# Caveats:
- I added `follower` and `keypair` here as well, but we really should have `poc-common` or some such to move those things to. At least the reward-server (both iot and mobile) will use that for txn signing and submission.

Fixes: https://github.com/helium/oracles/issues/66


madninja (commented):
Rename to something with iot in the name. Like poc_iot_injector to make it clear where it belongs

Cargo.toml
```diff
@@ -1,2 +1,3 @@
 [workspace]
-members = [ "ingest", "store", "mobile_rewards", "status", "entropy"]
+members = [ "ingest", "store", "mobile_rewards", "status", "entropy",
+"poc_iot_injector"]
```
jeffgrunewald, line 3: just iot_injector? poc_ seems redundant

madninja: well.. there'll also be a "packet injector" which injects state channels ?

madninja (approved):

Issue #70, opened 2022-09-14
Update IOT Injector to process output of IOT verifier and end blockchain poc receipts v2 txns to helium chain


vihu:
Todo:
- [x] Merge https://github.com/helium/oracles/pull/67
- [x] Add database handling for figuring out which next files to look at from the iot-verifier
- [x] Query iot-verifier S3 files
- [x] Construct `blockchain_txn_poc_receipts_v2` transaction using iot-verifier S3 files
- [x] Sign transaction using `iot_injector_keypair`
- [x] Submit transaction to txn_service

PR #71, opened 2022-09-14
Add poc-injector working logic

vihu:
# Summary
- Actual logic for the poc-iot-injector
- Fix: https://github.com/helium/oracles/issues/70

# Todo:
- [x] Merge https://github.com/helium/oracles/pull/67
- [x] Add database handling for figuring out which next files to look at from the iot-verifier
- [x] Query iot-verifier S3 files
- [x] Construct `blockchain_txn_poc_receipts_v2` transaction using iot-verifier S3 files
- [x] Sign transaction using `iot_injector_keypair`
- [x] Submit transaction to follower
- [x] Do reward_shares calculation using hex density scale
- [x] Merge https://github.com/helium/oracles/pull/25
- [x] Merge https://github.com/helium/proto/pull/173
- [x] Test receipt txn generation using `generate` cli and verify that the aux ledger can absorb the txn
- [ ] More exhaustive end-to-end testing with a running chain


poc_iot_injector/src/txn_service.rs
```diff
@@ -0,0 +1,49 @@
...
+        let channel = Endpoint::from(uri)
+            .connect_timeout(CONNECT_TIMEOUT)
+            .timeout(RPC_TIMEOUT)
+            .connect_lazy();
+        Ok(Self {
+            client: TransactionClient::new(channel),
+        })
+    }
+
+    pub async fn submit(&mut self, txn: BlockchainTxn, key: &[u8]) -> Result<TxnSubmitRespV1> {
```
jeffgrunewald, line 41: does the injector service not need to query the status of submitted transactions?

vihu: I think at least initially, we'll just do fire and forget. There isn't much we can do anyway if a receipt submission/acceptance fails tbh.

poc_iot_injector/src/cli/generate.rs
```diff
@@ -0,0 +1,49 @@
...
+
+        while let Some(Ok(msg)) = stream.next().await {
+            if let Ok(Some((txn, _hash, _hash_b64_url))) =
+                handle_report_msg(msg, &poc_oracle_key, ts)
+            {
+                let tx = BlockchainTxn {
+                    txn: Some(Txn::PocReceiptsV2(txn)),
+                };
+
+                println!("txn_bin: {:?}", tx.encode_to_vec());
```
madninja, line 44: use tracing::debug not println

vihu: Updated 6bf605f

poc_iot_injector/src/cli/generate.rs
```diff
@@ -0,0 +1,49 @@
...
+    /// Required start time to look for (inclusive)
+    #[clap(long)]
+    after: NaiveDateTime,
+    /// Required before time to look for (inclusive)
+    #[clap(long)]
+    before: NaiveDateTime,
+}
+
+impl Cmd {
+    pub async fn run(&self) -> Result {
```
jeffgrunewald, line 19: the mobile verifier PR raised the question: should we also be running our sql migrations in the generate cli functions to be on the safe side or does this not touch the DB so that's unnecessary?

poc_iot_injector/src/receipt_txn.rs
```diff
@@ -0,0 +1,244 @@
...
+use rust_decimal_macros::dec;
+use sha2::{Digest, Sha256};
+
+// TODO: These should ideally be coming in from some configuration service
+const WITNESS_REDUNDANCY: u32 = 4;
+const POC_REWARD_DECAY_RATE: Decimal = dec!(0.8);
+const HIP15_TX_REWARD_UNIT_CAP: Decimal = dec!(2.0);
+
+// TODO: This should be done better?
+const MHZTOHZ: u32 = 1000000;
```
jeffgrunewald, line 21: perhaps a macro?

poc_iot_injector/src/receipt_txn.rs
```diff
@@ -0,0 +1,244 @@
...
+
+type PocPath = Vec<BlockchainPocPathElementV1>;
+
+pub fn handle_report_msg(
+    msg: prost::bytes::BytesMut,
+    keypair: &Keypair,
+    timestamp: i64,
+) -> Result<Option<(BlockchainTxnPocReceiptsV2, Vec<u8>, String)>> {
+    // NOTE: This is intentionally not Vec::with_capacity(1) since we _may_ add multi-hop PocPath
```
jeffgrunewald, line 27: couldn't we just convert this to `vec![]` if and when we do that? presumably we'd need more code changes than that and it wouldn't be as simple as toggling configuration?

poc_iot_injector/src/cli/generate.rs
madninja: I don't think we should have those side effects in a cli.. Intuitively I'd expect it to be set up and error out if anything is wrong

vihu: This doesn't touch the DB. Just looks at before/after ts and queries the verified bucket on S3 for valid_poc files.

poc_iot_injector/src/receipt_txn.rs
```diff
@@ -0,0 +1,244 @@
...
+
+    let LoraValidPocV1 {
+        poc_id: _poc_id,
+        beacon_report,
+        witness_reports,
+    } = LoraValidPocV1::decode(msg)?;
+
+    let poc_witnesses = construct_poc_witnesses(witness_reports)?;
+
+    // We expect poc_receipt to always be here, once it is in a report
```
jeffgrunewald, line 39: where do we always expect the receipt? we're always expecting construct_poc_receipt to return `Some(poc_receipt)` given valid beacon/witness reports? if so should we log an error in the `else` block to better catch if this invariant is ever violated?

poc_iot_injector/src/receipt_txn.rs
madninja: use a function .. I think you have one already.. have the multiplier there and remove this const

poc_iot_injector/src/receipt_txn.rs
```diff
@@ -0,0 +1,244 @@
...
+    if let Some(poc_receipt) = construct_poc_receipt(beacon_report, poc_witnesses.len() as u32)? {
+        // TODO: Double check whether the gateway in the poc_receipt is challengee?
```
jeffgrunewald, line 41: is this a relevant condition? will the off-chain poc-injector be processing validator challenged beacons or will it always be processing self-beaconed reports?

poc_iot_injector/src/server.rs
```diff
@@ -1,28 +1,46 @@
...
+use futures::stream::{self, StreamExt};
+use helium_proto::{blockchain_txn::Txn, BlockchainTxn};
+use sqlx::{Pool, Postgres};
+use std::env;
 use tokio::time;
 
 // 30 mins
 pub const POC_IOT_TICK_TIME: time::Duration = time::Duration::from_secs(1800);
+// TODO: decide some default value, not that it really matters if we set it in the .env
+pub const DEFAULT_LAST_POC_SUBMISSION_TS: i64 = 0000000000;
```
madninja, line 17: I'm still not clear what units this is in? Seconds since epoch?

poc_iot_injector/src/receipt_txn.rs
```diff
@@ -0,0 +1,244 @@
...
+
+fn construct_poc_witnesses(
+    witness_reports: Vec<LoraValidWitnessReportV1>,
+) -> Result<Vec<BlockchainPocWitnessV1>> {
+    let num_witnesses = witness_reports.len();
+    let mut poc_witnesses: Vec<BlockchainPocWitnessV1> = Vec::with_capacity(num_witnesses);
+    for witness_report in witness_reports {
+        let LoraValidWitnessReportV1 {
+            received_timestamp: _witness_received_timestamp,
+            location: _witness_location,
```
jeffgrunewald, line 95: do you need these if you're not using them? can you not match on only the fields you care about and ignore the rest with `..`?

poc_iot_injector/src/receipt_txn.rs
```diff
@@ -0,0 +1,244 @@
...
+}
+
+fn sign_txn(txn: &BlockchainTxnPocReceiptsV2, keypair: &Keypair) -> Result<Vec<u8>> {
+    let mut txn = txn.clone();
+    txn.signature = vec![];
+    Ok(keypair.sign(&txn.encode_to_vec())?)
+}
+
+fn poc_challengee_reward_unit(num_witnesses: u32) -> Result<Decimal> {
+    if num_witnesses == 0 {
```
jeffgrunewald, line 218: shouldn't the verifier be marking any poc reports with 0 witnesses as invalid already?

poc_iot_injector/src/receipt_txn.rs
vihu: Updated 49d15fd

poc_iot_injector/src/server.rs
vihu: Updated 49d15fd

poc_iot_injector/src/receipt_txn.rs
vihu: It _is_ `vec![]`

poc_iot_injector/src/receipt_txn.rs
vihu: right, forgot about that. Updated 9cf0023

poc_iot_injector/src/receipt_txn.rs
vihu: I'm not sure about that tbh. I will revisit this once I have confirmation.

poc_iot_injector/src/receipt_txn.rs
jeffgrunewald: i meant why not make it `Vec::with_capacity(1)` now and convert it to a `vec![]` if and when we do enable multi-hop paths

vihu: Updated 16cbb24

jeffgrunewald (approved):

poc_iot_injector/src/receipt_txn.rs
madninja: This should be addressed in a subsequent PR.. Can we add a note here?

vihu: Added a todo 13c8771

madninja (approved):

PR #73, opened 2022-09-16
make BUCKET_ENDPOINT optional

madninja:


store/src/file_store.rs
```diff
@@ -17,12 +17,13 @@ pub struct FileStore {
...
 impl FileStore {
     pub async fn from_env() -> Result<Self> {
-        let endpoint_env = env::var("BUCKET_ENDPOINT")?;
-        let endpoint: Option<Endpoint> = Uri::from_str(&endpoint_env)
-            .map(Endpoint::immutable)
-            .map(Some)
-            .map_err(DecodeError::from)?;
-
+        let endpoint: Option<Endpoint> = match env::var("BUCKET_ENDPOINT") {
+            Ok(endpoint_env) => Uri::from_str(&endpoint_env)
```
maplant, line 21: This could probably be done with combinators instead of a match but I think it's fine either way

maplant (approved):

store/src/file_store.rs
madninja: I know.. this was just easier to read at the time I wrote it

PR #74, opened 2022-09-19
Change ingestor to wrap request in report with received timestamps in milliseconds

bbalser:


madninja (commented):
Good start.. The problem is that the FileInfo/FileType will fail reading the resulting files since they expect the original `_req`s and this is a new message type. We'd need to modify FileType as well, which would imply a new filename prefix, which then means co-ordinating a switchover in production

store/src/heartbeat.rs
```diff
@@ -57,3 +77,48 @@ impl From<CellHeartbeat> for CellHeartbeatReqV1 {
...
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use prost::Message;
+
+    fn bytes() -> Vec<u8> {
+        hex::decode("008f23e96ab6bbff48c8923cac831dc97111bcf33dba9f5a8539c00f9d93551af1").unwrap()
```
madninja, line 87: Try using `https://docs.rs/hex-literal/0.3.4/hex_literal/` instead.. it allows for const hex declarations

store/src/speedtest.rs
```diff
@@ -1,7 +1,7 @@
 use crate::{datetime_from_epoch, traits::MsgDecode, Error, Result};
-use chrono::{DateTime, Utc};
+use chrono::{DateTime, TimeZone, Utc};
```
madninja, line 2: Same question

store/src/heartbeat.rs
```diff
@@ -1,7 +1,7 @@
 use crate::{datetime_from_epoch, traits::MsgDecode, Error, Result};
-use chrono::{DateTime, Utc};
+use chrono::{DateTime, TimeZone, Utc};
```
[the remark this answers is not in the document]
madninja: Ah yes.. I missed that in the Timezone docs

store/src/heartbeat.rs
bbalser: Fixed fe126ca

store/src/cli/info.rs
```diff
@@ -67,6 +69,10 @@ fn get_timestamp(file_type: &FileType, buf: &[u8]) -> Result<DateTime<Utc>> {
         FileType::CellSpeedtest => {
             SpeedtestReqV1::decode(buf).map(|entry| datetime_from_epoch(entry.timestamp))?
         }
+        FileType::CellHeartbeatIngestReport => CellHeartbeatIngestReportV1::decode(buf)
+            .map(|entry| datetime_from_epoch(entry.report.unwrap().timestamp))?,
+        FileType::CellSpeedtestIngestReport => SpeedtestIngestReportV1::decode(buf)
+            .map(|entry| datetime_from_epoch(entry.report.unwrap().timestamp))?,
```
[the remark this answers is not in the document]
bbalser: Addressed in f7174b7, feels weird that I had to call `.map_err(|error| error.into())` but I couldn't find a better way to get it happy.

store/src/cli/info.rs
```diff
@@ -70,9 +70,26 @@ fn get_timestamp(file_type: &FileType, buf: &[u8]) -> Result<DateTime<Utc>> {
...
+                    Error::Custom(
+                        "CellHeartbeatIngestReportV1 does not contain a CellHeartbeatReqV1"
+                            .to_string(),
+                    )
+                })
+            })
+            .map(|heartbeat_req| datetime_from_epoch(heartbeat_req.timestamp))?,
         FileType::CellSpeedtestIngestReport => SpeedtestIngestReportV1::decode(buf)
-            .map(|entry| datetime_from_epoch(entry.report.unwrap().timestamp))?,
+            .map_err(|error| error.into())
```
madninja, line 84: I think you can do `.map_err(Error::from)` for a better shorthand?

bbalser: Fixed af297e3 and thank you, reads better

store/src/speedtest.rs
```diff
@@ -48,3 +57,55 @@ impl TryFrom<SpeedtestReqV1> for CellSpeedtest {
...
+                download_speed: 2,
+                latency: 1,
+                signature: vec![],
+            }),
+        };
+
+        let mut buffer = vec![];
+
+        report
+            .encode(&mut buffer)
```
madninja, line 99: use `encode_to_vec`

bbalser: Fixed 24ac815, can't believe I missed that function

madninja (approved):

PR #83, opened 2022-09-22
refactor out proto usage, add mor validations, misc fixes

andymck:
TODO:
- Add Readme to iot verifier
- Add deny list validation
- ~~Add internal struct for FollowerGatewayRespV1 proto~~
- Purge stale POC reports from DB
- Purge stale entropy from DB
- Add internal struct for entropy proto
- refactor refactor refactor make bad rust nice rust


store/src/lora_valid_poc.rs
```diff
@@ -111,9 +126,13 @@ impl TryFrom<LoraValidWitnessReport> for LoraValidWitnessReportV1 {
     type Error = Error;
     fn try_from(v: LoraValidWitnessReport) -> Result<Self> {
         let report: LoraWitnessReportReqV1 = v.report.into();
+        let location = match v.location {
```
madninja, line 129: Why is this still a String then?

store/src/lora_valid_poc.rs
```diff
@@ -98,9 +108,14 @@ impl TryFrom<LoraValidBeaconReport> for LoraValidBeaconReportV1 {
 impl TryFrom<LoraValidWitnessReportV1> for LoraValidWitnessReport {
     type Error = Error;
     fn try_from(v: LoraValidWitnessReportV1) -> Result<Self> {
+        let location = v
+            .location
+            .is_empty()
+            .not()
+            .then(|| v.location.parse::<u64>().unwrap());
```
madninja, line 115: nope.. don't unwrap.. you have a Result as an error.. also I don't think you need to do is_empty.not at all here

store/src/lora_valid_poc.rs
andymck: it is converted back to a string when going from internal struct to the proto def

store/src/lora_valid_poc.rs
```diff
@@ -86,9 +92,13 @@ impl TryFrom<LoraValidBeaconReport> for LoraValidBeaconReportV1 {
     type Error = Error;
     fn try_from(v: LoraValidBeaconReport) -> Result<Self> {
         let report: LoraBeaconReportReqV1 = v.report.into();
+        let location = match v.location {
```
madninja, line 95: would `v.location.unwrap_or_else(|| "").to_string()` work?

store/src/lora_valid_poc.rs
andymck: kinda have to unwrap here, otherwise i get

mismatched types
expected enum `std::option::Option<u64>`
   found enum `std::option::Option<std::result::Result<u64, std::num::ParseIntError>>`

madninja: call .transpose()

file_store/src/lora_invalid_poc.rs
```diff
@@ -0,0 +1,87 @@
...
+            .ok_or_else(|| Error::custom("unsupported invalid_reason"))?;
+        Ok(Self {
+            received_timestamp: datetime_from_epoch(v.received_timestamp),
+            reason: invalid_reason,
+            report: LoraBeaconReport::try_from(v.report.unwrap())?,
+        })
+    }
+}
+
+impl TryFrom<LoraInvalidBeaconReport> for LoraInvalidBeaconReportV1 {
```
jeffgrunewald, line 48: do we need this? doesn't the type safety assume a well-formed `LoraInvalidBeaconReport` struct such that you only need to encode it to get a `LoraInvalidBeaconReportV1` or call `into` to easily convert it? i understand the need for the `TryFrom` the proto struct into the internal struct but i would think this wouldn't be necessary

file_store/src/lora_valid_poc.rs
```diff
@@ -53,12 +54,34 @@ impl TryFrom<LoraValidPocV1> for LoraValidPoc {
...
+}
+
 impl TryFrom<LoraValidBeaconReportV1> for LoraValidBeaconReport {
     type Error = Error;
     fn try_from(v: LoraValidBeaconReportV1) -> Result<Self> {
+        let location = v
+            .location
+            .is_empty()
+            .not()
+            .then(|| v.location.parse::<u64>().unwrap());
```
jeffgrunewald, line 81: is this saying "if location is not empty then parse it as a u64"? what happens if the field _is_ empty? since the struct definition above is setting the location field to an `Option` i think the rust-y way to do this is to "assume a value is present and parsable and then handle the alternative if it isn't"; i believe this is equivalent and a little cleaner
```suggestion
        let location: Option<u64> = v.location.parse().ok();
```
or alternatively `let location = v.location.parse::<u64>().ok();` if you prefer.
the type annotation makes the assumption that the value will parse to a u64 and the `ok()` will convert the `Result(u64)` to a `Some(u64)` or else a `None` if the value is empty

jeffgrunewald: the `ok()` is actually even more forgiving, so _any_ error in parsing the string to a u64 will result in `None` meaning `"".parse()` and `"foo".parse()` will both be converted to `None`

lora_verifier/src/entropy.rs
```diff
@@ -15,6 +15,7 @@ pub struct Entropy {
     pub id: Vec<u8>,
     pub data: Vec<u8>,
     pub timestamp: DateTime<Utc>,
+    pub version: i32,
```
jeffgrunewald, line 18: does this want to be a `u32`? the proto version seems to be a `uint32`

jeffgrunewald: or is this a limitation of storing the value in postgres?

madninja: so.... the gateway repo has a `beacon` crate as part of the beacon PR that defines this exact structure We should endeavor to use that instead

file_store/src/lora_invalid_poc.rs
andymck: yeah good point.  to hasty with the copy pasta

file_store/src/lora_valid_poc.rs
andymck: if the proto location is an empty string then the struct value will be None.  If its not empty then the string will be parsed to an int.  Also has a suggestion from madninja to get rid of the unwrap and use transpose on the outer.  I plan on going over all the unwrap usage n a sep pass once i have ev thing functional

lora_verifier/src/entropy.rs
andymck: the reason for i32 here is that postgres wont accept a uint

file_store/src/lora_invalid_poc.rs
andymck: fixed here: c0be07d313edd828dc9308501946c2cfe445b779

store/src/lora_valid_poc.rs
andymck: It does not, doesnt like the empty string:
mismatched types
expected `u64`, found `&str`

ingest/src/server_lora.rs
```diff
@@ -89,9 +89,11 @@ pub async fn grpc_server(shutdown: triggered::Listener, server_mode: String) ->
 
     // Initialize uploader
     let (file_upload_tx, file_upload_rx) = file_upload::message_channel();
-    let file_upload = file_upload::FileUpload::from_env(file_upload_rx).await?;
+    let file_upload =
+        file_upload::FileUpload::from_env_with_prefix("INGESTOR", file_upload_rx).await?;
 
-    let store_path = dotenv::var("INGEST_STORE")?;
+    let store_path =
+        std::env::var("INGEST_STORE").unwrap_or_else(|_| String::from("/var/data/ingestor"));
```
jeffgrunewald, line 96:
```suggestion
       env::var("INGEST_STORE").unwrap_or_else(|_| String::from("/var/data/ingestor"));
```
you've got this imported in from a use statement above which looks like it can be combined with the other `std::{` on the line below it

jeffgrunewald (approved):

PR #85, opened 2022-09-23
Change mobile verifier to read heartbeat and speedtest ingest reports

bbalser:


maplant:
This LGTM, you just need to change the `file_list` that is passed to the  `GatheredShares::from_stream` function to one that takes in `FileType::CellHeartbeatIngestReport` and `FileType::CellSpeedtestIngestReport`

maplant (approved):

Issue #86, opened 2022-09-23
Cache rust dependencies in CI to speed up builds


jeffgrunewald:
CI runs can take quite a while, between the build, test, and packaging steps. This can be significantly sped up if we're able to cache the compiled dependencies from one run to the next when only the oracle code changes.

The rust Sqlx crate github actions cache through an action called https://github.com/launchbadge/sqlx/blob/76ae2865005cc79d56972d68d3be30805581103d/.github/workflows/sqlx.yml#L46 which may be applicable.

PR #87, opened 2022-09-23
Test cached CI

vihu:
# Summary
- Change order of `build` and `clippy` steps, do `build` first
- Add rust-cache CI step

Fix https://github.com/helium/oracles/issues/86

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -34,12 +34,6 @@ jobs:
         with:
           key: ${{ runner.os }}-build
 
-      - name: Run tests
-        run: cargo test
-      - uses: Swatinem/rust-cache@v1
-        with:
-          key: ${{ runner.os }}-test
-
       - name: Check formatting
         run: cargo fmt  -- --check
       - uses: Swatinem/rust-cache@v1
@@ -54,6 +48,12 @@ jobs:
       - uses: Swatinem/rust-cache@v1
         with:
           key: ${{ runner.os }}-clippy
+
+      - name: Run tests
+        run: cargo test
+      - uses: Swatinem/rust-cache@v1
+        with:
+          key: ${{ runner.os }}-test
 
       - name: PackageIngestorServer
         id: build_ingestor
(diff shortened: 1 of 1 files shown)


.github/workflows/rust.yml
```diff
@@ -19,26 +19,41 @@ jobs:
...
 
+      - name: Build
+        run: cargo build --all --release
+      - uses: Swatinem/rust-cache@v1
+        with:
+          key: ${{ runner.os }}-build
+
       - name: Run tests
         run: cargo test
+      - uses: Swatinem/rust-cache@v1
```
jeffgrunewald, line 39: This step will reuse cached dependencies and just recompile the application, right? Since tests aren’t compiled as a release?

.github/workflows/rust.yml
```diff
@@ -19,26 +19,41 @@ jobs:
...
+        with:
+          key: ${{ runner.os }}-test
 
       - name: Check formatting
         run: cargo fmt  -- --check
+      - uses: Swatinem/rust-cache@v1
```
jeffgrunewald, line 45: May not matter with as fast as I’ve seen rust tests run so far but if the test suite ever gets longer we probably wanna run fmt and possibly clippy first since those are more or less constant time tasks and failing those first will return the result to be fixed faster than if we wait on any long test suites to pass only to fail on a format or linter error

vihu: > we probably wanna run fmt and possibly clippy first since those are more or less constant time tasks and failing those first will return the result to be fixed faster

Well according to [this](https://www.reillywood.com/blog/rust-faster-ci/) blog, clippy can reuse build artifacts (which is why I flipped the order)

jeffgrunewald: Sorry, I meant “first” after the build step. So:
- build
- fmt
- clippy
- test
- If/then package

vihu: Sure, updated in bf11069

jeffgrunewald (approved):

PR #80, opened 2022-09-20
Use helium-crypto-rs

madninja:
This upgrades helium-crypto-rs to 0.4.4 which includes a sqlx-postgres feature. No more custom public_key.rs needed for any of the workspace members

PR #88, opened 2022-09-25
basic iot network rewards emission framework 

jeffgrunewald:
Implements the basic framework for a rewards emission server for helium iot network. 

The server.rs module needs only to have the logic of handling reward shares reports from the iot verifier to prepare rewards transactions to submit to the destination chain.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2381,6 +2381,41 @@ dependencies = [
 ]
 
 [[package]]
+name = "poc-iot-rewards"
+version = "0.1.0"
+dependencies = [
+ "async-trait",
+ "base64",
+ "chrono",
+ "db-store",
+ "dotenv",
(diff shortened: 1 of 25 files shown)


vihu:
I don't think we need a separate `public_key.rs` module in the workspace after #80 landed

jeffgrunewald:
> I don't think we need a separate `public_key.rs` module in the workspace after #80 landed

you're right; this was sitting out there as an open branch for the last two weeks and i hadn't rebased it since. fixed

jeffgrunewald:
before this gets merged, can i get a confirmation on the name `iot_` vs `lora_`? i can rename if we prefer to keep it consistent and more specific to the protocol (in case there are other "iot networks" on non-lora wireless protocols in the future)

PR #94, opened 2022-09-29
Instrument IO

dpezely:
Summary:
1. instruments:
- `file_store::file_store::{put, remove}` to:
  - __time__ writes to AWS (reads timing is less clear, since it returns a stream)
- `file_store::file_sink::write` to:
  - __count__ all successes and failures
- `db_store::{insert, fetch_or_insert_with, update}` to:
  - __count__ all successes and failures
  - __time__ execution durations
2. updates all affected call sites with the newly-required static tag
3. extends `poc_metrics::ActiveRequests` layer to also __time__ request handling durations (renaming it to simply `poc_metrics::Requests`) in addition to __counting__ currently active connections that it was already doing
4. updates the metrics layer usage in:
  - `ingest::server_5g`
5. adds the metrics layer to:
  - `ingest::server_lora`
  - `entropy::server`

### DIFF
--- a/db_store/src/lib.rs
+++ b/db_store/src/lib.rs
@@ -31,6 +31,21 @@ impl<T> MetaValue<T> {
     }
 }
 
+macro_rules! query_exec_timed {
+    ( $name:literal, $query:expr, $meth:ident, $exec:expr ) => {{
+        match poc_metrics::record_duration!(concat!($name, "_duration"), $query.$meth($exec).await) {
+            Ok(x) => {
+                metrics::increment_counter!(concat!($name, "_count"), "status" => "ok");
+                Ok(x)
+            }
+            Err(e) => {
+                metrics::increment_counter!(concat!($name, "_count"), "status" => "error");
+                Err(MetaError::SqlError(e))
+            }
+        }
+    }};
+}
+
 impl<T> MetaValue<T>
 where
     T: ToString,
@@ -39,19 +54,17 @@ where
     where
         E: sqlx::Executor<'c, Database = sqlx::Postgres>,
     {
-        sqlx::query(
+        let query = sqlx::query(
             r#"
-            insert into meta (key, value) 
-            values ($1, $2) 
-            on conflict (key) do update set 
-            value = EXCLUDED.value;
-            "#,
+                insert into meta (key, value)
+                values ($1, $2)
+                on conflict (key) do update set
+                value = EXCLUDED.value;
+                "#,
         )
         .bind(&self.key)
-        .bind(self.value.to_string())
-        .execute(exec)
-        .await?;
-        Ok(())
+        .bind(self.value.to_string());
+        query_exec_timed!("db_store_metavalue_insert", query, execute, exec).map(|_| ())
     }
 }
 
@@ -67,14 +80,13 @@ where
     where
         E: sqlx::Executor<'c, Database = sqlx::Postgres> + Copy,
     {
-        let str_val = sqlx::query_scalar::<_, String>(
+        let query = sqlx::query_scalar::<_, String>(
             r#"
             select value from meta where key = $1;
             "#,
         )
-        .bind(key)
-        .fetch_optional(exec)
-        .await?;
+        .bind(key);
+        let str_val = query_exec_timed!("db_store_metavalue_fetch", query, fetch_optional, exec)?;
 
         match str_val {
             Some(str_val) => {
@@ -97,7 +109,7 @@ where
     where
         E: sqlx::PgExecutor<'c>,
     {
-        sqlx::query(
+        let query = sqlx::query(
             r#"
             insert into meta (key, value) 
             values ($1, $2) 
@@ -106,9 +118,8 @@ where
             "#,
         )
         .bind(&self.key)
-        .bind(new_val.to_string())
-        .execute(exec)
-        .await?;
+        .bind(new_val.to_string());
+        let _ = query_exec_timed!("db_store_metavalue_update", query, execute, exec)?;
         Ok(std::mem::replace(&mut self.value, new_val))
     }
 }
--- a/entropy/src/entropy_generator.rs
+++ b/entropy/src/entropy_generator.rs
@@ -1,6 +1,6 @@
 use crate::{Error, Result};
 use chrono::Utc;
-use file_store::file_sink;
+use file_store::{file_sink, file_sink_write};
 use futures::TryFutureExt;
 use helium_proto::EntropyReportV1;
 use jsonrpsee::{
@@ -152,7 +152,12 @@ impl EntropyGenerator {
             entropy.to_string(),
             entropy.timestamp
         );
-        file_sink::write(file_sink, EntropyReportV1::from(entropy)).await?;
+        file_sink_write!(
+            "report_submission",
+            file_sink,
+            EntropyReportV1::from(entropy)
+        )
+        .await?;
         Ok(())
     }
 
--- a/ingest/src/server_5g.rs
+++ b/ingest/src/server_5g.rs
@@ -1,7 +1,7 @@
 use crate::{error::DecodeError, required_network, Error, EventId, Result};
 use chrono::Utc;
 use file_store::traits::MsgVerify;
-use file_store::{file_sink, file_upload, FileType};
+use file_store::{file_sink, file_sink_write, file_upload, FileType};
 use futures_util::TryFutureExt;
 use helium_crypto::{Network, PublicKey};
 use helium_proto::services::poc_mobile::{
@@ -72,16 +72,8 @@ impl poc_mobile::PocMobile for GrpcServer {
             received_timestamp: timestamp,
         };
 
-        match file_sink::write(&self.speedtest_req_tx, event).await {
-            Ok(_) => (),
-            Err(err) => tracing::error!("failed to store speedtest: {err:?}"),
-        }
-
-        match file_sink::write(&self.speedtest_report_tx, report).await {
-            Ok(_) => (),
-            Err(err) => tracing::error!("failed to store speedtest: {err:?}"),
-        }
-
+        _ = file_sink_write!("speedtest_req", &self.speedtest_req_tx, event).await;
+        _ = file_sink_write!("speedtest_report", &self.speedtest_report_tx, report).await;
         metrics::increment_counter!("ingest_server_speedtest_count");
         Ok(Response::new(event_id.into()))
     }
@@ -107,17 +99,8 @@ impl poc_mobile::PocMobile for GrpcServer {
             report: Some(event.clone()),
             received_timestamp: timestamp,
         };
-
-        match file_sink::write(&self.heartbeat_req_tx, event).await {
-            Ok(_) => (),
-            Err(err) => tracing::error!("failed to store heartbeat: {err:?}"),
-        }
-
-        match file_sink::write(&self.heartbeat_report_tx, report).await {
-            Ok(_) => (),
-            Err(err) => tracing::error!("failed to store heartbeat: {err:?}"),
-        }
-
+        _ = file_sink_write!("heartbeat_req", &self.heartbeat_req_tx, event).await;
+        _ = file_sink_write!("heartbeat_report", &self.heartbeat_report_tx, report).await;
         metrics::increment_counter!("ingest_server_heartbeat_count");
         // Encode event digest, encode and return as the id
         Ok(Response::new(event_id.into()))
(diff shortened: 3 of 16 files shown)


status/Cargo.toml
```diff
@@ -39,6 +39,7 @@ tonic = "0"
 http = "*"
 metrics = "0"
 strum = "*"
+axum-prometheus = "0.1.0"
```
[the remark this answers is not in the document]
xandkar: @madninja primarily for the http request timers that it conveniently implements, rather than for the endpoint.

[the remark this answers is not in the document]
madninja: that just seems like it's replicating features that aren't quite what we need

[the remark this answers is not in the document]
xandkar: ok, i'll replace it with artisanal timers.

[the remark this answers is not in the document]
madninja: Note that the status crate has been deleted since it was only meant to be temporary

db_store/src/lib.rs
```diff
@@ -31,6 +31,21 @@ impl<T> MetaValue<T> {
     }
 }
 
+macro_rules! query_exec_timed {
```
jeffgrunewald, line 34: shouldn't this macro be in the `poc-metrics` support crate so we could time other db access than just the db_store uses? the db_store is a shared behavior support crate for check-pointing pipeline progress within a particular service but then each service does _tons_ of db access "directly" via sqlx

poc_iot_injector/src/server.rs
```diff
@@ -70,6 +70,7 @@ impl Server {
     }
 
     async fn handle_poc_tick(&mut self) -> Result {
+        metrics::increment_counter!("poc_iot_injector_server_tick_count");
```
jeffgrunewald, line 73: what's the utility of this measurement? this seems like it's just measuring rust's ability to loop on a duration?

db_store/src/lib.rs
xandkar: maybe? it just didn't seem well-thought-out-enough for general use.

metrics/src/lib.rs
```diff
@@ -32,46 +32,67 @@ pub fn install_metrics() {
...
+    pub metric_name_time: &'static str,
 }
 
-impl ActiveRequestsLayer {
-    pub fn new(metric_name: &'static str) -> Self {
-        Self { metric_name }
+#[macro_export]
+macro_rules! request_layer {
+    ( $metric_name:literal ) => {{
+        poc_metrics::RequestsLayer {
```
jeffgrunewald, line 60: is this module prefix necessary because the context the macro is executed in may not have the `poc_metrics::RequestsLayer` in scope or can it infer the necessary struct by virtue of the macro being defined in the same scope as the struct?

metrics/src/lib.rs
```diff
@@ -32,46 +32,67 @@ pub fn install_metrics() {
 // function block. Such code is possible when async closures become stable.
 #[macro_export]
 macro_rules! record_duration {
-    ( $metric_name:literal, $e:expr ) => {{
+    ( $metric_name:expr, $e:expr ) => {{
```
xandkar, line 35: This is relaxed because `metrics` can also accept static `str` refs and owned `String` in addition to literals. Static refs is what I mainly needed it to accept in order to avoid duplicating counter calls everywhere.

jeffgrunewald (commented):
looks generally good to me and pretty useful metrics being captured. one thing that would be great is if we can shorten those labels at the call site for each measurement by having the macro grab the crate name under the hood and concatenating it with the passed-in label? so if it were possible for the macro to create the complete label by string appending `env!(CARGO_NAME)` with whatever context-specific label you want to add that would be awesome!

metrics/src/lib.rs
xandkar: I _assumed_ the former and included it, but IDK for sure.

poc_iot_injector/src/server.rs
xandkar: Agreed. Removing this one.

xandkar:
> looks generally good to me and pretty useful metrics being captured. one thing that would be great is if we can shorten those labels at the call site for each measurement by having the macro grab the crate name under the hood and concatenating it with the passed-in label? so if it were possible for the macro to create the complete label by string appending `env!(CARGO_NAME)` with whatever context-specific label you want to add that would be awesome!

that is a neat idea! it would need a wrapper macro for `file_sink::write` calls - let me give it a shot.

ingest/src/server_lora.rs
```diff
@@ -97,13 +103,18 @@ impl poc_lora::PocLora for GrpcServer {
...
-            Ok(_) => tracing::debug!(
+        let _ = file_sink::write(
+            "ingest_server_lora_witness_report",
+            &self.lora_witness_report_tx,
+            report,
+        )
+        .await
+        .map(|_| {
+            tracing::debug!(
                 "successfully processed witness report, returning response {}",
```
dpezely, line 114: Since the `Err(...)` clause of the original `match` went away, the debug log message shouldn't assume success:
```suggestion
                "processed witness report, returning response {}",
```

dpezely (commented):
GitHub won't let me approve since I initiated this PR way back when.

Looking from perspective of what I had learned about Prometheus + Grafana, this **looks good** with one minor suggestion for text within quoted string for log message.

xandkar:
> > looks generally good to me and pretty useful metrics being captured. one thing that would be great is if we can shorten those labels at the call site for each measurement by having the macro grab the crate name under the hood and concatenating it with the passed-in label? so if it were possible for the macro to create the complete label by string appending `env!(CARGO_NAME)` with whatever context-specific label you want to add that would be awesome!
> 
> that is a neat idea! it would need a wrapper macro for `file_sink::write` calls - let me give it a shot.

ok, so i was able to _reduce_ the required prefix by auto prefixing it with `module_path!()`, but couldn't yet find enough static location data sources to completely _eliminate_ the manual tag.

ingest/src/server_lora.rs
xandkar: @dpezely `.map` maps `Ok` case and `.map_err` map the `Err` one and both cases are already logged generically within `write`.

So this one is kinda redundant, the only reason I put it back is because I saw that it was logging `event_id` on success (which can't be replicated generically by `write`) and thought that maybe this particular message is operationally relied upon - otherwise I'd rather remove this logging map altogether.

@bbalser do you know if this log message is particularly relied upon by anyone?

andymck (approved):

entropy/src/entropy_generator.rs
```diff
@@ -152,7 +152,12 @@ impl EntropyGenerator {
             entropy.to_string(),
             entropy.timestamp
         );
-        file_sink::write(file_sink, EntropyReportV1::from(entropy)).await?;
+        file_store::write!(
+            "report_submission",
```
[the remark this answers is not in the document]
xandkar: Ah, so I did have it manually fully-qualified at first, but then on @jeffgrunewald's suggestion switched it to automatically prefix module caller path, so these tags are now just extra semantic context hints on top of the path: https://github.com/helium/oracles/pull/94/commits/3a0b85b4c26c8a4d980436c9d8cc91c9f98763b4

ingest/src/server_5g.rs
```diff
@@ -198,9 +183,7 @@ pub async fn grpc_server(shutdown: triggered::Listener, server_mode: String) ->
     //TODO start a service with either the poc mobile or poc lora endpoints only - not both
     //     use _server_mode (set above ) to decide
     let server = transport::Server::builder()
-        .layer(poc_metrics::ActiveRequestsLayer::new(
-            "ingest_server_grpc_connection_count",
-        ))
+        .layer(poc_metrics::request_layer!("ingest_server_grpc_connection"))
```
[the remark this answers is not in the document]
xandkar: Yes - count of _and_ time to process a connection/request.
Handling: `poc_metrics::Service.call` in `metrics/src/lib.rs`: before handler is called - counter is incremented and timer is started, after handler returns - counter is decremented and timer is stopped.

entropy/src/entropy_generator.rs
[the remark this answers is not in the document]
xandkar: One possible problem here is that I'm not sure how `::` will be handled in Prometheus and Grafana paths.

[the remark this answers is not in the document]
jeffgrunewald: I would play it safe and string-replace the colons with dashes or underscores, whichever Prometheus prefers

ingest/src/server_lora.rs
xandkar: > To my knowledge our logs aren't going anywhere, so I don't believe this log is being relied upon.

Right, but people could also be looking for this message manually.

entropy/src/entropy_generator.rs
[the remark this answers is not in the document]
xandkar: So compile-time string manipulation is possible, but somewhat involved, so I leaned on `const_format` crate: https://github.com/helium/oracles/pull/94/commits/b20a6d8ed3afc49afccd232cf510f38a14ad51a2

@madninja how objectionable does that look?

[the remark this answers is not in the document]
jeffgrunewald: I think it’s new as of 1.64 but it looks like `const_str` is also an option?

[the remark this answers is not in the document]
jeffgrunewald: unless someone if more familiar with `const_str` this one is probably better by virtue of number of downloads

ingest/src/server_lora.rs
```diff
@@ -65,7 +65,7 @@ impl poc_lora::PocLora for GrpcServer {
             .map_err(|_| Status::invalid_argument("invalid signature"))?;
 
         let event_id = EventId::from(&event);
-        match file_sink::write(&self.lora_beacon_report_tx, report).await {
+        match file_store::write!("beacon_report", &self.lora_beacon_report_tx, report).await {
```
jeffgrunewald, line 68: i would lean on keeping this macro/module naming consistent with the current function/method strategy, i.e. we read files from a file_store (s3 bucket) and write files to a file_sink (local file system that gets synced to s3 bucket out-of-band) to preserve the semantics of the two operational strategies

poc_mobile_verifier/src/subnetwork_rewards.rs
```diff
@@ -66,13 +66,14 @@ impl SubnetworkRewards {
         self,
         subnet_rewards_tx: &file_sink::MessageSender,
     ) -> file_store::Result<oneshot::Receiver<file_store::Result>> {
-        file_sink::write(
+        file_store::write!(
+            "subnet_rewards",
```
jeffgrunewald, line 70: unless you want to take some conditional extra load into the macro that preps the prometheus label we may want to rethink the `module_path!()` as a component of the label; in a situation like this it would produce the label `poc_mobile_verifier_subnetwork_rewards_subnet_rewards`; it's probably still a good strategy if the macro either allows for the label to be optional altogether (but would rely on implicit understanding of the naming strategy which the macro is hiding) or else conditionally lop off the last element of the path when it matches the supplied suffix.

the name of the calling app plus the supplied tag would probably be a good happy medium if this gets overly redundant

metrics/src/lib.rs
dpezely: This builds cleanly *without* the prefix, so please omit it.

Otherwise, we should rename `metrics` crate to `poc_metrics` to ensure consistency.  Not all oracles involve PoC.  e.g., in my branch, I named the alias `oracle_metrics` instead of `poc_metrics`.

xandkar: > This builds cleanly _without_ the prefix, so please omit it.
> 

I think it only works accidentally, because `poc_metrics` happens to already be in scope.

> Otherwise, we should rename `metrics` crate to `poc_metrics` to ensure consistency. Not all oracles involve PoC. e.g., in my branch, I named the alias `oracle_metrics` instead of `poc_metrics`.

It _is_ named `poc_metrics` in `Cargo.toml` and only the directory is called `metrics` - not ideal - :100: agree. Probably best to handle it in a separate PR though, IMO.

ingest/src/server_lora.rs
xandkar: Good call. I renamed the macro to `file_sink_write`. Deep namespacing doesn't seem possible with macros - you only get one level - crate.

poc_mobile_verifier/src/subnetwork_rewards.rs
xandkar: I made it optional. But, the more I think about it - your happy medium of crate+label seems best - no implicit magic to worry about and tags remain stable after refactoring (assuming crate names won't change much, of course).

xandkar: @jeffgrunewald Switched it to the happy medium.

ingest/src/server_lora.rs
xandkar: I resolved this by adding an optional logging context parameter, where things like `event_id` can be passed and subsequently included in the otherwise-generic log statements: https://github.com/helium/oracles/pull/94/commits/88d40d4ed408b6eefa88a700d3223a8f7ea6856e

bbalser (approved):

PR #100, opened 2022-10-01
Fix received timestamps of iot ingest reports to have milliseconds

bbalser:


jeffgrunewald (approved):

madninja (commented):
Please look at the `LoraBeaconReportReqV1` and `LoraBeaconIngestReportV1` and any of the report/ingest_report struct try_from conversion functions. They're assuming the timestamp of the actual beacon/witness is in seconds and should be in nanos.

PR #101, opened 2022-10-03
denylist verification for witness and receipt reports

andymck:
Adds support for denylist verification for beacon and witness reports.  
The denylist is fetched from github on startup and then periodically checked for updates.   

Once a denylist has been fetched from github, the filter will be saved locally.  Next time the verifier is restarted, this locally saved filter will be used as the initial filter during init and replaced with any newer version from github.
This enables a default filter to always be in use should github be unreachable for whatever reason.

### DIFF
--- a/denylist/src/client.rs
+++ b/denylist/src/client.rs
@@ -2,7 +2,7 @@ use crate::{error::Error, models::metadata::DenyListMetaData, Result};
 use std::{str, time::Duration};
 
 /// The default client useragent for denylist http requests
-static USERAGENT: &str = "https://github.com/helium/miner";
+static USERAGENT: &str = "oracle/poc_iot_verifier/1.0";
 /// The default timeout for http requests
 pub const DEFAULT_TIMEOUT: Duration = Duration::from_secs(5);
 
--- a/denylist/src/client.rs
+++ b/denylist/src/client.rs
@@ -1,0 +1,47 @@
+use crate::{error::Error, models::metadata::DenyListMetaData, Result};
+use std::{str, time::Duration};
+
+/// The default client useragent for denylist http requests
+static USERAGENT: &str = "https://github.com/helium/miner";
+/// The default timeout for http requests
+pub const DEFAULT_TIMEOUT: Duration = Duration::from_secs(5);
+
+#[derive(Clone, Debug)]
+pub struct DenyListClient {
+    pub client: reqwest::Client,
+}
+
+// TODO: basic non production client. productionise....
+impl DenyListClient {
+    pub fn new() -> Result<Self> {
+        let client = reqwest::Client::builder()
+            .gzip(true)
+            .user_agent(USERAGENT)
+            .timeout(DEFAULT_TIMEOUT)
+            .build()
+            .map_err(Error::from)?;
+        Ok(Self { client })
+    }
+
+    pub async fn get_metadata(&mut self, url: &String) -> Result<DenyListMetaData> {
+        let response = self.client.get(url).send().await?;
+        match response.status() {
+            reqwest::StatusCode::OK => {
+                let json = response.json::<DenyListMetaData>().await?;
+                Ok(json)
+            }
+            other => Err(Error::UnexpectedStatus(other.to_string())),
+        }
+    }
+
+    pub async fn get_bin(&mut self, url: &String) -> Result<Vec<u8>> {
+        let response = self.client.get(url).send().await?;
+        match response.status() {
+            reqwest::StatusCode::OK => {
+                let bytes = response.bytes().await?;
+                Ok(bytes.to_vec())
+            }
+            other => Err(Error::UnexpectedStatus(other.to_string())),
+        }
+    }
+}
--- a/denylist/src/denylist.rs
+++ b/denylist/src/denylist.rs
@@ -13,6 +13,7 @@ pub const SERIAL_SIZE: usize = 32;
 const PUB_KEY_B58: &str = "1SbEYKju337P6aYsRd9DT2k4qgK5ZK62kXbSvnJgqeaxK3hqQrYURZjL";
 /// a copy of the last saved filter bin downloaded from github
 /// if present will be used to initialise the denylist upon verifier startup
+// TODO: look at using the tempfile crate to handle this
 const FILTER_BIN_PATH: &str = "./tmp/last_saved_filter.bin";
 
 #[derive(Serialize)]
@@ -91,7 +92,7 @@ impl DenyList {
     pub async fn check_key(&self, pub_key: &PublicKey) -> bool {
         if self.filter.len() == 0 {
             tracing::warn!("empty denylist filter, rejecting key");
-            return false;
+            return true;
         }
         self.filter.contains(&public_key_hash(pub_key))
     }
--- a/denylist/src/denylist.rs
+++ b/denylist/src/denylist.rs
@@ -1,0 +1,134 @@
+use crate::denylist::client::DenyListClient;
+use crate::Result;
+use crate::{models::metadata::Asset, *};
+use bytes::Buf;
+use helium_crypto::{PublicKey, Verify};
+use serde::Serialize;
+use std::{fs, hash::Hasher, path, str::FromStr};
+use twox_hash::XxHash64;
+use xorf::{Filter as XorFilter, Xor32};
+
+pub const SERIAL_SIZE: usize = 32;
+
+/// the pubkey used to verify the signature of denylist updates
+// TODO: is there a better home for this key ?
+const PUB_KEY_B58: &str = "1SbEYKju337P6aYsRd9DT2k4qgK5ZK62kXbSvnJgqeaxK3hqQrYURZjL";
+/// a copy of the last saved filter bin downloaded from github
+/// if present will be used to initialise the denylist upon verifier startup
+const FILTER_BIN_PATH: &str = "./tmp/last_saved_filter.bin";
+
+#[derive(Serialize)]
+pub struct DenyList {
+    pub tag_name: u64,
+    #[serde(skip_serializing)]
+    pub filter: Xor32,
+}
+
+impl Default for DenyList {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl DenyList {
+    pub fn new() -> Self {
+        tracing::debug!("initializing new denylist");
+        // if exists default to the local saved filter bin,
+        // otherwise default to empty filter
+        // a local filter should always be present
+        // after the verifier has been run at least once
+        // in the current dir and has previously successfully downloaded
+        // a filter from github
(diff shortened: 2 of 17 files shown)


denylist/src/client.rs
```diff
@@ -0,0 +1,66 @@
+use crate::Result;
+use crate::{error::Error, models::metadata::DenyListMetaData};
+use std::str;
+use std::time::Duration;
+
+/// The default client useragent for denylist http requests
+static DEFAULT_USERAGENT: &str = "https://github.com/helium/miner";
```
jeffgrunewald, line 7: is this actually being verified against anything? we should probably make it something more specific to the repo instead of referencing the old miner one (which doesn't technically meet the standards for user-agent anyway)

madninja: No it's not verified.. it's just useful when you ask GH questions/reports about origins.. make it something more user agent centric and related to this oracle

jeffgrunewald (commented):
looks like we have two patterns emerging here for "check some hotspot detail against a list" between the denylist and the hex density scaler; i wonder if we prefer the pattern defined here where the denylist is strictly a library and all of its updating is performed by the calling applications vs. the pattern in the hex density scaler where the scaler is an independent server that just needs to be managed/monitored by the caller and they communicate over channels? probably not necessary to answer during this first pass but good to get some consensus started now

denylist/src/models/metadata.rs
```diff
@@ -0,0 +1,66 @@
+// model representing the JSON payload returned from:
```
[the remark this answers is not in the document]
andymck: yeah most fields dont need to be public, updated here: ee85731ad9962e5d2599ce65fadcac396d153117 

However i do think it is appropriate to maintain the full JSON model for correctness.

denylist/src/models/mod.rs
```diff
@@ -0,0 +1 @@
+pub mod metadata;
```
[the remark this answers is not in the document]
andymck: i dunno, i think i prefer to keep the models dir...keeps it distinct that it is a JSON model

poc_iot_verifier/src/runner.rs
```diff
@@ -27,29 +28,44 @@ use sqlx::PgPool;
...
 }
 
 impl Runner {
     pub async fn from_env() -> Result<Self> {
         let pool = mk_db_pool(LOADER_DB_POOL_SIZE as u32).await?;
         let store_path =
             std::env::var("VERIFIER_STORE").unwrap_or_else(|_| String::from("/var/data/verifier"));
-        Ok(Self { pool, store_path })
+        let deny_list_latest_url = std::env::var("DENYLIST_URL").unwrap_or_else(|_| {
+            String::from("https://api.github.com/repos/helium/denylist/releases/latest")
```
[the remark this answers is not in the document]
andymck: fixed here: 5f309c27590cd50f72ba914e060c580dce07db39

denylist/src/models/metadata.rs
```diff
@@ -0,0 +1,66 @@
...
+pub struct DenyListMetaData {
+    url: String,
+    assets_url: String,
+    upload_url: String,
+    html_url: String,
+    id: i64,
+    author: Item,
+    pnode_id: String,
+    pub tag_name: String,
+    target_commitish: String,
```
vihu, line 16: Is `target_commitish` a typo?

denylist/src/denylist.rs
```diff
@@ -0,0 +1,165 @@
+use crate::denylist::client::DenyListClient;
+use crate::Result;
+use crate::{models::metadata::Asset, *};
```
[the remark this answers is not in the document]
jeffgrunewald: Using the `*` pattern above is importing all of the module beneath `crate` which would mean all modules, which seems excessive and also redundant when you have selectively imported just a few named modules from the crate prior to it.

denylist/src/client.rs
```diff
@@ -0,0 +1,61 @@
...
```
jeffgrunewald, line 7: This shouldn’t be `miner`

jeffgrunewald: It’s also not overridable so calling it default seems misleading

denylist/src/models/metadata.rs
jeffgrunewald: unfortunately not; can confirm

denylist/src/client.rs
```diff
@@ -0,0 +1,51 @@
+use crate::Result;
+use crate::{error::Error, models::metadata::DenyListMetaData};
+use std::str;
+use std::time::Duration;
```
jeffgrunewald, line 4:
```suggestion
use crate::{error::Error, models::metadata::DenyListMetaData, Result};
use std::{str, time::Duration};
```

denylist/src/client.rs
```diff
@@ -0,0 +1,51 @@
...
+
+/// The default client useragent for denylist http requests
+static DEFAULT_USERAGENT: &str = "https://github.com/helium/miner";
```
jeffgrunewald, line 7:
```suggestion
static DEFAULT_USERAGENT: &str = "github.com/helium/oracles poc-iot-verifier rust2021ed";
```

poc_iot_verifier/src/settings.rs
```diff
@@ -11,6 +11,9 @@ pub struct Settings {
     pub log: String,
     /// Cache location for generated verified reports
     pub cache: String,
+
+    #[serde(default = "default_denylist_url")]
+    pub denylist_url: String,
```
jeffgrunewald, line 16: shouldn't these be in the denylist settings module and then in the iot verifier's toml file under a `[denylist]` key? then the entry for iot verifier would just be `pub denylist: denylist: denylist::Settings,`

denylist/src/denylist.rs
```diff
@@ -0,0 +1,135 @@
...
+            // call to update_to_latest
+            tag_name: 0,
+            filter,
+        }
+    }
+
+    pub async fn update_to_latest(&mut self, metadata_url: &String) -> Result {
+        tracing::info!("checking for updated denylist");
+        let mut dl_client = DenyListClient::new()?;
+        let metadata = dl_client.get_metadata(metadata_url).await?;
```
jeffgrunewald, line 60:
```suggestion
            client: DenyListClient::new()?,
            filter,
        }
    }

    pub async fn update_to_latest(&mut self, metadata_url: &String) -> Result {
        tracing::info!("checking for updated denylist");
        let metadata = self.client.get_metadata(metadata_url).await?;
```

denylist/src/denylist.rs
```diff
@@ -0,0 +1,135 @@
...
+                .assets
+                .into_iter()
+                .filter(|a| a.name == "filter.bin")
+                .collect();
+            // we should be left with a single asset
+            // at least this is the assumption the erlang implementation followed
+            if let Some(asset) = assets.first() {
+                tracing::debug!("found asset for tag");
+                let asset_url = &asset.browser_download_url;
+                let bin = dl_client.get_bin(asset_url).await?;
```
jeffgrunewald, line 76:
```suggestion
                let bin = self.client.get_bin(asset_url).await?;
```

denylist/src/denylist.rs
```diff
@@ -0,0 +1,135 @@
...
+fn public_key_hash(public_key: &PublicKey) -> u64 {
+    let mut hasher = XxHash64::default();
+    hasher.write(&public_key.to_vec());
+    hasher.finish()
+}
+
+/// save a copy of the xor file locally
+// the local copy will be used should during init
+// github be unreachable
+pub fn save_local_filter_bin(bin: &Vec<u8>, path: &str) {
```
jeffgrunewald, line 126: make this return a result?

poc_iot_verifier/src/runner.rs
```diff
@@ -34,30 +35,39 @@ use tokio::time;
 const DB_POLL_TIME: time::Duration = time::Duration::from_secs(30);
 /// the cadence in seconds at which hotspots are permitted to beacon
 const BEACON_INTERVAL: i64 = 10 * 60; // 10 mins
-
+/// the cadence at which we poll for an updated denylist
+const DENYLIST_POLL_TIME: time::Duration = time::Duration::from_secs(180);
```
jeffgrunewald, line 39: this should be in the settings for the denylist

denylist/src/client.rs
```diff
@@ -0,0 +1,47 @@
...
+static USERAGENT: &str = "oracle/poc_iot_verifier/1.0";
+/// The default timeout for http requests
+pub const DEFAULT_TIMEOUT: Duration = Duration::from_secs(5);
+
+#[derive(Clone, Debug)]
+pub struct DenyListClient {
+    pub client: reqwest::Client,
+}
+
+// TODO: basic non production client. productionise....
```
bbalser, line 14: what needs to be done to productionise this?

denylist/src/denylist.rs
```diff
@@ -0,0 +1,142 @@
...
+use helium_crypto::{PublicKey, Verify};
+use serde::Serialize;
+use std::{fs, hash::Hasher, path, str::FromStr};
+use twox_hash::XxHash64;
+use xorf::{Filter as XorFilter, Xor32};
+
+pub const SERIAL_SIZE: usize = 32;
+
+/// the pubkey used to verify the signature of denylist updates
+// TODO: is there a better home for this key ?
```
bbalser, line 12: Still have a TODO here

andymck: ...oh just have to remove the TODO.  Its all good now.

denylist/src/client.rs
andymck: ..oh just have to remove the TODO. Its all good now.

bbalser (approved):

vihu (approved):

denylist/src/denylist.rs
```diff
@@ -0,0 +1,142 @@
...
+const PUB_KEY_B58: &str = "1SbEYKju337P6aYsRd9DT2k4qgK5ZK62kXbSvnJgqeaxK3hqQrYURZjL";
+/// a copy of the last saved filter bin downloaded from github
+/// if present will be used to initialise the denylist upon verifier startup
+const FILTER_BIN_PATH: &str = "./tmp/last_saved_filter.bin";
```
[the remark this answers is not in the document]
andymck: added todo here: 3d1b318ac6f8b41adf98cf3901a5167be6c1589a

denylist/src/denylist.rs
```diff
@@ -0,0 +1,142 @@
...
+        }
+        Ok(())
+    }
+
+    pub async fn check_key(&self, pub_key: &PublicKey) -> bool {
+        if self.filter.len() == 0 {
+            tracing::warn!("empty denylist filter, rejecting key");
+            return false;
+        }
+        self.filter.contains(&public_key_hash(pub_key))
```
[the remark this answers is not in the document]
andymck: well spotted, if the filter is empty it should be returning true...not false

PR #106, opened 2022-10-04
density scaler crate

jeffgrunewald:
implement a density scaling crate; the density_scaler runs as a separate tokio process within a service. it schedules a recurring process at a regular interval that queries a blockchain-node, retrieves the h3dex index location of all active gateways over a grpc stream and computes the scaling factor for all res 11 hexes, storing them in a wrapped hashmap. the scaler process also provides a query api wrapped around message channels for communicating between the calling service and the scaler.
- [x] test this with a collection of hexes

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,17 @@
 [workspace]
-members = [ "ingest", "db_store", "file_store", "mobile_rewards", "poc_mobile_verifier", "entropy",
-"poc_iot_verifier", "poc_iot_injector", "node_follower"]
+members = [
+    "ingest",
+    "db_store",
+    "file_store",
+    "mobile_rewards",
+    "poc_mobile_verifier",
+    "entropy",
+    "poc_iot_verifier",
+    "poc_iot_injector",
+    "node_follower",
+    "density_scaler",
+    "metrics",
+]
 
 [workspace.package]
 authors = ["Nova Labs <info@nova-labs.com>"]
@@ -51,4 +62,3 @@ futures-util = "*"
 prost = "*"
 once_cell = "1"
 lazy_static = "1"
-
--- a/density_scaler/src/hex.rs
+++ b/density_scaler/src/hex.rs
@@ -22,9 +22,10 @@ impl HexResConfig {
 
 type HexMap = HashMap<H3Cell, u64>;
 
-const MAX_RES: u8 = 11;
 const DENSITY_TGT_RES: u8 = 4;
+const MAX_RES: u8 = 11;
 const USED_RES: Range<u8> = DENSITY_TGT_RES..MAX_RES;
+const SCALING_RES: Range<u8> = DENSITY_TGT_RES..MAX_RES + 2;
 
 lazy_static! {
     static ref HIP17_RES_CONFIG: HashMap<u8, HexResConfig> = {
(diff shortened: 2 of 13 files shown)


Cargo.toml
```diff
@@ -1,6 +1,15 @@
...
+members = [
+    "ingest",
+    "db_store",
+    "file_store",
+    "mobile_rewards",
+    "status",
+    "mobile_verifier",
+    "entropy",
+    "lora_verifier",
+    "poc_iot_injector",
```
xandkar, line 11: Missing `density_scaler`?

jeffgrunewald: That’s my assumption since cargo can’t find my test there. I had thought this was intentional since the metrics support crate isn’t listed either and it is likewise only run in the context of other crate services but I’m guessing that was just an oversight i cargo-culted

andymck:
@jeffgrunewald what is the expectation re the verifier and how it utilizes this ?  Is the plan still to provide an API where the verifier passes in the public key and gets a hex scale value in response ?

jeffgrunewald:
> @jeffgrunewald what is the expectation re the verifier and how it utilizes this ? Is the plan still to provide an API where the verifier passes in the public key and gets a hex scale value in response ?

@andymck that's close to how you use the density scaler. You'd pull this into the verifier by doing the following:
- add a field for the density scaler's query sender struct type to the verifier server struct
- when you're starting the verifier and creating an instance of the verifier server struct call density scaler's query_channel function and store the sender in the verifier server
- pass the query channel receiver to the density scaler's run method, along with a clone of the verifier's shutdown trigger and call that run as one of your arms in the verifier's tokio::select run loops (this would be roughly equivalent to starting and supervising an erlang process)
- whenever you need to query the density scaler, you'd pass the gateway's _location_ string to the QuerySender.query` method and just await the response, which would be a Result<Option<f64>> <- unwrap that appropriately and you're good to generate your report

density_scaler/src/hex.rs
```diff
@@ -0,0 +1,300 @@
...
+        conf_map.insert(4, HexResConfig::new(1, 250, 800));
+        conf_map.insert(5, HexResConfig::new(1, 100, 400));
+        conf_map.insert(6, HexResConfig::new(1, 25, 100));
+        conf_map.insert(7, HexResConfig::new(2, 5, 20));
+        conf_map.insert(8, HexResConfig::new(2, 1, 4));
+        conf_map.insert(9, HexResConfig::new(2, 1, 2));
+        conf_map.insert(10, HexResConfig::new(2, 1, 1));
+        conf_map
+    };
+}
```
JayKickliter, line 45: A static hashmap here may be a overkill. An alternative would be an array:

```suggestion
static HIP17_RES_CONFIG: [Option<HexResConfig>; 11] = [
    None,                                 // 0
    None,                                 // 1
    None,                                 // 2
    None,                                 // 3
    Some(HexResConfig::new(1, 250, 800)), // 4
    Some(HexResConfig::new(1, 100, 400)), // 5
    Some(HexResConfig::new(1, 25, 100)),  // 6
    Some(HexResConfig::new(2, 5, 20)),    // 7
    Some(HexResConfig::new(2, 1, 4)),     // 8
    Some(HexResConfig::new(2, 1, 2)),     // 9
    Some(HexResConfig::new(2, 1, 1)),     // 10
];
```

density_scaler/src/hex.rs
```diff
@@ -0,0 +1,300 @@
...
+    pub fn new(neighbors: u64, target: u64, max: u64) -> Self {
+        Self {
+            neighbors,
+            target,
+            max,
+        }
+    }
+}
+
+type HexMap = HashMap<H3Cell, u64>;
```
JayKickliter, line 23: If HashMaps become a bottleneck, I'm happy to add whatever is needed to make HexTrees usable here.

density_scaler/src/hex.rs
jeffgrunewald: i like this better for the use; TIL `pub const fn`

vihu (approved):

bbalser (approved):

PR #111, opened 2022-10-05
Make poc_mobile_verifier rewards calculation the same as that of mobile_rewards

maplant:


jeffgrunewald (approved):

andymck (approved):

poc_mobile_verifier/src/subnetwork_rewards.rs
```diff
@@ -645,8 +645,11 @@ mod test {
         moving_avgs.update(&speed_shares);
 
         let hotspot_shares = hotspot_shares(&shares, &moving_avgs);
+        // TODO: re-add this later
```
xandkar, line 648: Under which conditions should this be re-added later? A little more info would be helpful.

PR #115, opened 2022-10-06
Separate node-follower workspace

vihu:
- Add node_follower directory
- Common behavior only exists in node_follower/src/lib.rs
- Update mobile_rewards to use node_follower
- Update poc_iot_verifier to use node_follower


poc_iot_verifier/src/traits/gateway_info_resolver.rs
```diff
@@ -0,0 +1,8 @@
+use crate::{gateway_resp::FollowerGatewayResp, Result};
+use async_trait::async_trait;
+use helium_crypto::PublicKey;
+
+#[async_trait]
```
maplant, line 5: You should put the implementation for the follower service along side the gatewayinforesolver trait. Additionally, there's no reason to put these traits inside their own "traits" module. Move this module into the root. (You wouldn't have a "structs" module for example)

mobile_rewards/src/cli/generate.rs
```diff
@@ -21,7 +20,7 @@ pub struct Cmd {
 impl Cmd {
     pub async fn run(&self) -> Result {
         let store = FileStore::from_env().await?;
-        let mut follower_service = FollowerService::from_env()?;
+        let mut follower_service = FollowerService::from_env().expect("follower_service env error");
```
madninja, line 23: Return the error instead of crashing. That'll mean extending the Error enum

mobile_rewards/src/server.rs
```diff
@@ -52,7 +55,7 @@ impl Server {
         let result = Self {
             pool,
             keypair,
-            follower_service: FollowerService::from_env()?,
+            follower_service: FollowerService::from_env().expect("follower service env error"),
```
madninja, line 58: ditto

mobile_rewards/src/cli/generate.rs
jeffgrunewald: How would you recover from this error? Not being able to contact the one thing this service is responsible for talking to feels like a good opportunity to fail fast to indicate to the operator a basic criteria for operation hasn’t been satisfied

madninja: you don't, and that's why the function returns a `Result` the introduced `expect` ends up crashing the program instead of returning an error

mobile_rewards/src/server.rs
vihu: Fixed in [656f9f4](https://github.com/helium/oracles/pull/115/commits/656f9f4d1984816f0f4cb7069a9bed68b81d25c9)

mobile_rewards/src/cli/generate.rs
vihu: [656f9f4](https://github.com/helium/oracles/pull/115/commits/656f9f4d1984816f0f4cb7069a9bed68b81d25c9)

poc_iot_verifier/src/traits/gateway_info_resolver.rs
vihu: > You should put the implementation for the follower service along side the gatewayinforesolver trait.

I updated this in 656f9f4

> Additionally, there's no reason to put these traits inside their own "traits" module. Move this module into the root. (You wouldn't have a "structs" module for example)

We should probably do another pass for this, not in this PR since that is happening in several other places .

poc_iot_verifier/src/follower.rs
```diff
@@ -1,24 +1,23 @@
...
-pub struct FollowerService {
-    client: FollowerClient,
+#[async_trait]
+impl GatewayInfoResolver for FollowerService {
+    async fn resolve_gateway_info(&mut self, address: &PublicKey) -> Result<FollowerGatewayResp> {
+        let req = FollowerGatewayReqV1 {
+            address: address.to_vec(),
+        };
+        let proto = self.client.find_gateway(req).await?.into_inner();
+        let res: FollowerGatewayResp = FollowerGatewayResp::try_from(proto).unwrap();
```
andymck, line 18: Should the FollowerGatewayResp struct and its various TryFrom's currently defined in poc_iot_verifier::follower not be moved to the new follower service itself ?  Perhaps other oracles currently do not require the full resp payload but still they could avail of it and take what they need from it

vihu: That's a fair point. I propose we add that to the node-follower once another service has the need to use the `FollowerGatewayResp` struct.

jeffgrunewald: the density_scaler, and by extension the iot verifier will need this

vihu: Ah fair enough, I'll move it over to node-follower then.

vihu: Update [1518ee8](https://github.com/helium/oracles/pull/115/commits/1518ee8df38a015f53d850492c3ce43765a3159c)

node_follower/src/error.rs
```diff
@@ -0,0 +1,19 @@
...
+pub type Result<T = ()> = std::result::Result<T, Error>;
+
+#[derive(Error, Debug)]
+pub enum Error {
+    #[error("uri error")]
+    Uri(#[from] http::uri::InvalidUri),
+    #[error("grpc {}", .0.message())]
+    Grpc(#[from] tonic::Status),
+    #[error("custom error")]
+    Custom(String),
```
jeffgrunewald, line 12: when would we use this custom error now that we're not calling the follower directly but instead putting it in a sub-crate?

also, do we want to de-dupe these errors from the error enums in the other apps? you've added the `FollowerError` to the error enum in iot verifier and mobile rewards server but they still have the grpc error there as well

node_follower/src/lib.rs
```diff
@@ -0,0 +1,58 @@
...
+const CONNECT_TIMEOUT: Duration = Duration::from_secs(5);
+const RPC_TIMEOUT: Duration = Duration::from_secs(5);
+const DEFAULT_URI: &str = "http://127.0.0.1:8080";
+
+type FollowerClient = follower::Client<Channel>;
+
+#[derive(Debug, Clone)]
+pub struct FollowerService {
+    pub client: FollowerClient,
+}
```
jeffgrunewald, line 19:
```suggestion
#[derive(Debug, Clone)]
pub struct FollowerClient {
    pub client: follower::Client<Channel>,
}
```
especially now that this is being abstracted behind a sub-crate i think it would be helpful to reaffirm this is just a client for a "remote" grpc server and it doesn't really operate like a "service" in the sense of a looping thread/process?

andymck (approved):

node_follower/src/gateway_resp.rs
```diff
@@ -0,0 +1,57 @@
...
+    fn try_from(v: FollowerGatewayRespV1) -> Result<Self> {
+        let staking_mode: GatewayStakingMode = GatewayStakingMode::from_i32(v.staking_mode)
+            .ok_or_else(|| Error::custom("unsupported staking_mode"))?;
+        let region: Region =
+            Region::from_i32(v.region).ok_or_else(|| Error::custom("unsupported region"))?;
+        let location = v
+            .location
+            .is_empty()
+            .not()
+            .then(|| v.location.parse::<u64>().unwrap());
```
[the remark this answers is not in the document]
vihu: Updated 4580f3a

poc_iot_verifier/src/runner.rs
```diff
@@ -1,31 +1,35 @@
...
-    lora_valid_poc::LoraValidPoc, lora_witness_report::LoraWitnessIngestReport, FileType,
+    file_sink,
+    file_sink::MessageSender,
+    file_upload,
+    lora_beacon_report::LoraBeaconIngestReport,
+    lora_beacon_report::LoraBeaconReport,
+    lora_invalid_poc::LoraInvalidBeaconReport,
+    lora_invalid_poc::LoraInvalidWitnessReport,
+    lora_valid_poc::LoraValidBeaconReport,
+    lora_valid_poc::LoraValidPoc,
```
jeffgrunewald, line 19:
```suggestion
    lora_beacon_report::{LoraBeaconReport, LoraBeaconIngestReport},
    lora_invalid_poc::{LoraInvalidBeaconReport, LoraInvalidWitnessReport},
    lora_valid_poc::{LoraValidPoc, LoraValidBeaconReport},
```

jeffgrunewald (approved):

vihu:
@jeffgrunewald @madninja I was thinking if I should move `txn_service` in this workspace as well (I have a patch ready to commit here). Its pretty isolated and only mobile_rewards and injector would be using it. Both of which implement pretty much the same module. Thoughts?

jeffgrunewald:
> @jeffgrunewald @madninja I was thinking if I should move `txn_service` in this workspace as well (I have a patch ready to commit here). Its pretty isolated and only mobile_rewards and injector would be using it. Both of which implement pretty much the same module. Thoughts?

Why not?

vihu:
> > @jeffgrunewald @madninja I was thinking if I should move `txn_service` in this workspace as well (I have a patch ready to commit here). Its pretty isolated and only mobile_rewards and injector would be using it. Both of which implement pretty much the same module. Thoughts?
> 
> Why not?

Added 787e9b2

jeffgrunewald (approved):

jeffgrunewald:
I say ship it; it’ll make it much easier for me to update the proto version to the one that returns a “gateway not found” error which has additional levels to the proto struct that need to be handled; would love to handle that in only one place!

PR #118, opened 2022-10-07
save newly downloaded denylist to disk, load on startup

andymck:


denylist/src/denylist.rs
```diff
@@ -110,8 +104,42 @@ impl DenyList {
...
+pub fn filter_from_bin(bin: &Vec<u8>) -> Result<Xor32> {
+    if bin.is_empty() {
+        tracing::debug!("empty filter bin");
+        return Err(Error::InvalidBinary("empty filter bin".to_string()));
+    }
+    let mut buf: &[u8] = bin;
+    let _version = buf.get_u8();
+    let signature_len = buf.get_u16_le() as usize;
+    let signature = buf.copy_to_bytes(signature_len).to_vec();
+    let pubkey = PublicKey::from_str(PUB_KEY_B58).expect("failed to decode pub key");
```
madninja, line 117: Find all your uses of expect and get rid of them

denylist/src/denylist.rs
```diff
@@ -110,8 +104,42 @@ impl DenyList {
...
+    match pubkey.verify(buf, &signature) {
+        Ok(_) => {
+            tracing::info!("updating filter to latest");
+            let _serial = buf.get_u32_le();
+            Ok(bincode::deserialize::<Xor32>(buf).unwrap())
```
madninja, line 122: Same with unwraps. They’re a code small afaict

denylist/src/denylist.rs
```diff
@@ -110,8 +104,42 @@ impl DenyList {
...
 fn public_key_hash(public_key: &PublicKey) -> u64 {
     let mut hasher = XxHash64::default();
     hasher.write(&public_key.to_vec());
     hasher.finish()
 }
+
+pub fn save_local_filter_bin(bin: &Vec<u8>, path: &str) {
+    if let Some(parent) = path::PathBuf::from(path).parent() {
+        if fs::create_dir_all(parent).is_ok() {
+            _ = fs::write(path, &bin);
```
madninja, line 142: This surely can return errrors. What’s expected to happen if a file write fails?

denylist/src/denylist.rs
andymck: what so bad with expect ? I see it in use across all projects in this workspace

denylist/src/denylist.rs
andymck: this is still a draft PR so not finished with this yet...but i dont think it really matters much if the write fails...I should log it should it occur but an attempt is to write will occur again at the next change in the remote denylist file.

The saved file is only used as a backup during init and used only if github is down and unable to retrieve the remote. denylist.  

i think the approach here should be if the denylist cannot be retrieved from github and we have no saved file then the application should halt and prevent running without any denylist filter

madninja: that makes some sense.. and isn't evident here quite yet. When I saw this function I figured it was the lowest level "write this filter" function which I guess I was expecting to return a Result for some higher level to use to do the logic you describe

PR #120, opened 2022-10-08
Change aws provider to use configured first then fall back to default

bbalser:


file_store/src/file_store.rs
```diff
@@ -50,7 +50,7 @@ impl FileStore {
         default_region: impl ProvideRegion + 'static,
         bucket: impl Into<String>,
     ) -> Result<Self> {
-        let region_provider = RegionProviderChain::default_provider().or_else(default_region);
+        let region_provider = RegionProviderChain::first_try(default_region).or_default_provider();
```
madninja, line 53: probably reads better to have the `default_region` parameter just be `region` since it's not the default ?

bbalser: Fixed 4356cc5

madninja (approved):

PR #122, opened 2022-10-10
Test run fixes - WIP

andymck:
IGNORE

this is a WIP branch, where i am pushing fixes from issues encountered during verifier testing on mainnet

This fixes two issues observed during ongoing testing on mainnet:

1.  The incoming timestamps from ingestor are in nano secs, verifier was expecting seconds.  
2. Whilst verifier was iterating over a list of beacons to process, should one of those fail to verify, verifier would halt processing of all beacons at that point rather than moving to the next beacon iteration.

### DIFF
--- a/poc_iot_verifier/src/loader.rs
+++ b/poc_iot_verifier/src/loader.rs
@@ -13,7 +13,7 @@ use helium_proto::{
     Message,
 };
 
-use file_store::{datetime_from_epoch, FileStore, FileType};
+use file_store::{traits::MsgTimestamp, FileStore, FileType};
 use sha2::{Digest, Sha256};
 use sqlx::PgPool;
 use tokio::time;
@@ -164,7 +164,7 @@ impl Loader {
                     id_hash,
                     packet_data,
                     buf_as_vec,
-                    &datetime_from_epoch(event.timestamp),
+                    &event.timestamp.to_timestamp_nanos()?,
                     ReportType::Beacon,
                 )
                 .await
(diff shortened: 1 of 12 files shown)


bbalser (approved):

poc_iot_verifier/src/loader.rs
```diff
@@ -184,7 +184,7 @@ impl Loader {
                     id_hash,
                     packet_data,
                     buf_as_vec,
-                    &datetime_from_epoch(event.timestamp),
+                    &event.timestamp.to_timestamp_nanos()?,
```
jeffgrunewald, line 187: why are we doing witness ingest report to nanos but beacon ingest to millis?

poc_iot_verifier/src/loader.rs
```diff
@@ -196,7 +196,7 @@ impl Loader {
                     &self.pool,
                     &id,
                     &event.data,
-                    &datetime_from_epoch(event.timestamp),
+                    &event.timestamp.to_timestamp()?,
```
jeffgrunewald, line 199: and here we're just going to the default timestamp units? shouldn't these all be to the same level of precision (presumably millis or nanos)?

poc_iot_verifier/src/loader.rs
andymck: this branch is stale, fixed in latest

poc_iot_verifier/src/loader.rs
andymck: as above, branch is stale, fixed in latest

PR #129, opened 2022-10-12
switch signal from int to float

andymck:


andymck:
decision was to keep signal as a float

PR #131, opened 2022-10-12
Separate out the verification and reward periods in the verifier 

maplant:
Additionally, heartbeat are now stored in a database to prevent gamification. 

This design should be a fair amount cleaner than the previous. 

Remaining tasks:
- [x] Write out `shares` and `invalid_shares` to file store
- [x] Figure out how to properly add a generate cli (does it write to the database?)
- [x] Select between the running services and shutdown.


madninja:
If I read it right the heartbeats table collects heartbeats for the rewardable period. I expect the heartbeats table to get cleaned up at the and of that period and not accumulate data forever since that'll end up hurting us down the line maintenance wise

poc_mobile_verifier/src/cli/generate.rs
```diff
@@ -18,37 +16,61 @@ pub struct Cmd {
     #[clap(long)]
     before: NaiveDateTime,
     #[clap(long)]
-    input_bucket: String,
+    periods: i32,
```
jeffgrunewald, line 19: what are `periods`? can we document this with a comment if there's no descriptive type that better conveys what this is? is it a number of hours? an index of time within a duration?

jeffgrunewald (commented):
is it just me or is the `before` and `after` hard for anyone else to grok? i know when i actually think about it we want a time period that falls `after_this_time -- event -- before_this_period` but looking at it in passing i always think `before -- after` based on the semantics of the word; would love to standardize on something less relative like `start -- event -- end` or `beginning -- event -- end`

poc_mobile_verifier/src/cli/generate.rs
```diff
@@ -18,37 +16,61 @@ pub struct Cmd {
...
+        let mut now = after;
+
+        for period in 0..periods {
+            let verify_epoch = verifier.get_verify_epoch(now);
+            tracing::info!("Verifying epoch {period} of {periods}...");
+            let _ = verifier
+                .verify_epoch(&mut transaction, verify_epoch)
+                .await?;
+            now += verification_period;
+        }
```
jeffgrunewald, line 65: would this for loop with an external mutable index be better expressed as an iter-fold over the periods?

maplant:
> If I read it right the heartbeats table collects heartbeats for the rewardable period. I expect the heartbeats table to get cleaned up at the and of that period and not accumulate data forever since that'll end up hurting us down the line maintenance wise

Indeed it does, we `TRUNCATE TABLE` at the end of the rewards period

poc_mobile_verifier/src/cli/generate.rs
```diff
@@ -18,37 +16,61 @@ pub struct Cmd {
...
+
+        tracing::info!("Calculating rewards...");
+
+        let rewards = verifier
+            .reward_epoch(&mut transaction, after..before)
+            .await?;
+
+        transaction.rollback().await?;
```
jeffgrunewald, line 73: At the very least can we add a comment here as to why we built this transaction and then just unrolled it? I get why we do it here in the cli but it feels like that's showing a coupling between the verifier and the database that should be less restrictive. Perhaps later the verifier can instead take anything here that implements a "checkpoint trait" and then the cli can just pass in an empty impl or if you're actually using the transaction to store state the verifier uses to _calculate_ the reward_epoch then the cli impl could use a simpler in-memory data structure in place of the db transaction?

poc_mobile_verifier/src/cli/server.rs
```diff
@@ -21,7 +34,81 @@ impl Cmd {
...
+                .map_err(Error::from),
+            subnet_sink.run(&shutdown_listener).map_err(Error::from),
+            file_upload.run(&shutdown_listener).map_err(Error::from),
+            // I don't _think_ that this needs to be in a task.
+            verifier_daemon.run(&shutdown_listener),
+        )?;
+
+        tracing::info!("Shutting down verifier server");
+
+        // run_server(pool, shutdown_listener).await?;
```
jeffgrunewald, line 111: delete?

poc_mobile_verifier/src/heartbeats.rs
```diff
@@ -0,0 +1,51 @@
...
+use helium_crypto::PublicKey;
+use rust_decimal::Decimal;
+use sqlx::{Postgres, Transaction};
+use std::collections::HashMap;
+
+#[derive(sqlx::FromRow)]
+pub struct Heartbeat {
+    pub id: PublicKey,
+    pub weight: Decimal,
+    pub timestamp: NaiveDateTime,
```
jeffgrunewald, line 14: this seems redundant to the `HeartbeatValue` struct below; are we not just making this struct `Heartbeat { pub id: PublicKey, weight: HearbeatValue }` because it wouldn't convert cleanly to a sql row?

poc_mobile_verifier/src/lib.rs
```diff
@@ -2,14 +2,15 @@ mod cell_type;
 mod error;
 mod mobile;
 mod reward_share;
-mod reward_speed_share;
+// mod reward_speed_share;
```
jeffgrunewald, line 5:
```suggestion
```

poc_mobile_verifier/src/cli/server.rs
```diff
@@ -1,6 +1,19 @@
...
+};
+use file_store::{file_sink, file_upload, FileStore, FileType};
+use futures_util::TryFutureExt;
+use helium_proto::services::{follower, Endpoint, Uri};
 use sqlx::postgres::PgPoolOptions;
 
+use super::{CONNECT_TIMEOUT, DEFAULT_URI, RPC_TIMEOUT};
+
+pub const DEFAULT_REWARD_PERIOD_HOURS: i64 = 24;
+pub const DEFAULT_VERIFICATIONS_PER_PERIOD: i32 = 8;
```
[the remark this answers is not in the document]
maplant: No stylistic or performance reasons, these are just the types that `chrono` expects - `Duration::hours` takes an `i64`, and Durations can only be divided by `i32`.

poc_mobile_verifier/src/shares.rs
```diff
@@ -0,0 +1,127 @@
...
+            });
+
+        Ok(shares)
+    }
+
+    pub async fn write(
+        self,
+        valid_shares_tx: &file_sink::MessageSender,
+        invalid_shares_tx: &file_sink::MessageSender,
+    ) -> Result<()> {
```
jeffgrunewald, line 58: is this redundant when you've imported the crate-local `Result` type?

poc_mobile_verifier/src/subnetwork_rewards.rs
```diff
@@ -92,600 +49,30 @@ impl SubnetworkRewards {
...
-
-    #[async_trait]
-    impl OwnerResolver for FixedOwnerResolver {
-        async fn resolve_owner(&mut self, _address: &PublicKey) -> Result<Option<PublicKey>> {
-            Ok(Some(self.owner.clone()))
-        }
-    }
-
-    #[tokio::test]
-    async fn check_rewards() {
```
jeffgrunewald, line 363: can these tests not be salvaged and preserved, even with some conversion?

poc_mobile_verifier/src/verifier.rs
```diff
@@ -0,0 +1,197 @@
...
+impl VerifierDaemon {
+    pub async fn run(mut self, shutdown: &triggered::Listener) -> Result {
+        tracing::info!("Starting verifier service");
+
+        let reward_period = Duration::hours(self.reward_period_hours);
+        let verification_period = reward_period / self.verifications_per_period;
+
+        loop {
+            let now = Utc::now();
+            // Maybe name these "epoch_since_*" and "epoch_since_*_duration"
```
jeffgrunewald, line 35: i like this; `get_*_epoch` implies a beginning and end and both of those aren't clear when you only have `now` as the frame of reference; "is it from some point in the past until now or is it from now until some point in the future?"

poc_mobile_verifier/src/verifier.rs
```diff
@@ -0,0 +1,192 @@
...
+        let transaction = self.pool.begin().await?;
+
+        let shares: Shares = self.verifier.verify_epoch(&epoch).await?;
+
+        // Should we remove the heartbeats that were not new
+        // from valid shares
+        for share in shares.valid_shares.clone() {
+            let heartbeat = Heartbeat::from(share);
+            heartbeat.save(&self.pool).await?;
+        }
```
jeffgrunewald, line 90: if you're doing these all at once why not do a bulk insert or insert in batches?

poc_mobile_verifier/src/verifier.rs
```diff
@@ -0,0 +1,192 @@
...
+
+        // Update the last verified end time:
+        self.verifier
+            .last_verified_end_time
+            .update(&self.pool, epoch.end.timestamp() as i64)
+            .await?;
+
+        transaction.commit().await?;
```
jeffgrunewald, line 98: i'm confused about this `transaction` now; what is it even doing? i see you creating and committing it but you're not passing it around anywhere so is it even being used?

poc_mobile_verifier/src/shares.rs
```diff
@@ -0,0 +1,126 @@
+//! Heartbeat storage
+
+use crate::Result;
+use chrono::{DateTime, Utc};
+use file_store::heartbeat::CellHeartbeat;
+use file_store::{file_sink, FileStore};
+use helium_proto::services::poc_mobile as proto;
+use std::ops::Range;
+
+use crate::cell_type::CellType;
```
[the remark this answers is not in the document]
maplant: No, I just assumed that cargo fmt would do it :-/

poc_mobile_verifier/src/subnetwork_rewards.rs
maplant: they are preserved, they got moved to `reward_share.rs` (which I think is a more appropriate place for them)

poc_mobile_verifier/src/shares.rs
maplant: yup

poc_mobile_verifier/src/verifier.rs
```diff
@@ -0,0 +1,192 @@
...
+impl VerifierDaemon {
+    pub async fn run(mut self, shutdown: &triggered::Listener) -> Result {
+        tracing::info!("Starting verifier service");
+
+        let reward_period = Duration::hours(self.reward_period_hours);
+        let verification_period = reward_period / self.verifications_per_period;
+
+        loop {
+            let now = Utc::now();
+            // Maybe name these "epoch_since_*" and "epoch_since_*_duration"
```
jeffgrunewald, line 35:
```suggestion
```

poc_mobile_verifier/src/verifier.rs
maplant: I don't think that's actually possible with sqlx: 
https://github.com/launchbadge/sqlx/issues/294

poc_mobile_verifier/src/cli/generate.rs
maplant: a better name would be `number_of_verification_periods`, so it is the number of verification epochs for the given duration

poc_mobile_verifier/src/verifier.rs
maplant: I think this is different... update_all is updating all entries with a given set of ids to the same timestamp. We don't necessarily want to update all of these entries to the same values, so in order to use this method, we'd have to group up our entries that are being set to the same value, which kinda just becomes the same thing we're doing now but with more pre-processing

jeffgrunewald: granted, that's a simplistic case b/c you're bulk updating the records that match the `any` clause with the same values, but skimming that issue it looks like there are a couple of ways sqlx will let you do it that aren't even very hacky anymore (UNNEST, ToRow, QueryBuilder); seems like a good opportunity to throw a `TODO` in here and revisit it when the thing is up and running?

maplant: Yeah, a todo would make sense here

poc_mobile_verifier/migrations/3_heartbeats.sql
```diff
@@ -0,0 +1,5 @@
+create table heartbeats (
+       id bytea primary key not null,
+       weight decimal not null,
```
vihu, line 3: nit: Is the `weight` reward_weight? If so can we rename it to `reward_weight` plz?

poc_mobile_verifier/src/cli/generate.rs
```diff
@@ -18,37 +21,62 @@ pub struct Cmd {
     #[clap(long)]
     before: NaiveDateTime,
     #[clap(long)]
-    input_bucket: String,
+    periods: i32,
```
vihu, line 24: Maybe add some docstring for what the `periods` is supposed to represent

poc_mobile_verifier/src/error.rs
```diff
@@ -32,6 +32,12 @@ pub enum Error {
     MetaError(#[from] db_store::MetaError),
     #[error("join error")]
     JoinError(tokio::task::JoinError),
+    #[error("time out of range error")]
+    OutOfRangeError,
```
vihu, line 36: Perhaps useful to have `OutOfRangeError(String)` and then update the error msg with the value which made it out of range?

poc_mobile_verifier/src/cli/generate.rs
maplant: has been removed, so resolving

poc_mobile_verifier/src/cli/generate.rs
maplant: removed so resolving

poc_mobile_verifier/src/error.rs
maplant: I don't think this error will ever occur, and unfortunately I have no way of actually typing the error type. I'd like to fix this properly by using `anyhow`

poc_mobile_verifier/src/verifier.rs
```diff
@@ -0,0 +1,176 @@
...
+            .update(&mut transaction, epoch.end.timestamp() as i64)
+            .await?;
+
+        transaction.commit().await?;
+
+        rewards
+            .write(&self.subnet_rewards_tx)
+            .await?
+            // Await the returned one shot to ensure that we wrote the file
+            .await??;
```
jeffgrunewald, line 134: just verifying this isn't a merge confict resolution gone wrong?

file_store/src/file_sink.rs
```diff
@@ -28,16 +28,28 @@ fn transport_sink(transport: &mut Transport) -> &mut Sink {
...
     mpsc::channel(size)
 }
 
-pub async fn write<T: prost::Message>(tx: &MessageSender, item: T) -> Result {
-    let buf = item.encode_to_vec();
-    tx.send(buf).await.map_err(|_| Error::channel())
+pub async fn write<T: prost::Message>(
+    tx: &MessageSender,
+    item: T,
+) -> Result<oneshot::Receiver<Result>> {
```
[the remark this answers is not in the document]
maplant: Yeah, I think the writer should be able to choose whether or not it waits for the write.

poc_mobile_verifier/migrations/3_heartbeats.sql
```diff
@@ -0,0 +1,5 @@
+create table heartbeats (
+       id text primary key not null,
+       reward_weight decimal not null,
+       timestamp timestamp not null
```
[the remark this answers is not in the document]
maplant: I think `timestamp` is more appropriate, it's always going to be UTC

poc_mobile_verifier/src/verifier.rs
maplant: Yup, this is correct. One result for awaiting the oneshot, one result for if the file was written properly.

poc_mobile_verifier/src/cli/generate.rs
```diff
@@ -1,14 +1,17 @@
...
 use chrono::{DateTime, NaiveDateTime, Utc};
 use file_store::FileStore;
 use helium_crypto::PublicKey;
 use helium_proto::services::{follower, Endpoint, Uri};
 use serde_json::json;
+use sqlx::postgres::PgPoolOptions;
+
+use super::{CONNECT_TIMEOUT, DEFAULT_URI, RPC_TIMEOUT};
+ */
+
```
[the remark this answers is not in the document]
maplant: This is outdated, it's all there right now

poc_mobile_verifier/src/verifier.rs
jeffgrunewald: What about the first `await?`

maplant: That is for sending the bytes over the channel to the file sink

jeffgrunewald (approved):

bbalser (approved):

PR #132, opened 2022-10-13
General timestamp trait

madninja:


mobile_rewards/src/server.rs
```diff
@@ -441,7 +434,7 @@ async fn reward_period(client: &mut follower::Client<Channel>) -> Result<Range<u
 }
 
 pub fn get_time_range(last_reward_end_time: i64) -> (DateTime<Utc>, DateTime<Utc>) {
-    let after_utc = datetime_from_epoch(last_reward_end_time);
+    let after_utc = Utc.timestamp(last_reward_end_time, 0);
```
jeffgrunewald, line 437: why not use TimestampDecode for this and either impl for i64 or else just cast `as u64`?

jeffgrunewald (commented):
I'm on the fence about the parameterized traits with the same method names returning different types (`u64` vs `Result<DateTime<Utc>>`). I get that it's clear what value each impl returns at the point the trait is implemented, but then at the callsite, especially when types are being elided and you just call `Thing.timestamp()` I don't know what format the return type is and I have to go hunt down the definition of `Thing` and the definition of `impl MsgTimestamp<> for Thing` to know if I'm expecting one timestamp format or another. Is this just me?

madninja:
> I'm on the fence about the parameterized traits with the same method names returning different types (`u64` vs `Result<DateTime<Utc>>`). I get that it's clear what value each impl returns at the point the trait is implemented, but then at the callsite, especially when types are being elided and you just call `Thing.timestamp()` I don't know what format the return type is and I have to go hunt down the definition of `Thing` and the definition of `impl MsgTimestamp<> for Thing` to know if I'm expecting one timestamp format or another. Is this just me?

See that's the thing you don't have to know... you need a timestamp of the right type .. just call timestamp on the thing you're getting it from

mobile_rewards/src/server.rs
madninja: Because this was the _only_ place that would need that.. and casting as u64 looses the negative timestamp ability which by the contract of get_time_range is allowed

jeffgrunewald: that makes sense; i just figured it was because postgres expects them to be i64s

jeffgrunewald:
> See that's the thing you don't have to know... you need a timestamp of a given type? .. just call timestamp on the thing you're getting it from
that's not _entirely_ true; i need to know if i need to unwrap the result when it's a `Result<DateTimeUtc>>` vs. just letting it fly when it's a `u64` but generally speaking this is mostly just an adjustment and it does mean you're in for a lot of red on your screen when you go to compile if you don't prefer the noise of type hints from the rust-analyzer running in real time as you code

jeffgrunewald (approved):

madninja:
> > See that's the thing you don't have to know... you need a timestamp of a given type? .. just call timestamp on the thing you're getting it from
> 
> that's not _entirely_ true; i need to know if i need to unwrap the result when it's a `Result<DateTimeUtc>>` vs. just letting it fly when it's a `u64` but generally speaking this is mostly just an adjustment and it does mean you're in for a lot of red on your screen when you go to compile if you don't prefer the noise of type hints from the rust-analyzer running in real time as you code.
> 
> i also get why you can just return a `u64` when that's what you need but a `DateTime<Utc>` has the potential to fail to convert so you need to wrap that in a result; you could solve that by just _always_ wrapping them in a result but i understand if that's not idiomatic since the `u64` is sort of an infallible output

Yeah .. I did consider using `Result<R>` instead and want to do that as a separate PR because it affects a bunch of `From` impls

vihu (approved):

PR #135, opened 2022-10-13
Switch to for_each_concurrent for processing

vihu:
Summary
----
This doesn't change any functionality just switches stream processing to be more concurrent.

### DIFF
--- a/poc_iot_injector/src/cli/generate.rs
+++ b/poc_iot_injector/src/cli/generate.rs
@@ -1,7 +1,7 @@
 use crate::{
     keypair::{load_from_file, Keypair},
     receipt_txn::handle_report_msg,
-    Result,
+    Result, LOADER_WORKERS, STORE_WORKERS,
 };
 use chrono::{NaiveDateTime, TimeZone, Utc};
 use file_store::{FileStore, FileType};
@@ -31,7 +31,6 @@ impl Cmd {
             .list_all(FileType::LoraValidPoc, after_utc, before_utc)
             .await?;
 
-        let stream = store.source(stream::iter(file_list).map(Ok).boxed());
         let before_ts = before_utc.timestamp_millis();
 
         let poc_injector_kp_path =
@@ -39,14 +38,14 @@ impl Cmd {
         let poc_oracle_key = load_from_file(&poc_injector_kp_path)?;
         let shared_key = Arc::new(poc_oracle_key);
 
-        stream
-            .for_each_concurrent(10, |msg| {
+        store
+            .source_unordered(LOADER_WORKERS, stream::iter(file_list).map(Ok).boxed())
+            .for_each_concurrent(STORE_WORKERS, |msg| {
                 let shared_key_clone = shared_key.clone();
                 async move {
-                    if let Ok(m) = msg {
-                        process_msg(m, shared_key_clone, before_ts).await;
-                    } else {
-                        tracing::error!("unable to process msg: {:?}", msg)
+                    match msg {
+                        Ok(m) => process_msg(m, shared_key_clone, before_ts).await,
+                        Err(e) => tracing::error!("unable to process msg due to: {:?}", e),
                     }
                 }
             })
@@ -58,12 +57,14 @@ impl Cmd {
 
 async fn process_msg(msg: prost::bytes::BytesMut, shared_key_clone: Arc<Keypair>, before_ts: i64) {
     if let Ok(Some((txn, _hash, _hash_b64_url))) =
-        handle_report_msg(msg, shared_key_clone, before_ts)
+        handle_report_msg(msg.clone(), shared_key_clone, before_ts)
     {
         let tx = BlockchainTxn {
             txn: Some(Txn::PocReceiptsV2(txn)),
         };
 
         tracing::debug!("txn_bin: {:?}", tx.encode_to_vec());
+    } else {
+        tracing::error!("unable to construct txn for msg {:?}", msg)
     }
 }
--- a/poc_iot_injector/src/cli/generate.rs
+++ b/poc_iot_injector/src/cli/generate.rs
@@ -1,8 +1,13 @@
-use crate::{keypair::load_from_file, receipt_txn::handle_report_msg, Result};
+use crate::{
+    keypair::{load_from_file, Keypair},
+    receipt_txn::handle_report_msg,
+    Result,
+};
 use chrono::{NaiveDateTime, TimeZone, Utc};
 use file_store::{FileStore, FileType};
 use futures::stream::{self, StreamExt};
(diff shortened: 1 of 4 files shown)


vihu:
I could use some advice on reducing the if-let chaining done in the closures. cc: @maplant

poc_iot_injector/src/server.rs
```diff
@@ -19,29 +18,30 @@ pub const DEFAULT_LAST_POC_SUBMISSION_TS_SECS: i64 = 0000000000;
...
-            .unwrap_or_else(|_| DEFAULT_LAST_POC_SUBMISSION_TS_SECS.to_string())
-            .parse()
-            .map_err(DecodeError::from)?;
+        // Check meta for last_poc_submission_ts, if not found, use the env var and insert it
+        let last_poc_submission_ts =
+            MetaValue::<i64>::fetch_or_insert_with(&pool, "last_reward_end_time", || {
+                env::var("LAST_POC_SUBMISSION_TS")
+                    .unwrap_or_else(|_| DEFAULT_LAST_POC_SUBMISSION_TS_SECS.to_string())
+                    .parse()
+                    .unwrap_or_default()
```
andymck, line 35: would erroring out here not be more appropriate than defaulting to type default ?

andymck (approved):
lg, i will see if i can apply the same concurrency model to my POC verification on iot verifier

poc_iot_injector/src/server.rs
xandkar: +1 - type default is a smell and kind of a dubious concept in general. If anything, use an explicit, domain-specific default value.

poc_iot_injector/src/cli/generate.rs
```diff
@@ -26,24 +31,39 @@ impl Cmd {
...
-
-                tracing::debug!("txn_bin: {:?}", tx.encode_to_vec());
-            }
-        }
         Ok(())
     }
 }
+
+async fn process_msg(msg: prost::bytes::BytesMut, shared_key_clone: Arc<Keypair>, before_ts: i64) {
+    if let Ok(Some((txn, _hash, _hash_b64_url))) =
```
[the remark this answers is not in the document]
vihu: Updated 00c37db

poc_iot_injector/src/server.rs
vihu: Updated 00c37db. cc: @madninja there's an expect here now, I think its okay to have it since we do want it to blow up if the `LAST_POC_SUBMISSION` is not found or is set incorrectly.

poc_iot_injector/src/server.rs
```diff
@@ -99,36 +102,56 @@ impl Server {
...
+}
+
+async fn handle_txn_submission(
+    msg: prost::bytes::BytesMut,
+    shared_key: Arc<Keypair>,
+    shared_txn_service: Arc<Mutex<TransactionService>>,
+    before_ts: i64,
+) {
+    if let Ok(Some((txn, hash, hash_b64_url))) = handle_report_msg(msg, shared_key, before_ts) {
+        let mut txn_service = shared_txn_service.lock().await;
```
madninja, line 141: One of the questions I have is why we need a lock on the txn service to serialize txns? The concurrency now means they could already be unordered. 

1. Is the client concurrency safe? if so you don't need a lock
2. If it's not safe you could just create a client per txn? That way the concurrency isn't serialized through a single lock

poc_iot_injector/src/server.rs
```diff
@@ -99,36 +102,56 @@ impl Server {
...
     after_utc: DateTime<Utc>,
     before_utc: DateTime<Utc>,
 ) -> Result {
     let file_list = store
         .list_all(FileType::LoraValidPoc, after_utc, before_utc)
         .await?;
 
-    let mut stream = store.source(stream::iter(file_list).map(Ok).boxed());
-    let ts = before_utc.timestamp_millis();
+    let stream = store.source(stream::iter(file_list).map(Ok).boxed());
```
madninja, line 115: If you really don't care in which order the files themselves  are processed either (separate from the ordering within the file) you could use `store.source_unordered` which will load files concurrently too

poc_iot_injector/src/server.rs
madninja: Thanks for not using expect 👍🏻

poc_iot_injector/src/server.rs
vihu: Yeah, I don't think the order of processing matters as long as it is done for the batch after the `last_poc_submission_ts`. I changed it to `store.source_unordered` here 838418c

poc_iot_injector/src/server.rs
```diff
@@ -99,36 +102,56 @@ impl Server {
...
-            {
-                tracing::debug!("txn submitted successfully, hash: {:?}", hash_b64_url);
-            } else {
-                tracing::warn!("txn submission failed!, hash: {:?}", hash_b64_url);
+    stream
+        .for_each_concurrent(10, |msg| {
+            let shared_txn_service = txn_service.clone();
+            let shared_key = keypair.clone();
+            async move {
+                if let Ok(m) = msg {
```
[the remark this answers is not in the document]
vihu: Addressed in 838418c

poc_iot_injector/src/server.rs
vihu: Hmm, fair enough. I mostly just did it this way to get around the rust compiler tbh. So if I understand this correctly, we just clone away txn service for each txn submission? Or create new one via `TransactionService::from_env` and fire it?

madninja: I believe you can just clone tonic clients since they hold very little state

poc_iot_injector/src/server.rs
```diff
@@ -17,29 +20,32 @@ pub const DEFAULT_LAST_POC_SUBMISSION_TS_SECS: i64 = 0000000000;
...
     txn_service: TransactionService,
     iot_verifier_store: FileStore,
     last_poc_submission_ts: MetaValue<i64>,
 }
 
 impl Server {
     pub async fn new(pool: Pool<Postgres>, keypair: Keypair) -> Result<Self> {
-        // TODO: Find this time in meta store db first
-        let last_poc_submission_ts = env::var("LAST_POC_SUBMISSION_TS")
+        let env_last_poc_submission_ts = env::var("LAST_POC_SUBMISSION_TS")
```
jeffgrunewald, line 31: @JayKickliter had the good suggestion to me earlier that ambiguous envvars (poc being a pretty broad category for us) might want app or network name spacing; avoid future conflicts with any other services running that might want a similar environment variable?

madninja: Please see https://github.com/helium/oracles/pull/142 for where this is going

PR #136, opened 2022-10-14
misc fixes from mainnet testing

andymck:
Includes a bunch of fixes identified from testing with real data on mainnet.  Also some refactor to centralise report ID generation

### DIFF
--- a/file_store/src/lora_beacon_report.rs
+++ b/file_store/src/lora_beacon_report.rs
@@ -8,7 +8,6 @@ use helium_crypto::PublicKey;
 use helium_proto::services::poc_lora::{LoraBeaconIngestReportV1, LoraBeaconReportReqV1};
 use helium_proto::DataRate;
 use serde::Serialize;
-use sha2::{Digest, Sha256};
 
 #[derive(Serialize, Clone, Debug)]
 pub struct LoraBeaconReport {
@@ -33,26 +32,6 @@ pub struct LoraBeaconIngestReport {
 
 impl MsgDecode for LoraBeaconIngestReport {
     type Msg = LoraBeaconIngestReportV1;
-}
-
-impl LoraBeaconIngestReport {
-    pub fn generate_id(&self) -> Vec<u8> {
-        let mut id: Vec<u8> = self.report.data.clone();
-        let mut public_key = self.report.pub_key.to_vec();
-        id.append(&mut self.received_timestamp.to_string().as_bytes().to_vec());
-        id.append(&mut public_key);
-        Sha256::digest(&id).to_vec()
-    }
-}
-
-impl LoraBeaconReport {
-    pub fn generate_id(&self, received_ts: DateTime<Utc>) -> Vec<u8> {
-        let mut id: Vec<u8> = self.data.clone();
-        let mut public_key = self.pub_key.to_vec();
-        id.append(&mut received_ts.to_string().as_bytes().to_vec());
-        id.append(&mut public_key);
-        Sha256::digest(&id).to_vec()
-    }
 }
 
 impl TryFrom<LoraBeaconReportReqV1> for LoraBeaconIngestReport {
--- a/file_store/src/lora_beacon_report.rs
+++ b/file_store/src/lora_beacon_report.rs
@@ -8,8 +8,9 @@ use helium_crypto::PublicKey;
 use helium_proto::services::poc_lora::{LoraBeaconIngestReportV1, LoraBeaconReportReqV1};
 use helium_proto::DataRate;
 use serde::Serialize;
+use sha2::{Digest, Sha256};
 
-#[derive(Serialize, Clone)]
+#[derive(Serialize, Clone, Debug)]
 pub struct LoraBeaconReport {
     #[serde(alias = "pubKey")]
     pub pub_key: PublicKey,
@@ -24,7 +25,7 @@ pub struct LoraBeaconReport {
     pub signature: Vec<u8>,
 }
 
-#[derive(Serialize, Clone)]
+#[derive(Serialize, Clone, Debug)]
 pub struct LoraBeaconIngestReport {
     pub received_timestamp: DateTime<Utc>,
     pub report: LoraBeaconReport,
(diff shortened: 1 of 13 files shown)


poc_iot_verifier/src/meta.rs
```diff
@@ -58,12 +58,9 @@ impl Meta {
         .await?
         .and_then(|v| {
             v.parse::<u64>()
-                .map_or_else(|_| None, |secs| Some((secs + 10).to_timestamp()))
-            //TODO: remove the hardcoded + 10 above
-            //      fix resolution of datetime function, dropping millisecs resulting in files being reprocessed by loaded
-        })
-        .transpose()?;
-        Ok(height)
+                .map_or_else(|_| None, |ts| Some(Utc.timestamp_millis(ts as i64)))
```
madninja, line 61: Use the `TimestampDecode` trait for u64

poc_iot_verifier/src/poc.rs
```diff
@@ -112,8 +109,11 @@ impl Poc {
         let beacon_received_time = self.beacon_report.received_timestamp;
         if beacon_received_time < self.entropy_start || beacon_received_time > self.entropy_end {
             tracing::debug!(
-                "beacon verification failed, reason: {:?}",
-                InvalidReason::BadEntropy
+                "beacon verification failed, reason: {:?}. beacon_received_time: {:?}, entropy_start_time: {:?}, entropy_end_time: {:?}",
+                InvalidReason::BadEntropy,
+                beacon_received_time,
```
madninja, line 114: you know you can inline these single variable names right? i.e. in the format string you can `{beacon_received_time:?}` and remove this positional argument.

poc_iot_verifier/src/poc.rs
```diff
@@ -344,7 +360,12 @@ impl Poc {
...
         let min_rcv_signal =
             calc_fspl(tx_power, witness.frequency, witness_distance, gain).unwrap();
-        if witness.signal as f64 > min_rcv_signal {
+        tracing::debug!(
+            "signal: {:?}, min_rcv_signal: {:?}",
+            witness.signal,
+            min_rcv_signal
+        );
+        if witness.signal < min_rcv_signal as i32 {
             tracing::debug!(
```
madninja, line 369: I guess these will all turn into actual Error returns for some new WitnessValidationError or something?

andymck: yeah atm i have disabled the Error for RSSI check, until I get the rssi calculation updated to account for the signal being in ddbm

poc_iot_verifier/src/poc.rs
andymck: yip am aware, switch between the two for no valid reason

file_store/src/lora_beacon_report.rs
```diff
@@ -34,6 +35,26 @@ impl MsgDecode for LoraBeaconIngestReport {
     type Msg = LoraBeaconIngestReportV1;
 }
 
+impl LoraBeaconIngestReport {
```
jeffgrunewald, line 38: this looks like a great opportunity for an `ReportId` trait? maybe a TODO if it's not worth adding right now?

andymck: yeah was originally going to add as a trait but need to get it in...will keep it as an improvement

poc_iot_verifier/src/loader.rs
```diff
@@ -150,53 +152,46 @@ impl Loader {
...
+                tracing::debug!("beacon report from ingestor: {:?}", &beacon);
+                let packet_data = beacon.report.data.clone();
+                tracing::debug!("beacon data: {:?}", &packet_data);
                 Report::insert_into(
                     &self.pool,
-                    id_hash,
+                    beacon.generate_id(),
                     packet_data,
-                    buf_as_vec,
-                    &event.timestamp()?,
```
jeffgrunewald, line 167: why are we undoing the MsgTimestamp trait here?

poc_iot_verifier/src/poc.rs
```diff
@@ -10,15 +10,14 @@ use h3ron::{to_geo::ToCoordinate, H3Cell, H3DirectedEdge, Index};
...
-
 /// C is the speed of light in air in meters per second
 pub const C: f64 = 2.998e8;
 /// R is the (average) radius of the earth
 pub const R: f64 = 6.371e6;
 
 /// measurement in seconds of an entropy
 /// TODO: determine a sane value here, set high for testing
-const ENTROPY_LIFESPAN: i64 = 90000;
+const ENTROPY_LIFESPAN: i64 = 60;
```
jeffgrunewald, line 20: was this 90 seconds previously?

poc_iot_verifier/src/meta.rs
andymck: fixed here: 22bbdbe0ff321c5eeb1c4661758329296a028188

poc_iot_verifier/src/poc.rs
andymck: 60 is what i am using atm on mainnet

poc_iot_verifier/src/loader.rs
andymck: changed to using the received_timestamp  from the beacon report, this is already a DateTime as a result of the LoraBeaconIngestReportV1 -> LoraBeaconIngestReport at line 155.

file_store/src/traits/report_id.rs
```diff
@@ -0,0 +1,47 @@
...
+            fn report_id(&self, received_ts: DateTime<Utc>) -> Vec<u8> {
+                let mut id: Vec<u8> = self.data.clone();
+                let mut public_key = self.pub_key.to_vec();
+                id.append(&mut received_ts.to_string().as_bytes().to_vec());
+                id.append(&mut public_key);
+                Sha256::digest(&id).to_vec()
+            }
+        }
+    };
+}
```
jeffgrunewald, line 42: i'm not sure about the utility of these macros since they depend on the fields of the implementing structs pretty heavily; from what i can tell, it's more idiomatic to just put the `impl ReportId for ReportType` block, typically in the same file that defines the `ReportType` struct (whichever that might be)

poc_iot_verifier/src/runner.rs
```diff
@@ -196,7 +202,9 @@ impl Runner {
                 Some(res) => res,
                 None => {
                     tracing::debug!("beacon verification failed, reason: EntropyNotFound");
-                    return Ok(());
+                    _ = Report::update_attempts(&self.pool, &beacon_report.ingest_id(), Utc::now())
+                        .await;
```
jeffgrunewald, line 206: if we're not logging here each time we fail to validate the report are we logging the hash or timestamp or some other identifier for all of the reports when they finally get purged from the DB at max_failed_attempts?

file_store/src/traits/report_id.rs
andymck: well i mean thats the way it was previously.  this change was a patch created by vihu to move the ID generation out of the report `impl` and into their own trait.

jeffgrunewald (approved):

file_store/src/traits/report_id.rs
Vagabond: it'd be great to have the signature in here as well

madninja (approved):

PR #137, opened 2022-10-17
Clean up the mobile rewards server with stream combinators

maplant:


mobile_rewards/src/server.rs
```diff
@@ -299,66 +306,67 @@ impl Server {
             )
             .await?;
 
-        // TODO: We need a better pattern for this:
-        let mut stream = self
+        let new_reward_end_time = time_range.end.timestamp();
```
madninja, line 309: wasn't the new_reward_end_time the timestamp of the last processed verifier file to avoid it missing delayer verifier reports? Or are we ok dropping them?

jeffgrunewald: this is the same value it's always been (the end of the utc range converted to the unix timestamp); just grabbing it here before the range is moved into the async closures.
if you wanted to set it to the `end_epoch` on the last processed subnet rewards file is there a better way to do that then just setting it to `mut` and updating it on each element of the stream?

andymck (approved):

mobile_rewards/src/server.rs
madninja: this is not really a rust question but more a "are we doing the right thing" here. Perhaps we can hash it out on a call.

PR #141, opened 2022-10-19
heartbeats are tracked by radio and summed up to owner

bbalser:


jeffgrunewald (approved):

poc_mobile_verifier/migrations/4_heartbeats.sql
```diff
@@ -0,0 +1,9 @@
+drop table heartbeats;
+
+create table heartbeats (
+  pub_key bytea not null,
```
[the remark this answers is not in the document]
bbalser: Fixed ed03d55

poc_mobile_verifier/src/heartbeats.rs
```diff
@@ -43,75 +55,19 @@ impl Heartbeat {
...
-            heartbeats.insert(
-                id,
-                HeartbeatValue {
-                    reward_weight,
-                    timestamp,
-                },
-            );
-        }
-        Self { heartbeats }
+        iter.into_iter().map(Heartbeat::from).collect()
```
maplant, line 71: Since collect takes a FromIterator bound, you might as well entirely remove this impl and put this line wherever you want heartbeats

bbalser: Fixed 7dc32cc

maplant (approved):

jeffgrunewald (commented):
i don't understand what the compound types are buying us here if we're re-implementing the into-iter trait to just aggregate reward shares against the same heartbeat struct we were using already?

maplant:
> i don't understand what the compound types are buying us here if we're re-implementing the into-iter trait to just aggregate reward shares against the same heartbeat struct we were using already?

We need the same database logic (i.e. one heartbeat for (pubkey, cbsd_id)) elsewhere. For example, the verifier generate cli. Another bonus is that the input to the rewards epoch is required to mimic the same semantics as the database.

Effectively, the data structure is a mock database, and you cannot construct one that does not match the same semantics of the database.

jeffgrunewald:
> Effectively, the data structure is a mock database, and you cannot construct one that does not match the same semantics of the database.

the way I understand the database _records_ the single heartbeat struct already mirrors the records in the database but what you’re saying is you want the rust struct to mirror the virtual primary key index of the composite (pubkey, cbsid)?

maplant:
> 

I want subnetwork rewards to be able to express the following invariant: All of these heartbeats will have a unique (pubkey, cbsid)

jeffgrunewald:
Is that going to entail any pain to undo that invariant when you need to allow for multiple heartbeats per radio and the only unique restriction is the hotspot pubkey? In the database that would be as simple as updating the insert statement not to overwrite the previous timestamp and reward weight values but you’d have to then dismantle this constraint from the type system, no?

maplant:
> Is that going to entail any pain to undo that invariant when you need to allow for multiple heartbeats per radio and the only unique restriction is the hotspot pubkey? In the database that would be as simple as updating the insert statement not to overwrite the previous timestamp and reward weight values but you’d have to then dismantle this constraint from the type system, no?
> 
> that’s a “future problem” of course, but it’s on the order of weeks not months+ and I think optimizing for the future is a different argument when you’re advocating making things “simpler” for the future as opposed to adding extra work and complication that’s not needed immediately

In my mind it would still be a HashMap, even if the key changes from a tuple to just a pub key. Before this I still wanted to express the invariant that you wouldn't see multiple heartbeats with the same pubkey. 

Whatever invariants we encode in the database I would like to be expressed in this data structure. The thing that's difficult to change is the database, otherwise this is pretty malleable. Especially because we should only be changing the internal implementation details - everything else interacts with heartbeats via functions

vihu:
Looks ok to me, we should do a manual `generate` command with some old timestamps and check owner emissions though.

vihu (approved):

PR #142, opened 2022-10-19
mobile indexer template & config

madninja:
This adds the mobile indexer which is intended to track lifetime hotspot rewards for use by the Solana reward programs. 

The indexer will retrieve verified shares, and accumulate hotspot rewards (note: not shares) in a database for Solana programs to access. It will also emit output for hotspot rewards per attached cell  as it processes verified reward shares

- [x] settings entropy
- [x] settings ingest
- [x] file_store
- [x] node_follower
- [x] settings mobile_rewards
- [x] settings poc_iot_injector
- [x] settings poc_mobile_verifier
- [x] settings poc_iot_verifier

For each workspace member:

- [ ] update `*.service` file for `server` command line
- [ ] create settings-template.toml setup

Deployment

- [ ] Update ops/deployment scripts to install service specific settings file


mobile_index/src/settings.rs
```diff
@@ -0,0 +1,75 @@
...
+    /// URL to access the postgres database. For example:
+    /// =postgres://postgres:postgres@127.0.0.1:5432/mobile_index_db
+    #[serde(with = "http_serde::uri")]
+    pub url: Uri,
+}
+
+impl Settings {
+    /// Load Settings from a given path. Settings are loaded from a given
+    /// optional path and can be overriden with environment variables.
+    ///
```
jadeallenx, line 33: Config is expected in TOML? Where are the valid values enumerated? Is there a default / expected path for config files?

mobile_index/pkg/deb/mobile-index.service
```diff
@@ -0,0 +1,17 @@
+[Unit]
+Description=mobile-index
+After=network.target
+StartLimitInterval=60
+StartLimitBurst=3
+
+[Service]
+Type=simple
+ExecStart=/opt/mobile-mobile_index_db/bin/mobile-index
```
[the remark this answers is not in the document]
xandkar: what should it be?

mobile_index/src/settings.rs
madninja: Each attribute name in the Settings struct is either a value (like interval) or a section if it's a set of values.

madninja: @jadeallenx I added settings-template.toml example file with inline documentation

mobile_index/Cargo.toml
```diff
@@ -0,0 +1,52 @@
...
+version = "0.1.0"
+description = "Reward Indexer for the Helium Mobile Network"
+edition.workspace = true
+authors.workspace = true
+license.workspace = true
+
+
+[dependencies]
+config = {workspace = true}
+dotenv = {workspace = true}
```
xandkar, line 12: How did `dotenv` sneak back in there? I thought you removed its usage.

file_store/src/file_upload.rs
```diff
@@ -24,6 +24,13 @@ pub struct FileUpload {
 }
 
 impl FileUpload {
+    pub async fn from_settings(settings: &Settings, messages: MessageReceiver) -> Result<Self> {
+        Ok(Self {
+            messages: UnboundedReceiverStream::new(messages),
+            store: FileStore::from_settings(settings).await?,
+        })
+    }
+
```
jeffgrunewald, line 33: This `from_env` still necessary?

ingest/pkg/deb/settings-template.toml
```diff
@@ -0,0 +1,46 @@
+
+# log settings for the application (RUST_LOG format). Default below
+# 
+# log = "poc_ingest=debug,poc_store=info"
+
+# Mode to run the ingest in. This adjustst the allowed grpc endpoints. "lora" or
+# "mobile". Required
+mode = "lora"
+
+# Token for ingest grpc endpoint bearer authentication. This is required for
```
jeffgrunewald, line 10: Deprecated api token?

ingest/src/server_5g.rs
```diff
@@ -183,16 +175,19 @@ pub async fn grpc_server(shutdown: triggered::Listener, server_mode: String) ->
         speedtest_report_tx,
     )?;
 
-    let api_token = std::env::var("API_TOKEN").map(|token| {
-        format!("Bearer {}", token)
-            .parse::<MetadataValue<_>>()
-            .unwrap()
-    })?;
+    let api_token = settings
```
jeffgrunewald, line 178: I thought we’d removed all this api bearer token auth stuff for the ingest server?

jeffgrunewald: I guess this is just for the 5g server which _is_ still authenticated?

file_store/src/file_store.rs
```diff
@@ -6,16 +6,44 @@ use aws_sdk_s3::{types::ByteStream, Client, Endpoint, Region};
...
 use futures::{stream, StreamExt, TryFutureExt, TryStreamExt};
 use http::Uri;
+use serde::{Deserialize, Serialize};
 use std::str::FromStr;
 use std::{env, path::Path};
 
+#[derive(Debug, Serialize, Deserialize)]
+pub struct Settings {
+    /// Bucket name for the store. Required
+    pub bucket: String,
```
jeffgrunewald, line 16: How does the file store configure multiple buckets for services that have both an input and an output or even multiples of each? Like the iot verifier will have the input bucket for ingest reports, and output bucket for valid/invalid ingest reports, an output bucket for verified iot reward shares, and an output bucket indirectly managed for the density scaler’s periodic hex scaling factor global map?

mobile_index/src/pending_txn.rs
```diff
@@ -0,0 +1,190 @@
...
+                    completed_at = $2
+                where hash = any($3); 
+                "#
+            }
+        }
+    }
+}
+
+#[derive(sqlx::FromRow, Deserialize, Serialize, Debug)]
+#[sqlx(type_name = "pending_txn")]
```
jeffgrunewald, line 74: What even is a pending txn anymore on solana when you’re adding to the lifetime balance of a hotspot? Shouldn’t it either update the db immediately or else throw an error since the indexer is just adding the monotonically increasing reward total for each hotspot? And in the mobile world shouldn’t this be attached to the mobile radio instead of the hotspot since someone could sell or give away their radio but the lifetime rewards for that radio would stay with the hotspot right now?

mobile_index/src/indexer.rs
```diff
@@ -0,0 +1,102 @@
...
+        let file_list = self
+            .verifier_store
+            .list_all(
+                FileType::SubnetworkRewards,
+                time_range.start,
+                time_range.end,
+            )
+            .await?;
+
+        let _subnet_rewards: Stream<SubnetworkReward> = self
```
jeffgrunewald, line 70: This is still in WIP? This doesn’t appear to be doing anything with the verified rewards within the window

jeffgrunewald (commented):
It looks like multiple settings for the same component (file store for instance) want to have different namespaces within the calling application toml file and it’s up to the calling application to instantiate the correct one with the correct names settings based on its context? So indexer has “verified” settings and “output” settings each for an instance of file store and it’s up to the indexer to call into the correct section of its settings module when it creates the instance of each one?

mobile_index/Cargo.toml
madninja: yeah, odd.. removed and trimmed the mobile_index member source since it's not doing anything quite yet as we migrate config approach

ingest/pkg/deb/settings-template.toml
madninja: As you figured out a few comments later.. mobile data comes from a known source which we add an additional api token to for some added security. Not needed for lora since that's not quite single source data. 

I expect that api token to go away from mobile too as we add a verifier to verify existence of hotspots on chain

ingest/src/server_5g.rs
madninja: Correct

mobile_index/src/pending_txn.rs
madninja: Yup,, copy pasta from mobile_rewards. Cleared out a bunch of files now just to get a clean baseline while we do the config switchover. I'll fix up the schema for mobile_index as soon as that lands

mobile_index/src/indexer.rs
madninja: yup.. just initial code that doesn't do anything with the db yet.

madninja:
> It looks like multiple settings for the same component (file store for instance) want to have different namespaces within the calling application toml file and it’s up to the calling application to instantiate the correct one with the correct names settings based on its context? So indexer has “verified” settings and “output” settings each for an instance of file store and it’s up to the indexer to call into the correct section of its settings module when it creates the instance of each one?

that's exactly right..

node_follower/src/settings.rs
```diff
@@ -0,0 +1,56 @@
...
+pub fn default_rpc_timeout() -> u64 {
+    5
+}
+
+pub fn default_batch_size() -> u32 {
+    100
+}
+
+impl Settings {
+    pub fn connect_follower(&self) -> Result<follower::Client<Channel>> {
```
jeffgrunewald, line 41: do we need two separate settings since the follower will _always_ be the same node we're following the chain from as the one we're submitting transactions to? can the node follower crate here abstract that detail away for the calling rust code and just instantiate a single `follower` client that the app can use to submit transaction rpcs to _and_ follower rpcs to?

poc_iot_verifier/src/main.rs
```diff
@@ -1,38 +1,85 @@
...
+    cmd: Cmd,
+}
+
+impl Cli {
+    pub async fn run(self) -> Result {
+        let settings = Settings::new(self.config)?;
+        tracing_subscriber::registry()
+            .with(tracing_subscriber::EnvFilter::new(&settings.log))
+            .with(tracing_subscriber::fmt::layer())
+            .init();
```
jeffgrunewald, line 27: is this redundant with the one defined in the `server` subcommand? if you want the tracing subscriber here to allow logging to std when running one-off commands can we remove the one in the server subcommand?

poc_iot_verifier/src/runner.rs
```diff
@@ -38,15 +40,16 @@ const LOADER_DB_POOL_SIZE: usize = 2 * LOADER_WORKERS;
...
 }
 
 impl Runner {
-    pub async fn from_env() -> Result<Self> {
-        let pool = mk_db_pool(LOADER_DB_POOL_SIZE as u32).await?;
-        let store_path =
-            std::env::var("VERIFIER_STORE").unwrap_or_else(|_| String::from("/var/data/verifier"));
-        Ok(Self { pool, store_path })
+    pub async fn from_settings(settings: &Settings) -> Result<Self> {
+        let pool = settings.database.connect(LOADER_DB_POOL_SIZE).await?;
```
jeffgrunewald, line 48: should these be configurable settings as well?

poc_iot_verifier/src/runner.rs
```diff
@@ -224,8 +227,16 @@ impl Runner {
             // top level checks complete, verify the POC reports
             //
 
+            let follower =
+                follower_service::FollowerService::from_settings(&self.settings.follower)?;
```
jeffgrunewald, line 231: this is going to establish a new grpc connection to the follower node for every single PoC we process; wouldn't it be better to establish a _single_ connection to the follower here in the running and then clone the client and pass it into the `Poc::new/4` function?

mobile_index/src/indexer.rs
```diff
@@ -0,0 +1,102 @@
...
+        let last_known_timestamp = MetaValue::<DateTime<Utc>>::fetch_or_insert_with(
+            &self.pool,
+            "last_verifier_timestamp",
+            || Utc.timestamp(0, 0),
+        )
+        .await?
+        .value()
+        .to_owned();
+
+        let time_range = last_known_timestamp..Utc::now();
```
jeffgrunewald, line 60: need to check-point somehow that this `now` was the cutoff for the collection of files processed from the bucket

node_follower/src/settings.rs
bbalser: This is a good question for @madninja, I am not sure his reasoning for creating 2 connections to the follower.

ingest/src/settings.rs
```diff
@@ -0,0 +1,78 @@
...
+
+#[derive(Debug, Deserialize)]
+pub struct Settings {
+    /// RUST_LOG compatible settings string. Default
+    /// "poc_ingest=debug,poc_store=info"
+    #[serde(default = "default_log")]
+    pub log: String,
+    /// Mode to run the server in (lora or mobile). Reuquired
+    pub mode: Mode,
+    /// Listen address. Required. Default is 0.0.0.0:9081
```
jontow, line 14: listing defaults inline here means 3 places they must be updated, should we ever need to, maybe they're not needed here?

mobile_rewards/src/settings.rs
```diff
@@ -0,0 +1,68 @@
...
+
+        if let Some(file) = path {
+            // Add optional settings file
+            builder = builder
+                .add_source(File::with_name(&file.as_ref().to_string_lossy()).required(false));
+        }
+        // Add in settings from the environment (with a prefix of MI)
+        // Eg.. `MI_DEBUG=1 ./target/app` would set the `debug` key
+        builder
+            .add_source(Environment::with_prefix("MI").separator("_"))
```
jontow, line 58: why `MI` here?  should be `MR` or similar to avoid conflict with mobile_index overrides?

db_store/src/settings.rs
```diff
@@ -0,0 +1,28 @@
+use crate::Result;
+use http::Uri;
+use serde::Deserialize;
+use sqlx::{postgres::PgPoolOptions, Pool, Postgres};
+
+#[derive(Debug, Deserialize, Clone)]
+pub struct Settings {
+    /// Max open connections to the database. If absent a default is calculated
+    /// by application code
```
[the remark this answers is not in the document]
bbalser: it is done this way because there can be multiple processes in an application that each connect to the db and want to have a different number of max connections.  The example is poc_iot_verifier.

db_store/src/settings.rs
```diff
@@ -0,0 +1,28 @@
...
+    pub max_connections: Option<u32>,
+    /// URL to access the postgres database. For example:
+    /// postgres://postgres:postgres@127.0.0.1:5432/mobile_index_db
+    #[serde(with = "http_serde::uri")]
+    pub url: Uri,
+}
+
+impl Settings {
+    pub async fn connect(&self, default_max_connections: usize) -> Result<Pool<Postgres>> {
```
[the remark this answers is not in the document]
bbalser: It is a valid point, but I like the simplicity of just having the settings module know how to use the configuration it owns to establish the connection.  I vote we leave it as is.

db_store/src/settings.rs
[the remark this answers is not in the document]
jeffgrunewald: But this is just for getting meta values out of the db, not generic Postgres access. You’re saying there’s applications that have multiple processes each with different requirements for number of max connections they want to allow for setting and getting state k/v pairs like “last_something_something_processed”?

file_store/src/cli/info.rs
```diff
@@ -27,7 +27,7 @@ pub struct Cmd {
 }
 
 impl Cmd {
-    pub async fn run(&self) -> Result {
+    pub async fn run(&self, _settings: &Settings) -> Result {
```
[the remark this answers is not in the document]
bbalser: It not part of a contract, my guess is that it is an attempt to standardize on clii command args. Most clis will want some settings so we will just always hand them over.  We can make a change if you think it adds confusion.

mobile_rewards/src/error.rs
```diff
@@ -30,8 +30,8 @@ pub enum Error {
     NotFound(String),
     #[error("transaction error")]
     TransactionError(String),
-    #[error("meta error")]
-    MetaError(#[from] db_store::MetaError),
+    #[error("db error")]
+    DbError(#[from] db_store::Error),
```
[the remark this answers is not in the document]
bbalser: This is a still a weakness for me, but won't the DbError contain specifics around the error and why it occurred?

poc_iot_verifier/src/main.rs
bbalser: @madninja You got an reasoning for this, or just an oversight?

poc_iot_verifier/src/runner.rs
bbalser: This relates back to the answer earlier, but there are multiple processes that each determine this value.  We could make a different database section in the config for each process if we feel like that adds clarity, but feels like overkill to me.

poc_iot_verifier/src/runner.rs
bbalser: @andymck ??

mobile_rewards/src/error.rs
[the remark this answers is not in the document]
jeffgrunewald: I don’t know what specifics per se but the two errors the db_store is exporting are just decode errors and sqlx errors, neither of which are specific to the db_store; if the error will include the context that those particular root problems came up by way of db_store and not any other decode or sqlx error source in the app it’s probably okay so long as we remember “db_store means meta values” when debugging in anger

mobile_index/src/decimal_scalar.rs
```diff
@@ -0,0 +1,112 @@
...
+            }
+        }
+
+        impl From<i32> for $stype {
+            fn from(v: i32) -> Self {
+                if let Some(mut data) = Decimal::from_i32(v) {
+                    data.set_scale($scale).unwrap();
+                    return Self(data);
+                }
+                panic!("u64 could not be converted into Decimal")
```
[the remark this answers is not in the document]
bbalser: Fixed 5655a52

mobile_rewards/src/settings.rs
bbalser: Good catch, Fixed 92080ef

db_store/src/settings.rs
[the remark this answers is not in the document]
madninja: I know the current code is. But the crate is called db_store which implies more general db functions

db_store/src/settings.rs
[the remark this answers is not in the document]
madninja: Yeah. I tried both ways. This reduced repetition

mobile_rewards/src/error.rs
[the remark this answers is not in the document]
madninja: It’ll clear itself up once we get the top level apps to remove their errors like mattie is rolling in

db_store/src/settings.rs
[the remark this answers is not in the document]
jeffgrunewald: That makes sense

node_follower/src/settings.rs
madninja: Different clients are returned

db_store/src/settings.rs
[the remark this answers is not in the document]
bbalser: Jeff and I talked this over and came to an understanding.

db_store/src/settings.rs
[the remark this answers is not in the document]
xandkar: Agree with Jeff here - this is a super-odd place to have a `connect` method. I don't expect my settings to connect to anything.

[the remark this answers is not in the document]
xandkar: Like, settings is generally expected to be nothing more than a database of values - it shouldn't know how to use itself.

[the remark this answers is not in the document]
xandkar: I guess there's some precedence in `cli::run()` pattern, but that one is somehow less jarring, since you expect a dispatch of CLI commands.

bbalser (approved):

ingest/src/settings.rs
```diff
@@ -0,0 +1,78 @@
...
+use serde::Deserialize;
+use std::{net::SocketAddr, path::Path, str::FromStr};
+
+#[derive(Debug, Deserialize)]
+pub struct Settings {
+    /// RUST_LOG compatible settings string. Default
+    /// "poc_ingest=debug,poc_store=info"
+    #[serde(default = "default_log")]
+    pub log: String,
+    /// Mode to run the server in (lora or mobile). Reuquired
```
xandkar, line 12:
```suggestion
    /// Mode to run the server in (lora or mobile). Required
```

ingest/src/settings.rs
xandkar: Good point. Default-getting function should be sufficiently helpful for code reading.

mobile_index/src/decimal_scalar.rs
```diff
@@ -0,0 +1,112 @@
...
+            }
+        }
+
+        impl From<u64> for $stype {
+            fn from(v: u64) -> Self {
+                if let Some(mut data) = Decimal::from_u64(v) {
+                    data.set_scale($scale).unwrap();
+                    return Self(data);
+                }
+                panic!("u64 could not be converted into Decimal")
```
xandkar, line 66: @madninja why not just call `.expect` on these?

xandkar (approved):

jeffgrunewald (approved):

poc_mobile_verifier/src/cli/reward_from_db.rs
```diff
@@ -31,18 +27,8 @@ impl Cmd {
...
-                .connect_lazy(),
-        );
-
-        let db_connection_str = dotenv::var("DATABASE_URL")?;
-        let pool = PgPoolOptions::new()
-            .max_connections(10)
-            .connect(&db_connection_str)
-            .await?;
+        let follower = settings.follower.connect_follower()?;
+        let pool = settings.database.connect(10).await?;
```
dpezely, line 31: Hard-coded value of `10`.  Does this belong in `settings.toml`?

poc_mobile_verifier/src/cli/server.rs
```diff
@@ -28,21 +19,14 @@ impl Cmd {
...
         });
 
-        let db_connection_str = dotenv::var("DATABASE_URL")?;
-
-        let pool = PgPoolOptions::new()
-            .max_connections(10)
-            .connect(&db_connection_str)
-            .await?;
-
+        let pool = settings.database.connect(10).await?;
```
dpezely, line 22: Hard-coded value... Maybe belongs in `settings.toml`?

poc_mobile_verifier/src/main.rs
```diff
@@ -1,40 +1,55 @@
...
-    Result,
+    Result, Settings,
 };
+use std::path;
 use tracing_subscriber::{layer::SubscriberExt, util::SubscriberInitExt};
 
 #[derive(clap::Parser)]
 #[clap(version = env!("CARGO_PKG_VERSION"))]
-#[clap(about = "Verify Shares")]
+#[clap(about = "Helium Mobile Share Server")]
```
dpezely, line 11: This value should appear in Cargo.toml for one place to maintain.
```suggestion
#[clap(about = env!("CARGO_PKG_DESCRIPTION"))]
```
See https://doc.rust-lang.org/cargo/reference/environment-variables.html

poc_iot_verifier/src/main.rs
```diff
@@ -1,38 +1,85 @@
-use poc_iot_verifier::{loader, mk_db_pool, purger, runner, Result};
+use clap::Parser;
+use poc_iot_verifier::{loader, purger, runner, Result, Settings};
+use std::path;
 use tokio::signal;
 use tracing_subscriber::{layer::SubscriberExt, util::SubscriberInitExt};
 
+#[derive(Debug, clap::Parser)]
+#[clap(version = env!("CARGO_PKG_VERSION"))]
+#[clap(about = "Helium POC IOT Verifier")]
```
dpezely, line 9: This value should appear in Cargo.toml for one place to maintain.
```suggestion
#[clap(about = env!("CARGO_PKG_DESCRIPTION"))]
```
See https://doc.rust-lang.org/cargo/reference/environment-variables.html

poc_iot_injector/src/main.rs
```diff
@@ -11,29 +12,42 @@ pub enum Cmd {
...
+        match self {
+            Self::Generate(cmd) => cmd.run(&settings).await,
+            Self::Server(cmd) => cmd.run(&settings).await,
+        }
+    }
+}
+
 #[derive(Debug, clap::Parser)]
 #[clap(version = env!("CARGO_PKG_VERSION"))]
 #[clap(about = "Helium POC IOT Injector Server")]
```
dpezely, line 26: This value should appear in Cargo.toml for one place to maintain.
```suggestion
#[clap(about = env!("CARGO_PKG_DESCRIPTION"))]
```
See https://doc.rust-lang.org/cargo/reference/environment-variables.html

mobile_rewards/src/main.rs
```diff
@@ -1,43 +1,77 @@
...
-    let rs_keypair = load_from_file(&keypair_file)?;
-
-    // Reward server
-    let mut reward_server = Server::new(pool.clone(), rs_keypair).await?;
-
-    reward_server.run(shutdown_listener.clone()).await?;
-    Ok(())
+#[derive(Debug, clap::Parser)]
+#[clap(version = env!("CARGO_PKG_VERSION"))]
+#[clap(about = "Helium Mobile Rewards Server")]
```
dpezely, line 9: This value should appear in Cargo.toml for one place to maintain.
```suggestion
#[clap(about = env!("CARGO_PKG_DESCRIPTION"))]
```
See https://doc.rust-lang.org/cargo/reference/environment-variables.html

mobile_index/src/main.rs
```diff
@@ -0,0 +1,76 @@
+use clap::Parser;
+use mobile_index::{settings::Settings, Indexer, Result};
+use std::path::PathBuf;
+use tokio::signal;
+use tracing_subscriber::{layer::SubscriberExt, util::SubscriberInitExt};
+
+#[derive(Debug, clap::Parser)]
+#[clap(version = env!("CARGO_PKG_VERSION"))]
+#[clap(about = "Helium Mobile Indexer")]
```
dpezely, line 9: This value should appear in Cargo.toml for one place to maintain.
```suggestion
#[clap(about = env!("CARGO_PKG_DESCRIPTION"))]
```
See https://doc.rust-lang.org/cargo/reference/environment-variables.html

dpezely (approved):
All points can be addressed in a future PR or ignored.

poc_iot_verifier/src/runner.rs
andymck: yeah that might be an issue.  I am not entirely certain of how the follower works here and the mechanics of opening and closing the connection but in theory it should be reusing the same grpc connection as the destination is the same.   In any case I will move it out into the runner

PR #146, opened 2022-10-19
Chunk up verification periods

maplant:


bbalser (approved):

poc_mobile_verifier/src/verifier.rs
```diff
@@ -46,8 +46,11 @@ impl VerifierDaemon {
                 || epoch_since_last_reward_duration >= reward_period
             {
                 let epoch_duration = epoch_since_last_verify_duration.min(verification_period);
+                let last_rewarded_end_time = Utc.timestamp(*self.last_verified_end_time.value(), 0);
```
madninja, line 49: Why is last_rewarded_end_time the same as last_verified_end_time on the line below?

poc_mobile_verifier/src/verifier.rs
```diff
@@ -46,8 +46,11 @@ impl VerifierDaemon {
...
                 let last_verified_end_time = Utc.timestamp(*self.last_verified_end_time.value(), 0);
-                let epoch = last_verified_end_time..(last_verified_end_time + epoch_duration);
+                let epoch = last_verified_end_time
+                    ..(last_verified_end_time + epoch_duration)
+                        .min(last_rewarded_end_time + Duration::hours(24));
```
madninja, line 53: This would limit the epoch to 24 hours right? Which would exclude actual being able to do a catch up across more than that if we're done for a length of time? Why is it being constrained?

madninja: Or is the `Duration::zero()` below telling me that you'd issue multiple 24 hour intervals instead of one "big one"

poc_mobile_verifier/src/verifier.rs
maplant: Fixed. Should be rewarded time

poc_mobile_verifier/src/verifier.rs
maplant: We'd issue multiple 24 hour periods (or any period actually, restricting to 24 was a mistake). The reason for this is that if we make the epoch based on `Utc::now()`, we continually drift longer than 24 hours.

poc_mobile_verifier/src/verifier.rs
```diff
@@ -174,3 +186,10 @@ impl Verifier {
 fn epoch_duration(epoch: &Range<DateTime<Utc>>) -> Duration {
     epoch.end - epoch.start
 }
+
+fn fmt_duration(duration: &Duration) -> String {
```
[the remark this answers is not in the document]
maplant: Good catch

poc_mobile_verifier/src/verifier.rs
maplant: It is now. (There were two)

jeffgrunewald (approved):

bbalser (approved):

PR #147, opened 2022-10-20
tidy up

andymck:
Eliminate unwraps, improve some comments and move some code around a lil 
Removes server module, which is not in use and accidentally checked in at some point


bbalser (approved):

bbalser (approved):

xandkar:
@madninja shall we consider adding `-W clippy::unwrap_used` to clippy's args in CI. It won't pass right now, but can be a goal.

madninja:
> @madninja shall we consider adding `-W clippy::unwrap_used` to clippy's args in CI. It won't pass right now, but can be a goal.

oh so `-D clippy::all` doesn't make unwrap_used a warning? annoying.. My only concern is that there are a couple of places where using unwrap is actually ok

xandkar:
> oh so `-D clippy::all` doesn't make unwrap_used a warning? annoying.

Yeah, it's a terribly-named group, since it effectively means "default", rather than actually "all". -D clippy::all` seems to just turn default warns into denies.

`unwrap` looks to be grouped in `clippy::restriction` group.

`clippy::pedantic` is even more fun.

PR #148, opened 2022-10-24
realign follower proto response with new structure

jeffgrunewald:
The addition of a "GatewayNotFound" response to the follower node GRPC API requires a restructuring of the response struct received by clients of the service querying the `find_gateway` RPC; this change adapts to the new proto structure to handle the possibility the gateway is not found and converts it to an appropriate error in that situation.

### DIFF
--- a/node_follower/src/follower_service.rs
+++ b/node_follower/src/follower_service.rs
@@ -1,19 +1,15 @@
 use crate::{
-    gateway_resp::{FollowerGatewayResp, GatewayInfoResolver},
-    Result,
+    gateway_resp::{GatewayInfo, GatewayInfoResolver},
+    Error, Result, CONNECT_TIMEOUT, DEFAULT_URI, RPC_TIMEOUT,
 };
 use helium_crypto::PublicKey;
 use helium_proto::services::{
-    follower::{self, FollowerGatewayReqV1},
+    follower::{self, follower_gateway_resp_v1::Result as GatewayResult, FollowerGatewayReqV1},
     Channel, Endpoint,
 };
 use http::Uri;
-use std::{env, str::FromStr, time::Duration};
+use std::{env, str::FromStr};
 use tonic::Streaming;
-
-const CONNECT_TIMEOUT: Duration = Duration::from_secs(5);
-const RPC_TIMEOUT: Duration = Duration::from_secs(5);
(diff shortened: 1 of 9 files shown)


bbalser (approved):

node_follower/src/follower_service.rs
```diff
@@ -24,12 +20,15 @@ pub struct FollowerService {
...
-    async fn resolve_gateway_info(&mut self, address: &PublicKey) -> Result<FollowerGatewayResp> {
+    async fn resolve_gateway_info(&mut self, address: &PublicKey) -> Result<GatewayInfo> {
         let req = FollowerGatewayReqV1 {
             address: address.to_vec(),
         };
         let res = self.client.find_gateway(req).await?.into_inner();
-        Ok(res.try_into()?)
+        match res.result {
+            Some(GatewayResult::Info(gateway_info)) => Ok(gateway_info.try_into()?),
+            _ => Err(Error::GatewayNotFound(format!("{address}"))),
```
andymck, line 30: Should you not explicitly match on `follower_error` here, rather than default all errors to gateway not found ?

poc_iot_verifier/src/runner.rs
```diff
@@ -233,8 +233,7 @@ impl Runner {
             match beacon_verify_result.result {
                 VerificationStatus::Valid => {
                     // beacon is valid, verify the witnesses
-                    let beacon_info: FollowerGatewayResp =
-                        beacon_verify_result.gateway_info.unwrap();
+                    let beacon_info: GatewayInfo = beacon_verify_result.gateway_info.unwrap();
```
andymck, line 236: ahhhhhh unwrap ? No worries, i have refactored this out on the tidy up PR but we should get this PR in first and then i will resolve the conflicts on the tidy up pr.

andymck (approved):
lgtm.  We should get this in ASAP as its blocking any subsequent PR which updates to latest protos ( such as one I am working on to handle the new `tmst` field for beacon/witness reports

node_follower/src/follower_service.rs
jeffgrunewald: Expediency at the moment; since there’s only the one `follower_error` alternative result at the moment.

PR #150, opened 2022-10-25
Fix bug where we will not perform enough verifications if the reward period has exceeded and there are multiple verification chunks

maplant:


bbalser:
I believe https://github.com/helium/oracles/pull/155 will fix this bug

PR #140, opened 2022-10-19
Issue #34:  Output speedtest averages to S3

maplant:
Tasks left to complete: 

- [x] Fix generate cli command
- [x] Port tests from reward_speed_share
- [x] Write report of averages out every verification period
- [x] Change binary check (valid/invalid) to coefficient multiplication

PR #151, opened 2022-10-25
support tmst field in beacon & witness reports

andymck:


andymck:
structs were already updated in #140

PR #152, opened 2022-10-26
Write poc receipt txns to s3 bucket

vihu:
# Summary
This adds support for writing poc receipt txns to s3 bucket once they have been submitted to the chain.

# Depends on
~~We should land https://github.com/helium/oracles/pull/135 prior to this one~~

# Caveats
- The if-else clauses could use some scrutiny
- Am I actually writing the txn correctly to S3? I mostly looked through poc_mobile_verifier for depositing the file
- The env var naming could also use some careful eyes


vihu:
Superseded by #167

PR #154, opened 2022-10-26
misc fixes

andymck:
- fix ambiguous columns issue on report query
- Adds on conflict clause to ignore duplicate poc report and entropy report inserts
- Adds max age for consuming events from s3 ( beacon/witness report or entropy report ).  Ignore reports outside of this age
- Adds check to incoming beacon and witness reports to identify if unknown gateway.  If true the drop the report
- Code tidy up

### DIFF
--- a/poc_iot_verifier/src/loader.rs
+++ b/poc_iot_verifier/src/loader.rs
@@ -202,6 +202,7 @@ impl Loader {
                     }
                     false => {
                         // if a gateway doesnt exist then drop the report
+                        // dont even insert into DB
                         tracing::warn!(
                             "dropping witness report as gateway not found: {:?}",
                             &witness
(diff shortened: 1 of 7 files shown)


poc_iot_verifier/src/loader.rs
```diff
@@ -157,34 +162,57 @@ impl Loader {
...
-                    &self.pool,
-                    beacon.ingest_id(),
-                    beacon.report.remote_entropy,
-                    packet_data,
-                    buf.to_vec(),
-                    &beacon.received_timestamp,
-                    ReportType::Beacon,
-                )
-                .await
+                match self.check_valid_gateway(&beacon.report.pub_key).await {
```
vihu, line 165: nit: Arguably this logic can be hoisted in a function like `insert_report_for_existing_gateway`? Just so we don't do too much nesting.

poc_iot_verifier/src/loader.rs
```diff
@@ -157,34 +162,57 @@ impl Loader {
...
-                    &self.pool,
-                    witness.ingest_id(),
-                    Vec::<u8>::with_capacity(0),
-                    packet_data,
-                    buf.to_vec(),
-                    &witness.received_timestamp,
-                    ReportType::Witness,
-                )
-                .await
+                match self.check_valid_gateway(&witness.report.pub_key).await {
```
vihu, line 193: Similar thing here, could potentially make a function to add both types of reports?

poc_iot_verifier/src/loader.rs
andymck: yeah maybe but I am not sure I like putting too much logic into the `DB` modules, ie in this case poc_report.rs

bbalser (approved):

vihu (approved):

PR #156, opened 2022-10-26
Simplify db access in poc_mobile_verifier

bbalser:
So I wanted to simplify the db access and move the responsibility into a single entity.

Current downside, if the 3 values `last_verified_end_time`, `last_rewarded_end_time`, and `next_rewarded_end_time` are not in the meta table it will crash the server.

### DIFF
--- a/db_store/src/lib.rs
+++ b/db_store/src/lib.rs
@@ -1,5 +1,7 @@
 use std::str::FromStr;
 
+pub mod meta;
+
 /// A key-value pair that is stored in the metadata table.
 pub struct MetaValue<T> {
     key: String,
@@ -12,6 +14,8 @@ pub enum MetaError {
     SqlError(#[from] sqlx::Error),
     #[error("Failed to decode meta value")]
     DecodeError,
+    #[error("not found")]
+    NotFound(String),
 }
 
 impl<T> MetaValue<T> {
--- a/db_store/src/lib.rs
+++ b/db_store/src/lib.rs
@@ -43,7 +43,7 @@ where
     where
         E: sqlx::Executor<'c, Database = sqlx::Postgres>,
     {
-        meta::save(exec, &self.key, &self.value.to_string()).await
+        meta::store(exec, &self.key, &self.value.to_string()).await
     }
 }
 
@@ -59,7 +59,7 @@ where
     where
         E: sqlx::Executor<'c, Database = sqlx::Postgres> + Copy,
     {
-        let result: Result<String, MetaError> = meta::get::<String>(exec, key).await;
+        let result: Result<String, MetaError> = meta::fetch::<String>(exec, key).await;
 
         match result {
             Ok(str_val) => {
@@ -83,7 +83,7 @@ where
     where
         E: sqlx::PgExecutor<'c>,
     {
-        meta::save(exec, &self.key, new_val.to_string()).await?;
+        meta::store(exec, &self.key, new_val.to_string()).await?;
         Ok(std::mem::replace(&mut self.value, new_val))
     }
 }
--- a/db_store/src/lib.rs
+++ b/db_store/src/lib.rs
@@ -58,17 +58,7 @@ where
     where
         E: sqlx::Executor<'c, Database = sqlx::Postgres>,
     {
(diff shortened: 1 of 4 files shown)


db_store/src/lib.rs
```diff
@@ -12,6 +14,8 @@ pub enum MetaError {
     SqlError(#[from] sqlx::Error),
     #[error("Failed to decode meta value")]
     DecodeError,
+    #[error("not found")]
```
jeffgrunewald, line 17:
```suggestion
    #[error("not found {0}")]
```
to report the missing string

db_store/src/meta.rs
```diff
@@ -0,0 +1,38 @@
+use std::str::FromStr;
```
jeffgrunewald, line 1: this all makes sense but it seems like it could get confusing that the `MetaValue` struct is defined in the `lib.rs` module and this more generic accessor function module is named `meta.rs`? i know these are dealing with just setting and getting key/value pairs but maybe these two modules can be unified in some way?

poc_mobile_verifier/src/verifier.rs
```diff
@@ -150,6 +127,45 @@ impl VerifierDaemon {
 
         Ok(())
     }
+
+    async fn last_verified_end_time(&self) -> Result<DateTime<Utc>> {
+        Ok(Utc.timestamp(meta::get(&self.pool, "last_verified_end_time").await?, 0))
+    }
+
+    async fn save_last_verified_end_time(
+        exec: impl PgExecutor<'_>,
```
jeffgrunewald, line 136: so the transaction also implements the PgExecutor? it makes sense to use the transaction in the context you've got it but independent of that this function signature seems strange that the getter uses `&self.pool` but the setter expects you to pass it a PgExecutor; is there a more specific type/trait you can gate this on or at least comment why this is set up the way it is to avoid future confusion?

db_store/src/lib.rs
bbalser: Hopefully fixed correctly in 4343973

db_store/src/meta.rs
bbalser: I agree it is confusing and I can drop the simpler `meta` if we want, but I did change MetaValue to use `meta` under the covers.  48754ee

jeffgrunewald (approved):

poc_mobile_verifier/src/verifier.rs
bbalser: agreed, I didn't like it when I wrote it.  I think the function should just take the PgExecutor because it doesn't need to know whether it is in a transaction or not.  I could change the reader functions to also accept exec parameter to make them all the same, or I could create an `Option` in the struct for the transaction and use that if available or pool if not, but I like that the least.

db_store/src/meta.rs
jeffgrunewald: i think having the metavalue's use the meta module functions is good

poc_mobile_verifier/src/verifier.rs
jeffgrunewald: it's not terrible as-is; if you do anything i'd just make the getter's take an executor as well but it's probably fine already

poc_mobile_verifier/src/verifier.rs
```diff
@@ -150,6 +127,45 @@ impl VerifierDaemon {
...
+        meta::save(exec, "last_verified_end_time", value.timestamp() as i64)
+            .await
+            .map_err(Error::from)
+    }
+
+    async fn last_rewarded_end_time(&self) -> Result<DateTime<Utc>> {
+        Ok(Utc.timestamp(meta::get(&self.pool, "last_rewarded_end_time").await?, 0))
+    }
+
+    async fn save_last_rewarded_end_time(
```
maplant, line 148: You shouldn't scope functions to a struct that don't either take `Self` as an argument or return a type that includes `Self` - these should all be free functions

bbalser: Good cal, Fixed ebaa599

xandkar (approved):

PR #157, opened 2022-10-26
Clean up speedtest tier comparisons

maplant:
This is a fun thing you can do with rust enums!

### DIFF
--- a/poc_mobile_verifier/src/speedtests.rs
+++ b/poc_mobile_verifier/src/speedtests.rs
@@ -348,10 +348,10 @@ const fn mbps(mbps: u64) -> u64 {
 
 #[derive(Copy, Clone, Debug, PartialEq, Eq, PartialOrd, Ord)]
 pub enum SpeedtestTier {
-    Failed,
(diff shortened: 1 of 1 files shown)


jeffgrunewald (commented):
this looks like it relies on the ordering of the enum members comparing where the earlier enum variants are < later variants? is this a stable / obvious behavior we can rely on and clippy or just someone really nitpicky won't re-order the enum variants in the future?

maplant:
> this looks like it relies on the ordering of the enum members comparing where the earlier enum variants are < later variants? is this a stable / obvious behavior we can rely on and clippy or just someone really nitpicky won't re-order the enum variants in the future?

I can also add numerical values to each

jeffgrunewald:
> I can also add numerical values to each

I think that would make it much safer. I said "someone really nitpicky" but what i mean is "it'd prevent _me_ from coming back and alphabetizing them" in the future

jeffgrunewald (approved):

andymck (approved):

PR #158, opened 2022-10-27
factor speedtest averages into mobile rewards

jeffgrunewald:
incorporate the rolling speedtest averages, when available, into the mobile rewards calculation for hotspots. when a speedtest rolling average cannot be calculated (the minimum number of speedtests have not been submitted within the required period) then the multiplier for speedtests defaults to 0.0 (decimal) and the hotspot does not receive any rewards for the reward cycle.

### DIFF
--- a/poc_mobile_verifier/src/speedtests.rs
+++ b/poc_mobile_verifier/src/speedtests.rs
@@ -9,6 +9,8 @@ use file_store::{
 use futures::stream::{self, StreamExt, TryStreamExt};
 use helium_crypto::PublicKey;
 use helium_proto::services::poc_mobile as proto;
+use rust_decimal::Decimal;
+use rust_decimal_macros::dec;
 use sqlx::{
     postgres::{types::PgHasArrayType, PgTypeInfo},
     FromRow, Type,
@@ -17,6 +19,8 @@ use std::{
     collections::{HashMap, VecDeque},
     ops::Range,
 };
+
+const SPEEDTEST_AVG_MAX_DATA_POINTS: usize = 6;
 
 #[derive(Debug, Clone, Type)]
 #[sqlx(type_name = "speedtest")]
@@ -93,7 +97,9 @@ impl SpeedtestRollingAverage {
         // Write out the speedtests to S3
         let average = Average::from(&self.speedtests);
         let validity = average.validity() as i32;
-        let reward_multiplier = average.reward_multiplier();
+        // this is guaratneed to safely convert and not panic as it can only be one of
+        // four possible decimal values based on the speedtest average tier
+        let reward_multiplier = average.reward_multiplier().try_into().unwrap();
         let Average {
             upload_speed_avg_bps,
             download_speed_avg_bps,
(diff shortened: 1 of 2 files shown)


poc_mobile_verifier/src/speedtests.rs
```diff
@@ -93,7 +97,9 @@ impl SpeedtestRollingAverage {
         // Write out the speedtests to S3
         let average = Average::from(&self.speedtests);
         let validity = average.validity() as i32;
-        let reward_multiplier = average.reward_multiplier();
+        // this is guaratneed to safely convert and not panic as it can only be one of
+        // four possible decimal values based on the speedtest average tier
+        let reward_multiplier = average.reward_multiplier().try_into().unwrap();
```
andymck, line 102: if guaranteed to not panic, why then use try_info and not a plain old From/Into impl ?

poc_mobile_verifier/src/subnetwork_rewards.rs
```diff
@@ -106,6 +113,52 @@ mod test {
         }
     }
 
+    fn bytes_per_s(mbps: i64) -> i64 {
+        mbps * 125000
+    }
+
+    fn cell_type_weight(cbsd_id: &String) -> Decimal {
+        CellType::from_cbsd_id(cbsd_id)
+            .expect("unable to get cell_type")
```
andymck, line 122: do you have to use expect here, the consensus has been to avoid (outside of tests ) as it will result in a panic

poc_mobile_verifier/src/speedtests.rs
jeffgrunewald: Decimal only implements TryFrom/Into for floats

poc_mobile_verifier/src/subnetwork_rewards.rs
jeffgrunewald: This is a function only in the test sub-module so it should have the same effect

poc_mobile_verifier/src/speedtests.rs
```diff
@@ -229,8 +235,8 @@ impl SpeedtestAverages {
             // Unwrap here is guaranteed never to panic
             let window = speedtests.get_mut(&pubkey).unwrap();
 
-            // If there are six speedtests in the window, remove the one in the front
-            while window.len() >= 6 {
+            // If there are N speedtests in the window, remove the one in the front
+            while window.len() >= SPEEDTEST_AVG_MAX_DATA_POINTS {
```
bbalser, line 239: nice!

poc_mobile_verifier/src/speedtests.rs
```diff
@@ -295,12 +301,7 @@ where
                 latency_avg_ms: sum_latency / window_size as u32,
             }
         } else {
-            Average {
-                window_size,
-                upload_speed_avg_bps: 0,
-                download_speed_avg_bps: 0,
-                latency_avg_ms: 0,
-            }
+            Default::default()
```
bbalser, line 304: NIT: could we make this `Average::default()` so it is clearer?

jeffgrunewald: Absolutely; agreed

jeffgrunewald: fixed [50a7d3e](https://github.com/helium/oracles/pull/158/commits/50a7d3efba826fdadbce3633e46bdf917c59b876)

maplant (approved):

PR #159, opened 2022-10-27
Add reward_from_db cli command

maplant:


bbalser (approved):

poc_mobile_verifier/src/cli/reward_from_db.rs
```diff
@@ -0,0 +1,70 @@
...
+    start: NaiveDateTime,
+    #[clap(long)]
+    end: NaiveDateTime,
+}
+
+impl Cmd {
+    pub async fn run(self) -> Result {
+        let Self { start, end } = self;
+
+        let start = DateTime::from_utc(start, Utc);
```
jeffgrunewald, line 26: any reason to pattern match the start and end out above vs. `self.start` and `self.end` here or just preference?

poc_mobile_verifier/src/cli/reward_from_db.rs
```diff
@@ -0,0 +1,70 @@
...
+            .rewards
+            .iter()
+            .map(|r| (PublicKey::try_from(r.account.as_slice())?, r.amount))
+            .collect();
+
+        println!(
+            "{}",
+            serde_json::to_string_pretty(&json!({
+                "rewards": rewards,
+                "total_rewards": total_rewards,
```
jeffgrunewald, line 64: can we have a CYA here like `expected_total` or similar that computes the number of mobile bones we expect to emit based on the duration of the epoch passed in on the command? then you can very quickly see on the command output that you're awarding the amount you expect

jeffgrunewald (approved):
one nice-to-have; otherwise lgtm

poc_mobile_verifier/src/cli/reward_from_db.rs
maplant: just preference

poc_mobile_verifier/src/cli/reward_from_db.rs
maplant: sure

jeffgrunewald (approved):
🚀

PR #160, opened 2022-10-27
The great error type cleanup part one: poc_mobile_verifier

maplant:
Arguably, verifier could have its own error type but I think anyhow is fine


poc_mobile_verifier/src/mobile.rs
```diff
@@ -1,22 +1,37 @@
-use crate::{error::DecodeError, Error, Result};
 use core::fmt;
 use rust_decimal::prelude::*;
 use serde::{de::Deserializer, Deserialize, Serialize};
 use std::str::FromStr;
 
+#[derive(thiserror::Error, Debug)]
+pub enum DecodeError {
```
madninja, line 7: Reduce this to just what's needed for MobileError vs all possible DecodeErrors?

poc_mobile_verifier/src/scheduler.rs
```diff
@@ -10,6 +9,12 @@ pub struct Scheduler {
     pub reward_period: Range<DateTime<Utc>>,
 }
 
+// I don't want to have to define this struct, but otherwise I'd have to
+// import the deprecated time crate.
+#[derive(thiserror::Error, Debug)]
```
madninja, line 14: You could import Tokio's time::Duration right? I know it's just a re-export.

Alternately just use std::time::Duration already.. We'll have a lot to update regardless when that deprecation finally hits?

poc_mobile_verifier/src/speedtests.rs
```diff
@@ -408,20 +408,19 @@ impl SpeedtestTier {
 
 #[async_trait::async_trait]
 pub trait SpeedtestStore {
-    async fn fetch(self, id: &PublicKey) -> Result<SpeedtestRollingAverage>;
+    async fn fetch(self, id: &PublicKey) -> Result<SpeedtestRollingAverage, sqlx::Error>;
```
madninja, line 411: This is the part I sorta dislike? It's exposing that it's using sqlx under the hood right?

poc_mobile_verifier/src/subnetwork_rewards.rs
```diff
@@ -29,7 +28,7 @@ impl SubnetworkRewards {
         heartbeats: Heartbeats,
         // TODO: Use speedtests as part of the reward calculation
         _speedtests: SpeedtestAverages,
-    ) -> Result<Self> {
+    ) -> Result<Self, tonic::Status> {
```
madninja, line 31: I'd prefer a ResolverError since vs exposing that it's using tonic indirectly?

madninja (commented):
I like it! Just a couple of comments about exposing what looks like mini library details that maybe shouldn't be and overspecifying the Mobile decode error

poc_mobile_verifier/src/heartbeats.rs
```diff
@@ -1,6 +1,6 @@
 //! Heartbeat storage
 
-use crate::{cell_type::CellType, Error, Result};
+use crate::cell_type::CellType;
```
andymck, line 3: curious to understand why 'anyhow' is in use on some modules and not others ? What is the decision making around whether to use it or default to standard rust Result type

andymck (commented):
this generally lgtm,  it will be good to be able to return Result<T, E>.  I always found the way we were handling errors confusing.  A question as to why we only use anyhow on some modules and not others, what is the rationale around when to use it and when not

poc_mobile_verifier/src/heartbeats.rs
maplant: The plan is to convert every crate over to use anyhow, but I figured one PR per module would be easier for people to digest

madninja: i.e. every _top level application_ will use the anyhow crate to make errors have more context. Any library behavior will use a specific Error subtype

poc_mobile_verifier/src/scheduler.rs
maplant: We have to go from chrono -> std -> tokio time

poc_mobile_verifier/src/speedtests.rs
```diff
@@ -403,22 +401,27 @@ impl SpeedtestTier {
...
 // nothing.
 
+#[derive(thiserror::Error, Debug)]
+#[error(transparent)]
+pub struct FetchError(#[from] sqlx::Error);
+
 #[async_trait::async_trait]
 pub trait SpeedtestStore {
-    async fn fetch(self, id: &PublicKey) -> Result<Option<SpeedtestRollingAverage>>;
+    async fn fetch(self, id: &PublicKey) -> Result<Option<SpeedtestRollingAverage>, FetchError>;
```
jeffgrunewald, line 410: why the special error struct here but we just used a `sqlx::Error` in the `heartbeats.rs` module?

maplant: mistake

poc_mobile_verifier/src/heartbeats.rs
```diff
@@ -195,7 +195,7 @@ impl Heartbeat {
         Ok(())
     }
 
-    pub async fn save(self, exec: impl sqlx::PgExecutor<'_>) -> Result<bool> {
+    pub async fn save(self, exec: impl sqlx::PgExecutor<'_>) -> Result<bool, sqlx::Error> {
```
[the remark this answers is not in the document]
maplant: This should be a new type error created by thiserror. Any module that behaves like an API should return a thiserror type

poc_mobile_verifier/src/ingest.rs
```diff
@@ -12,7 +11,7 @@ use std::ops::Range;
 pub async fn ingest_heartbeats(
     file_store: &FileStore,
     epoch: &Range<DateTime<Utc>>,
-) -> Result<impl Stream<Item = CellHeartbeat>> {
+) -> file_store::Result<impl Stream<Item = CellHeartbeat>> {
```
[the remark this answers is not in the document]
maplant: Yes - frankly I think these functions should eventually be moved into the file store crate

dpezely (approved):
Considering the rules of thumb from our earlier chat, this PR looks good to me.

For capturing those rules here:

> You can sum it up in three rules:
> 1. APIs define enum error types 
> 2. Hide implementation details in newtypes 
> 3. Applications use anyhow

poc_mobile_verifier/src/heartbeats.rs
andymck: Jeff is correct, i was referring to why only some 'modules' are using anyhow and others are continuing to use the prior style

andymck (approved):

poc_mobile_verifier/src/heartbeats.rs
```diff
@@ -154,12 +154,16 @@ struct HeartbeatSaveResult {
...
+#[error(transparent)]
+pub struct SaveHeartbeatError(#[from] sqlx::Error);
+
 impl Heartbeat {
     pub async fn validate_heartbeats<'a>(
         heartbeats: impl Stream<Item = CellHeartbeat> + 'a,
         epoch: &'a Range<DateTime<Utc>>,
-    ) -> Result<impl Stream<Item = Self> + 'a> {
-        Ok(heartbeats.map(move |heartbeat_report| {
+    ) -> impl Stream<Item = Self> + 'a {
```
jeffgrunewald, line 165: By removing this result wrapper are we saying that, while processing a stream’s elements may fail in some fashion, actually constructing the stream is a pure functional operation that can’t fail? If so that would make sense to me.

jeffgrunewald (approved):

poc_mobile_verifier/src/heartbeats.rs
maplant: yup!

PR #163, opened 2022-10-28
delete reports after verification, misc tweaks & fixes

andymck:


poc_iot_verifier/src/runner.rs
```diff
@@ -313,19 +309,40 @@ impl Runner {
...
+            lora_invalid_beacon_tx,
+            invalid_poc_proto
+        )
+        .await
+        {
+            Ok(_) => (),
+            Err(err) => {
+                tracing::error!("failed to save invalid_poc to s3, {err}");
+                Report::update_attempts(&self.pool, &beacon_report_id, Utc::now()).await?;
+                return Ok(());
```
vihu, line 338: Do we not want to execute the subsequent steps for deleting poc and saving invalid witness reports? Returning here would bypass all of that I think?

andymck: at this point if the file sink write fails we do want to halt and return as no artifacts have been pushed to the file sink as yet.  As such returning allows the same beacon to be reprocessed next tick and if the issue was a temporary blip, then hopefully it will be successful.  If it keeps on failing then it will exceed the max attempts and end up being purged

andymck:
so having considered the rewards side of verifier, it likely makes more sense to *not* delete the verified reports from the DB and instead stick with what we do atm and update their status.  This was the rewards generation can pull verified reports directly from the DB rather than reading in verified poc from s3

vihu (approved):

PR #165, opened 2022-10-29
push lapsed speedtests filter to write/reward time

jeffgrunewald:
we need to save up to 6 (current maximum data points for calculating speedtest averages) in the database but still ensure that speedtests that have a lapse period greater than the allowable window are _not_ included in average calculations when determining rewards and are _not_ written in reports to the verifier output bucket.

the following extends the maximum window the database looks for when querying saved speedtests for performing average calculations from 12 to 48 hours to ensure we capture a minimum of 2 speedtests if they are present (with a generous buffer to allow for randomness in the speedtest submission logic) as well as to filter out speedtests from the speedtest averages data structure before performing a write of average reports to s3 and calculating reward share weights. when speedtests have a gap of more than 48 hours between submissions starting from the most recent and working backward through the list, all previous speedtests are dropped.

### DIFF
--- a/poc_mobile_verifier/src/speedtests.rs
+++ b/poc_mobile_verifier/src/speedtests.rs
@@ -21,6 +21,7 @@ use std::{
 };
 
 const SPEEDTEST_AVG_MAX_DATA_POINTS: usize = 6;
+const SPEEDTEST_LAPSE: i64 = 48;
 
 #[derive(Debug, Clone, Type)]
 #[sqlx(type_name = "speedtest")]
@@ -163,6 +164,28 @@ impl SpeedtestAverages {
         self.speedtests.get(pub_key).map(Average::from)
     }
 
+    pub fn without_lapsed(self) -> Self {
+        let mut speedtests = HashMap::new();
+
+        for (pubkey, window) in self.speedtests.into_iter() {
+            let mut contiguous_tests = VecDeque::new();
+            let mut window = window.iter().peekable();
+            while let Some(st) = window.next() {
+                match window.peek() {
+                    Some(next_st)
+                        if (st.timestamp - next_st.timestamp)
+                            > Duration::hours(SPEEDTEST_LAPSE) =>
+                    {
+                        break
+                    }
+                    _ => contiguous_tests.push_back(st.clone()),
+                }
+            }
+            speedtests.insert(pubkey.clone(), contiguous_tests);
+        }
+        Self { speedtests }
+    }
+
     pub async fn validated(
         exec: impl sqlx::PgExecutor<'_> + Copy,
         period_end: DateTime<Utc>,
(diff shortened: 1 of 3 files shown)


poc_mobile_verifier/src/speedtests.rs
```diff
@@ -163,6 +164,28 @@ impl SpeedtestAverages {
...
+            let mut window = window.iter().peekable();
+            while let Some(st) = window.next() {
+                match window.peek() {
+                    Some(next_st)
+                        if (st.timestamp - next_st.timestamp)
+                            > Duration::hours(SPEEDTEST_LAPSE) =>
+                    {
+                        break
+                    }
+                    _ => contiguous_tests.push_back(st.clone()),
```
xandkar, line 181: Catchalls shouldn't generally be used for finite-variant types. It costs little extra to _write_ `None`, but is a boon to _read_. In the worst case it can cause a newly added variant to be handled with a handler one didn't really intend. Not really a problem for `Option` per se, but it's a good habit to have.
```suggestion
                    None => contiguous_tests.push_back(st.clone()),
```

bbalser: I don't think it can just be `None` in this case.  Because the previous match contains a guard clause there will be some values of `Some(speedtest)` that fall through to this branch.

jeffgrunewald: in this case i'm trying to catch two separate cases; i need to push the speedtest into the accumulator when the result of `window.peek()` is `None` _and_ when it is `Some(next_st)` with the implicit `if DIFFERENCE_IN_TIMESTAMPS < ARBITRARY_Duration`; explicitly writing `None` in the match loses the majority of the speedtests i care about accumulating and writing a third match head would duplicate the code in the match arm. I also _can't_ unfortunately match on `None | Some(next_st)` b/c the compiler will complain about a variable not being able to bind in all cases

xandkar: > I don't think it can just be `None` in this case. Because the previous match contains a guard clause there will be some values of `Some(speedtest)` that fall through to this branch.

Right! I totally looked passed the guard because it was on the next line _and_ I'm still getting used to guards being `if` and not `when`.

xandkar: > in this case i'm trying to catch two separate cases; i need to push the speedtest into the accumulator when the result of `window.peek()` is `None` _and_ when it is `Some(next_st)` with the implicit `if DIFFERENCE_IN_TIMESTAMPS < ARBITRARY_Duration`; explicitly writing `None` in the match loses the majority of the speedtests i care about accumulating and writing a third match head would duplicate the code in the match arm. I also _can't_ unfortunately match on `None | Some(next_st)` b/c the compiler will complain about a variable not being able to bind in all cases

`None | Some(_)` works for me locally.

xandkar: Case in point - explicit pattern made it very obvious for the reader (me, ahem ...) that there're actually 3 cases here, not 2.

bbalser (approved):

maplant (approved):
I'd be interested to see how your interator adaptor branch cleans this up but the logic looks fine

PR #166, opened 2022-10-29
Add setting to ingest for required network

bbalser:


jeffgrunewald (approved):

ingest/src/settings.rs
```diff
@@ -16,6 +17,8 @@ pub struct Settings {
     pub listen: String,
     /// Local folder for storing intermediate files
     pub cache: String,
+    /// Network required in all public keys:  mainnet | testnet
+    pub network: String,
```
madninja, line 21: You should be able to use the Network type instead of a String

jeffgrunewald (approved):

ingest/src/settings.rs
bbalser: Fixed 075fa20

PR #167, opened 2022-10-30
Add support for storing receipt txns on S3

vihu:
# Summary
This adds support for writing poc receipt txns to s3 bucket once they have been submitted to the chain.

# Caveats
- The if-else clauses could use some scrutiny
- Am I actually writing the txn correctly to S3? I mostly looked through poc_mobile_verifier for depositing the file

# TODO:
- [x] ~~Test with minio. There seems to be a potential bug with file_upload, it just stops after uploading one file to my local minio cluster. Still investigating.~~ We have decided not to do the local upload testing since it's not really meaningful.

### DIFF
--- a/poc_iot_injector/src/cli/generate.rs
+++ b/poc_iot_injector/src/cli/generate.rs
@@ -7,7 +7,7 @@ use futures::{
 };
 use helium_crypto::Keypair;
 use helium_proto::{blockchain_txn::Txn, BlockchainTxn, Message};
-use std::sync::Arc;
+use std::sync::{Arc, Mutex};
 
 /// Generate poc rewards
 #[derive(Debug, clap::Args)]
@@ -36,16 +36,31 @@ impl Cmd {
         let poc_oracle_key = settings.keypair()?;
         let shared_key = Arc::new(poc_oracle_key);
 
+        let success_counter = Arc::new(Mutex::new(0));
+        let failure_counter = Arc::new(Mutex::new(0));
+
         store
             .source_unordered(LOADER_WORKERS, stream::iter(file_list).map(Ok).boxed())
             .try_for_each_concurrent(STORE_WORKERS, |msg| {
                 let shared_key_clone = shared_key.clone();
+                let success_counter_ref = Arc::clone(&success_counter);
+                let failure_counter_ref = Arc::clone(&failure_counter);
                 async move {
-                    let _ = process_msg(msg, shared_key_clone, before_ts).await;
+                    if process_msg(msg, shared_key_clone, before_ts)
+                        .await
+                        .is_some()
+                    {
+                        *success_counter_ref.lock().unwrap() += 1;
+                    } else {
+                        *failure_counter_ref.lock().unwrap() += 1;
+                    }
                     Ok(())
                 }
             })
             .await?;
+
+        tracing::debug!("success_counter: {:?}", success_counter.lock().unwrap());
+        tracing::debug!("failure_counter: {:?}", failure_counter.lock().unwrap());
 
         Ok(())
     }
@@ -55,7 +70,7 @@ async fn process_msg(
     msg: prost::bytes::BytesMut,
     shared_key_clone: Arc<Keypair>,
     before_ts: i64,
-) -> Result<()> {
+) -> Option<BlockchainTxn> {
     if let Ok(Some((txn, _hash, _hash_b64_url))) =
         handle_report_msg(msg.clone(), shared_key_clone, before_ts)
     {
@@ -64,8 +79,9 @@ async fn process_msg(
         };
 
         tracing::debug!("txn_bin: {:?}", tx.encode_to_vec());
+        return Some(tx);
     } else {
         tracing::error!("unable to construct txn for msg {:?}", msg)
     }
-    Ok(())
+    None
 }
--- a/poc_iot_injector/src/cli/generate.rs
+++ b/poc_iot_injector/src/cli/generate.rs
@@ -1,4 +1,6 @@
-use crate::{receipt_txn::handle_report_msg, Result, Settings, LOADER_WORKERS, STORE_WORKERS};
+use crate::{
+    receipt_txn::handle_report_msg, Error, Result, Settings, LOADER_WORKERS, STORE_WORKERS,
+};
 use chrono::{NaiveDateTime, TimeZone, Utc};
 use file_store::{FileStore, FileType};
 use futures::{
@@ -6,7 +8,7 @@ use futures::{
     TryStreamExt,
 };
 use helium_crypto::Keypair;
-use helium_proto::{blockchain_txn::Txn, BlockchainTxn, Message};
+use helium_proto::{BlockchainTxn, Message};
 use std::sync::{Arc, Mutex};
 
 /// Generate poc rewards
@@ -46,10 +48,7 @@ impl Cmd {
                 let success_counter_ref = Arc::clone(&success_counter);
                 let failure_counter_ref = Arc::clone(&failure_counter);
                 async move {
-                    if process_msg(msg, shared_key_clone, before_ts)
-                        .await
-                        .is_some()
-                    {
+                    if process_msg(msg, shared_key_clone, before_ts).await.is_ok() {
                         *success_counter_ref.lock().unwrap() += 1;
                     } else {
                         *failure_counter_ref.lock().unwrap() += 1;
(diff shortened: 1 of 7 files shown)


poc_iot_injector/src/settings.rs
```diff
@@ -68,4 +72,10 @@ impl Settings {
     pub fn trigger_interval(&self) -> Duration {
         Duration::from_secs(self.trigger)
     }
+
+    pub fn receipt_sender_receiver(
```
jeffgrunewald, line 76: i don’t think this belongs here since it’s doesn’t really have anything to do with settings or use them at all; I can understand the argument for something like the database connections that use the settings but this just seems like the message channel could be called directly from within the module where the sink is established

vihu: Right, yes, fixed in 315f51c

andymck (approved):

poc_iot_injector/src/cli/generate.rs
```diff
@@ -35,22 +38,24 @@ impl Cmd {
 
         store
             .source_unordered(LOADER_WORKERS, stream::iter(file_list).map(Ok).boxed())
-            .for_each_concurrent(STORE_WORKERS, |msg| {
+            .try_for_each_concurrent(STORE_WORKERS, |msg| {
```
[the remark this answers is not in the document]
jeffgrunewald: i'd also like to understand the motivation/expected behavior of a stream of pocs to be processed here. once the poc has been validated (correlated beacons and witnesses) does order matter such that we should halt the processing of all reports and resolve any issues when we encounter them vs. continuing to process and reward on valid ones and write invalid reports or throw errors/alerts as side effects to troubleshoot and cleanup out-of-band any errors encountered?

poc_iot_injector/src/cli/generate.rs
```diff
@@ -33,24 +36,41 @@ impl Cmd {
...
+
+        tracing::debug!("success_counter: {:?}", success_counter.lock().unwrap());
+        tracing::debug!("failure_counter: {:?}", failure_counter.lock().unwrap());
 
         Ok(())
     }
 }
 
-async fn process_msg(msg: prost::bytes::BytesMut, shared_key_clone: Arc<Keypair>, before_ts: i64) {
+async fn process_msg(
```
jeffgrunewald, line 69: if we are moving to a `try_` version of the stream processor, would it be more idiomatic for this function to return a result rather than an option? the signature for these closures doesn't look like they expect the function to yield an option, so perhaps this is a carryover from an earlier version where we were iterating/streaming these synchronously?

poc_iot_injector/src/cli/generate.rs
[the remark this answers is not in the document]
vihu: Ideally, the iot-verifier would/should only ever output valid pocs/witness reports imho. However, currently we have some instances (a lot actually, which is why I added a counter for inspection in [99dd684](https://github.com/helium/oracles/pull/167/commits/99dd684680232c2393d8167ef9eee8ede7dcfb8e)) of invalid protos put in the bucket. I believe @andymck will cleanup that bucket up once some smaller PRs land.

As for the `try_for_each_concurrent`, it's mostly the same as `for_each_concurrent` except that it would exit early if there is a stream item which it is not able to process. Secondly, `try_for_each_concurrent` allows you to use `?` operators within the async move. Finally, it allows having the internal async closure return an Ok/Err variant.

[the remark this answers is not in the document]
jeffgrunewald: Allowing for results inside the closure makes sense

poc_iot_injector/src/cli/generate.rs
vihu: Yeah, simplified in [6e24642](https://github.com/helium/oracles/pull/167/commits/6e2464299930875f0e1d95b0e5d7f23449815140). Don't really like the `Result<Option<Thing>>` pattern, there's a specific `TxnConstruction` error enum now, and the result type is simplified to reflect the same.

poc_iot_injector/src/cli/generate.rs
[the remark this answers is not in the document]
madninja: > Ideally, the iot-verifier would/should only ever output valid pocs/witness reports imho

wut? The verifier should output both valid and invalid reports for audibility right?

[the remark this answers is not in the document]
vihu: Oh yes, I mean the `LoraValidPoc` struct specifically should have valid construction using the defined, which is not the case right now. The injector doesn't care about any invalid POCs.

poc_iot_injector/src/cli/generate.rs
```diff
@@ -33,33 +38,45 @@ impl Cmd {
...
+                        *failure_counter_ref.lock().unwrap() += 1;
                     }
+                    Ok(())
                 }
             })
-            .await;
+            .await?;
+
+        tracing::debug!("success_counter: {:?}", success_counter.lock().unwrap());
+        tracing::debug!("failure_counter: {:?}", failure_counter.lock().unwrap());
```
jeffgrunewald, line 62:
```suggestion
        tracing::debug!("success count: {:?}; failure count: {:?}", success_counter.lock().unwrap(), failure_counter.lock().unwrap());
```
are these unwraps more or less guaranteed safe operations since we're just handling the handoff of a mutex lock and trusting the compiler to negotiate that properly or do we just not care as much since it's a cli and not a production running server process?

poc_iot_injector/src/server.rs
```diff
@@ -111,20 +117,25 @@ async fn submit_txns(
...
+                    handle_txn_submission(msg, shared_key, &mut shared_txn_service, before_ts).await
+                {
+                    // NOTE: Can the txn be written as is?
+                    match file_sink_write!("signed_poc_receipt_txn", receipt_sender, txn).await {
+                        Ok(_) => tracing::debug!("txn written!"),
+                        Err(_) => tracing::error!("unable to write txn"),
                     }
-                    Err(e) => tracing::error!("unable to process msg due to: {:?}", e),
+                } else {
+                    tracing::error!("unable to submit txn")
```
jeffgrunewald, line 133: good opportunity for @xandkar to swoop in and send this to a prometheus gauge?

jeffgrunewald: or else similar to how we were handling rewards submissions in the mobile rewards server for a while you could just change this up so `handle_submission_txn` returns a result instead of an option and then both `handle_txn_submission` and `file_sink_write!` are just using the `?` operator since the file sink write macro is already emitting debug and error traces on success and failure and that's the only thing we're currently doing in the above

poc_iot_injector/src/cli/generate.rs
vihu: Yeah pretty much, I just added the counters for inspection. The `generate` cli isn't meant for prod.

poc_iot_injector/src/server.rs
vihu: Updated 3667dd8

poc_iot_injector/src/server.rs
```diff
@@ -100,9 +107,10 @@ async fn submit_txns(
     txn_service: &mut TransactionService,
     store: &mut FileStore,
     keypair: Arc<Keypair>,
+    receipt_sender: &file_sink::MessageSender,
     after_utc: DateTime<Utc>,
     before_utc: DateTime<Utc>,
-) -> Result {
+) -> Result<()> {
```
jeffgrunewald, line 113: isn't this covered by the definition of result in the error module?
```
pub type Result<T = ()> = std::result::Result<T, Error>;
```

poc_iot_injector/src/server.rs
```diff
@@ -111,20 +119,19 @@ async fn submit_txns(
...
-                    Ok(m) => {
-                        handle_txn_submission(m, shared_key, &mut shared_txn_service, before_ts)
-                            .await
-                    }
-                    Err(e) => tracing::error!("unable to process msg due to: {:?}", e),
+                if let Ok(txn) =
+                    handle_txn_submission(msg, shared_key, &mut shared_txn_service, before_ts).await
+                {
+                    file_sink_write!("signed_poc_receipt_txn", receipt_sender, txn).await?;
                 }
```
jeffgrunewald, line 130:
```suggestion
                handle_txn_submission(msg, shared_key, &mut shared_txn_service, before_ts).await?
                file_sink_write!("signed_poc_receipt_txn", receipt_sender, txn).await?
```
am i misunderstanding the behavior of the `?` in error handling or would the above give you the same conditional execution of the file_sink write b/c a failed txn submission would bail out to the caller first?

vihu: It should, but for some reason doing so makes the error code angry. Not quite sure how to fix tbh.

poc_iot_injector/src/server.rs
vihu: Oh yeah, missed it. I'll update.

poc_iot_injector/src/server.rs
jeffgrunewald: yeah, fair point; it's weird that it's trying to cast the error in `handle_txn_submission` _to_ the `file_store::Error` instead of the other way around; i suspect it's an oddity of the `file_sink_write!` being a macro and co-opting the way the errors typically propogate?

jeffgrunewald: maybe something to look into as a separate issue

jeffgrunewald (approved):

jeffgrunewald (approved):

bbalser (approved):

PR #175, opened 2022-11-01
move old speedtest filter to iterator transformer

jeffgrunewald:
moves the transformation of speedtest iterators from an impl method on the SpeedtestAverages struct to a free function that takes and returns an `Iterator<Item = &'a Speedtest>`. this allows us to keep the logic of the transformation isolated to the speedtests module and call it within the publicly exported functions that execute speedtest calculations elsewhere in the oracle such as when computing the subnetwork rewards.


maplant (approved):
very nice!

bbalser (approved):

PR #178, opened 2022-11-01
Convert validation of heartbeats and speedtests to use a stream

maplant:
This one is quite a doozy. If you have any questions be sure to ask, the mechanics of getting async working was pretty complicated so there's a lot of teachable stuff here.

One particularly proud moment was completely avoiding any use of "box".

### DIFF
--- a/poc_mobile_verifier/src/ingest.rs
+++ b/poc_mobile_verifier/src/ingest.rs
@@ -1,0 +1,74 @@
+use crate::Result;
+use chrono::{DateTime, Utc};
+use file_store::{
+    heartbeat::{CellHeartbeat, CellHeartbeatIngestReport},
+    speedtest::{CellSpeedtest, CellSpeedtestIngestReport},
+    traits::MsgDecode,
+    FileStore, FileType,
+};
+use futures::{stream, Stream, StreamExt};
+use std::ops::Range;
+
+pub async fn ingest_heartbeats(
+    file_store: &FileStore,
+    epoch: &Range<DateTime<Utc>>,
+) -> Result<impl Stream<Item = CellHeartbeat>> {
+    Ok(file_store
+        .source(
+            stream::iter(
+                file_store
+                    .list_all(FileType::CellHeartbeatIngestReport, epoch.start, epoch.end)
+                    .await?,
+            )
+            .map(Ok)
+            .boxed(),
+        )
+        .filter_map(|msg| async move {
+            msg.map_err(|err| {
+                tracing::error!("Error fetching heartbeat ingest report: {:?}", err);
+                err
+            })
+            .ok()
+        })
+        .filter_map(|msg| async move {
+            CellHeartbeatIngestReport::decode(msg).map_or_else(
+                |err| {
+                    tracing::error!("Could not decode cell heartbeat ingest report: {:?}", err);
+                    None
+                },
+                |report| Some(report.report),
+            )
+        }))
+}
+
+pub async fn ingest_speedtests(
+    file_store: &FileStore,
+    epoch: &Range<DateTime<Utc>>,
+) -> Result<impl Stream<Item = CellSpeedtest>> {
+    Ok(file_store
+        .source(
+            stream::iter(
+                file_store
+                    .list_all(FileType::CellSpeedtestIngestReport, epoch.start, epoch.end)
+                    .await?,
+            )
+            .map(Ok)
+            .boxed(),
+        )
+        .filter_map(|msg| async move {
+            msg.map_err(|err| {
+                tracing::error!("Error fetching speedtest ingest report: {:?}", err);
+                err
+            })
+            .ok()
+        })
+        .filter_map(|msg| async move {
+            CellSpeedtestIngestReport::decode(msg).map_or_else(
+                |err| {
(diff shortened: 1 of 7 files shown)


poc_mobile_verifier/src/speedtests.rs
```diff
@@ -70,6 +71,92 @@ impl SpeedtestRollingAverage {
...
+                stream::iter(
+                    file_store
+                        .list_all(FileType::CellSpeedtestIngestReport, epoch.start, epoch.end)
+                        .await?,
+                )
+                .map(Ok)
+                .boxed(),
+            )
+            .filter_map(move |msg| {
+                let speedtests = speedtests.clone();
```
madninja, line 91: why move followed by async move?

poc_mobile_verifier/src/speedtests.rs
```diff
@@ -70,6 +71,92 @@ impl SpeedtestRollingAverage {
...
+                async move {
+                    let msg = if let Ok(msg) = msg {
```
madninja, line 93: why not do this in the `move` section?

poc_mobile_verifier/src/verifier.rs
```diff
@@ -75,24 +76,34 @@ impl VerifierDaemon {
...
-        }
-
-        for speedtest in speedtests.clone().without_lapsed().into_iter() {
-            speedtest.write(&self.speedtest_avg_tx).await?;
-        }
-
-        for speedtest in speedtests.into_iter() {
-            speedtest.save(&mut transaction).await?;
-        }
+        let transaction = {
```
madninja, line 82: why the two transaction assignments?

poc_mobile_verifier/src/verifier.rs
```diff
@@ -171,9 +187,9 @@ impl Verifier {
     }
 }
 
-pub struct VerifiedEpoch {
-    pub heartbeats: Vec<Heartbeat>,
-    pub speedtests: SpeedtestAverages,
+pub struct VerifiedEpoch<H, S> {
```
madninja, line 190: huh.. couldn't declare them as the actual Stream Impls they're going to be?

maplant: They can't be named since their type decls include closures. Unless we used a boxed stream, but I wanted to see how far I could get without boxing

poc_mobile_verifier/src/speedtests.rs
maplant: The first move moves the Arc into the FnMut closure, the second moves the clone of the Arc into the async block

poc_mobile_verifier/src/verifier.rs
maplant: The transaction is moved into the `try_fold`, the result of which is the same transaction - so it needs to be re-assigned

bbalser:
just need unit tests and I think it is ready to go.

poc_mobile_verifier/src/ingest.rs
```diff
@@ -0,0 +1,74 @@
...
+        .filter_map(|msg| async move {
+            msg.map_err(|err| {
+                tracing::error!("Error fetching heartbeat ingest report: {:?}", err);
+                err
+            })
+            .ok()
+        })
+        .filter_map(|msg| async move {
+            CellHeartbeatIngestReport::decode(msg).map_or_else(
+                |err| {
```
jeffgrunewald, line 35: i didn't think the default closure in `map_or_else` took any arguments?

maplant: It doesn't if the type is an option.

poc_mobile_verifier/src/verifier.rs
```diff
@@ -147,14 +149,28 @@ impl Verifier {
...
-    ) -> Result<VerifiedEpoch> {
-        let heartbeats = Heartbeat::validate_heartbeats(&self.file_store, epoch).await?;
-        let speedtests =
-            SpeedtestAverages::validate_speedtests(pool, &self.file_store, epoch).await?;
+        pool: impl SpeedtestStore + Copy + 'a,
+        epoch: &'a Range<DateTime<Utc>>,
+    ) -> Result<
+        VerifiedEpoch<
+            impl Stream<Item = Heartbeat> + 'a,
+            impl Stream<Item = Result<SpeedtestRollingAverage>> + 'a,
```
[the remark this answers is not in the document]
bbalser: I like it returning a result because of the db call.  Most likely as we add more features there will be DB calls for heartbeats as well, so that will get converted to a Result.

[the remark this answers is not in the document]
jeffgrunewald: Is a db call any more likely to fail than any other step we already perform on heartbeat messages from ingest to reward verification?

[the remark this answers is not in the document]
maplant: It's not that it's more likely, it is that if the DB call fails we want to cancel the whole transaction. We want to handle the errors differently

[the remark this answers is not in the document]
jeffgrunewald: tricky use of `transpose()?` there

bbalser (approved):

jeffgrunewald (approved):

PR #182, opened 2022-11-03
Fix poc_report join query and add entropy_start_time migration

vihu:
Fixes some bugs with the `poc_report` query we found during deployment.

### DIFF
--- a/poc_iot_verifier/migrations/6_add_entropy_start_time.sql
+++ b/poc_iot_verifier/migrations/6_add_entropy_start_time.sql
@@ -1,0 +1,1 @@
+alter table poc_report add entropy_start_time timestamptz default now() not null;
--- a/poc_iot_verifier/migrations/6_add_entropy_start_time.sql
+++ b/poc_iot_verifier/migrations/6_add_entropy_start_time.sql
@@ -1,1 +1,7 @@
+-- :up
+
 alter table poc_report add entropy_start_time timestamptz default now() not null;
+
+-- :down
+
+alter table poc_report drop entropy_start_time;
--- a/poc_iot_verifier/src/poc_report.rs
+++ b/poc_iot_verifier/src/poc_report.rs
@@ -127,7 +127,7 @@ impl Report {
     {
         sqlx::query_as::<_, Self>(
             r#"
-            select  id,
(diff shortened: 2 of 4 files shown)


bbalser (approved):

poc_iot_verifier/migrations/6_add_entropy_start_time.sql
```diff
@@ -0,0 +1 @@
+alter table poc_report add entropy_start_time timestamptz default now() not null;
```
jeffgrunewald, line 1: does this need any sort of explicit rollback step or can it be inferred completely from the migration?

vihu: Hmm, good q. Probably best to add rollback I suppose. But it did work as is in the live deployment we tried.

poc_iot_verifier/src/poc_report.rs
```diff
@@ -127,7 +127,7 @@ impl Report {
     {
         sqlx::query_as::<_, Self>(
             r#"
-            select  id,
+            select poc_report.id,
```
jeffgrunewald, line 130: do we need the `poc_report` prefix on any of these now that it's the only table we're returning results from and we're explicitly setting the select query `from poc_report`?
would the `poc_report.entropy_start_time` be the same as the `entropy.timestamp` such that your filter could explicitly match
```
and entropy.timestamp = poc_report.entropy_start_time` instead of `entropy.timestamp < (NOW - Interval)`?
```

vihu: Not sure about that. @andymck might know. We did the changes just to get it up and running live.

poc_iot_verifier/migrations/6_add_entropy_start_time.sql
vihu: [1a47c12](https://github.com/helium/oracles/pull/182/commits/1a47c1266c0e3ceac9214472bf4f9744b245a734) Can confirm that this works locally with sqlx-cli

xandkar (approved):

jeffgrunewald (approved):

PR #185, opened 2022-11-07
Entropy related fixes

andymck:
Handles the entropy timestamp field being optionally returned from various poc report queries.  Defaults to Option None if not present.

Fixes some typos in queries

### DIFF
--- a/poc_iot_verifier/src/runner.rs
+++ b/poc_iot_verifier/src/runner.rs
@@ -148,7 +148,7 @@ impl Runner {
         tracing::info!("found {beacon_len} beacons ready for verification");
         for db_beacon in db_beacon_reports {
             let packet_data = &db_beacon.packet_data;
-            let entropy_start_time = db_beacon.entropy_start_time;
+            let entropy_start_time = db_beacon.timestamp;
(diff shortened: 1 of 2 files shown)


dpezely (approved):

poc_iot_verifier/src/runner.rs
```diff
@@ -147,8 +147,12 @@ impl Runner {
         let beacon_len = db_beacon_reports.len();
         tracing::info!("found {beacon_len} beacons ready for verification");
         for db_beacon in db_beacon_reports {
+            let entropy_start_time = match db_beacon.timestamp {
+                Some(v) => v,
+                None => return Ok(())
```
jeffgrunewald, line 152: this is going to stop processing _all_ the beacons in the set returned; do you want to do that or just skip over the invalid beacon and continue pulling associated witnesses for the valid ones?

andymck: bah, yeah it should just skip the current one.  Fix incoming

andymck: fixed here: f0a987d14808969b49debf76c98da275f91d40c8

vihu (approved):

PR #187, opened 2022-11-08
Add an additional setting for txn submission

vihu:
This adds an extra setting `do_submission` boolean flag to specify whether to submit txn to mainnet. Mostly because we want to start the injector but don't really want to start submitting txns to mainnet just yet.

### DIFF
--- a/poc_iot_injector/src/server.rs
+++ b/poc_iot_injector/src/server.rs
@@ -1,6 +1,6 @@
 use crate::{
     receipt_txn::{handle_report_msg, TxnDetails},
-    Result, Settings, LOADER_WORKERS, STORE_WORKERS,
+    Error, Result, Settings, LOADER_WORKERS, STORE_WORKERS,
 };
 use chrono::{DateTime, NaiveDateTime, TimeZone, Utc};
 use db_store::MetaValue;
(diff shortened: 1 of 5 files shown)


poc_iot_injector/src/server.rs
```diff
@@ -123,10 +128,31 @@ async fn submit_txns(
...
-                    file_sink_write!("signed_poc_receipt_txn", receipt_sender, txn).await?;
+                match (
+                    handle_report_msg(msg, shared_key.clone(), before_ts),
+                    do_submission,
+                ) {
+                    (Ok(txn_details), true) => {
+                        // do txn submission
+                        if handle_txn_submission(txn_details.clone(), &mut shared_txn_service)
+                            .await
+                            .is_ok()
```
andymck, line 139: handle_txn_submission will always return an ok response.  Should a failure scenario not be considered, for example what if the submit node is down or unreachable ?  Any txn submitted during this period will be lost ( in terms of not making it to the chain ).

jeffgrunewald: at the moment when we don't want the txns submitted to the chain for real logging the error seems like a decent enough approach but do we have a plan for handling failed txn construction or submission? either to store and retry later, crash, or something else?

jeffgrunewald: although `handle_txn_submission` can fail, so it's not guaranteed to return an `Ok`

poc_iot_injector/src/server.rs
```diff
@@ -136,17 +162,23 @@ async fn submit_txns(
...
+        tracing::debug!(
+            "txn submitted successfully, hash: {:?}",
+            txn_details.hash_b64_url
+        );
     } else {
-        tracing::warn!("txn submission failed!, hash: {:?}", hash_b64_url);
-        Err(Error::TxnSubmission(hash_b64_url))
+        tracing::warn!(
+            "txn submission failed!, hash: {:?}",
+            txn_details.hash_b64_url
```
jeffgrunewald, line 180: i think i see @andymck 's point here (didn't read ahead); are we sure we want to take out this submission error here? when the `do_submission` boolean is `false` and we're just writing out the proposed txn without submitting it we get the validation of the process without submitting or triggering any rewards but when we deliberately set that boolean to true wouldn't we want some sort of guard in place here in case something goes wrong with that submission?

also with the unconditional `Ok(())` below the `is_ok()` check on the `if handle_txn_submission` above seems unnecessary; if this is just a temporary setup and `handle_txn_submission` meant to potentially return Ok _and_ Err responses to the caller can we get a todo/comment here to note for later address?

vihu: Yup, that's fair. I'll bubble up the error if the txn submission fails so the whole `submi_txn` should fail if the node is down/unreachable. So when the next tick happens the `last_poc_submission_ts` wouldn't have moved forward and we'd still process the txn which went unsubmitted.

vihu: @andymck @jeffgrunewald Updated [13d7cf0](https://github.com/helium/oracles/pull/187/commits/13d7cf0457a382784e702f087f403e87806add9c).

So now we'll exit early if there's any error in `handle_report_msg` or `handle_txn_submission`, which also means that `submit_txn` itself would exit thus not updating the `last_poc_submission_ts`, so at the next we should pick up the txns we were not able to process.

Note that we would run into duplicate submissions from the previous time range, but that's something the chain should be able to handle and drop.

andymck (approved):

PR #188, opened 2022-11-08
Remove cell_heartbeat and cell_speedtest from mobile ingest

bbalser:


jeffgrunewald (approved):
We just never used these I take it?

PR #190, opened 2022-11-08
Change timings of mobile pipeline

bbalser:
1. mobile ingest will now wait up to 15 minutes before uploading `heartbeat_report` or `speedtest_report` to S3
2. poc_mobile_verifier will wait up to 15 minutes before uploading `validated_heartbeat` or `speedtest_avg` to S3
3. poc_mobile_verifier will now run verify and reward cycles offset from the period by 30 minutes. Example: for the verification period 01:00 to 04:00, verification will run at 04:30
4. poc_mobile_verifier will flush rewards file and upload to S3 at end of reward cycle, not waiting for the rolling period, since it is known that there is no more data.

### DIFF
--- a/poc_mobile_verifier/src/scheduler.rs
+++ b/poc_mobile_verifier/src/scheduler.rs
@@ -1,11 +1,6 @@
 use crate::{Error, Result};
 use chrono::{DateTime, Duration, Utc};
-use lazy_static::lazy_static;
 use std::ops::Range;
-
-lazy_static! {
-    static ref OFFSET: Duration = Duration::minutes(30);
-}
 
 #[derive(Debug)]
 pub struct Scheduler {
@@ -13,6 +8,7 @@ pub struct Scheduler {
     pub reward_period_length: Duration,
     pub verification_period: Range<DateTime<Utc>>,
     pub reward_period: Range<DateTime<Utc>>,
+    pub verification_offset: Duration,
 }
 
 impl Scheduler {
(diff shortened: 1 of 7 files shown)


poc_mobile_verifier/src/scheduler.rs
```diff
@@ -1,7 +1,12 @@
 use crate::{Error, Result};
 use chrono::{DateTime, Duration, Utc};
+use lazy_static::lazy_static;
 use std::ops::Range;
 
+lazy_static! {
+    static ref OFFSET: Duration = Duration::minutes(30);
```
maplant, line 7: would we want to make this a field in the scheduler?

maplant (approved):

ingest/src/server_5g.rs
```diff
@@ -117,6 +116,7 @@ pub async fn grpc_server(shutdown: triggered::Listener, settings: &Settings) ->
         heartbeat_report_rx,
     )
     .deposits(Some(file_upload_tx.clone()))
+    .roll_time(Duration::minutes(15))
```
vihu, line 119: nit: Might be better to move this to a const `INGEST_WAIT_DURATION` or similar?

vihu: Maybe define it in `lib.rs` since its used in multiple places

dpezely (approved):
Approving in the interest of merging sooner than later... However, please mark the outstanding comments/questions as `TODO` for a future PR.

andymck (approved):

poc_mobile_verifier/src/scheduler.rs
jeffgrunewald: Seems like it could make for a good test setup if it’s a configurable value in the verifier settings

jeffgrunewald (approved):

ingest/src/server_5g.rs
bbalser: Fixed cbbeb29

PR #191, opened 2022-11-09
Jg/density scaler settings

jeffgrunewald:
Configure the density scaler via the `config` crate like the rest of the Oracles and instantiate the density_scaler as a child process of the poc_iot_verifier in the main loop `tokio::try_join`; when verifying beacons and witnesses, submit  the candidate hotspot's location to the density scaler and await the return decimal, set as the scaling factor of the report

### DIFF
--- a/file_store/src/lora_valid_poc.rs
+++ b/file_store/src/lora_valid_poc.rs
@@ -5,17 +5,22 @@ use crate::{
     Error, Result,
 };
 use chrono::{DateTime, Utc};
+use density_scaler::SCALING_PRECISION;
 use helium_proto::services::poc_lora::{
     LoraBeaconReportReqV1, LoraValidBeaconReportV1, LoraValidPocV1, LoraValidWitnessReportV1,
     LoraWitnessReportReqV1,
 };
+use rust_decimal::{prelude::ToPrimitive, Decimal};
+use rust_decimal_macros::dec;
 use serde::Serialize;
+
+const SCALE_MULTIPLIER: Decimal = dec!(100);
 
 #[derive(Serialize, Clone)]
 pub struct LoraValidBeaconReport {
     pub received_timestamp: DateTime<Utc>,
     pub location: Option<u64>,
-    pub hex_scale: f32,
+    pub hex_scale: Decimal,
     pub report: LoraBeaconReport,
 }
 
@@ -23,7 +28,7 @@ pub struct LoraValidBeaconReport {
 pub struct LoraValidWitnessReport {
     pub received_timestamp: DateTime<Utc>,
     pub location: Option<u64>,
-    pub hex_scale: f32,
+    pub hex_scale: Decimal,
     pub report: LoraWitnessReport,
 }
 
@@ -100,7 +105,7 @@ impl TryFrom<LoraValidBeaconReportV1> for LoraValidBeaconReport {
         Ok(Self {
             received_timestamp: v.timestamp()?,
             location: v.location.parse().ok(),
-            hex_scale: v.hex_scale,
+            hex_scale: Decimal::new(v.hex_scale as i64, SCALING_PRECISION),
             report: v
                 .report
                 .ok_or_else(|| Error::not_found("lora valid beacon report v1"))?
@@ -120,7 +125,7 @@ impl From<LoraValidBeaconReport> for LoraValidBeaconReportV1 {
                 .location
                 .map(|l| l.to_string())
                 .unwrap_or_else(String::new),
-            hex_scale: v.hex_scale,
+            hex_scale: (v.hex_scale * SCALE_MULTIPLIER).to_u32().unwrap_or(0),
             report: Some(report),
         }
     }
@@ -133,7 +138,7 @@ impl TryFrom<LoraValidWitnessReportV1> for LoraValidWitnessReport {
         Ok(Self {
             received_timestamp,
             location: v.location.parse().ok(),
-            hex_scale: v.hex_scale,
+            hex_scale: Decimal::new(v.hex_scale as i64, SCALING_PRECISION),
             report: v
                 .report
                 .ok_or_else(|| Error::not_found("lora valid witness port v1"))?
@@ -152,7 +157,7 @@ impl From<LoraValidWitnessReport> for LoraValidWitnessReportV1 {
                 .location
                 .map(|l| l.to_string())
                 .unwrap_or_else(String::new),
-            hex_scale: v.hex_scale,
+            hex_scale: (v.hex_scale * SCALE_MULTIPLIER).to_u32().unwrap_or(0),
             report: Some(report),
         }
     }
--- a/poc_iot_verifier/src/main.rs
+++ b/poc_iot_verifier/src/main.rs
@@ -1,5 +1,7 @@
 use clap::Parser;
-use poc_iot_verifier::{loader, purger, runner, Result, Settings};
+use density_scaler::Server as DensityScaler;
+use futures::TryFutureExt;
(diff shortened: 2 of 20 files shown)


poc_iot_verifier/src/main.rs
```diff
@@ -59,6 +60,9 @@ impl Server {
         let pool = settings.database.connect(2).await?;
         sqlx::migrate!().run(&pool).await?;
 
+        // Create the density scaler query messaging channel
+        let (density_tx, density_rx) = density_scaler::query_channel(50);
```
andymck, line 64: What is the advantage of having the verifier communicate with the density scaler over message channels ?  Why not just instantiate it from within the verifier runner for example ( like what I do with the DenyList )?  Is there a distinct benefit with the message channels approach ?

poc_iot_verifier/src/poc.rs
```diff
@@ -193,11 +202,14 @@ impl Poc {
                     let gw_info: GatewayInfo = witness_result.gateway_info.ok_or_else(|| {
                         Error::not_found("invalid FollowerGatewayResp for witness")
                     })?;
-                    // TODO: perform hex density check here for a valid witness
+                    let scaling_factor = density_queries
+                        .query(gw_info.location.unwrap_or_default().to_string())
+                        .await?
+                        .unwrap_or(dec!(1.0));
```
andymck, line 208: is defaulting to a value of 1.0 what we want here ?  Previously i had that value as a placeholder but we should confirm it is appropriate

andymck (approved):

poc_iot_verifier/src/poc.rs
jeffgrunewald: i'm not married to this; multiplying by 1 always feels like a safe default since it's effectively a no-op but if we'd like to floor the results when we can't find a gateway's location and/or generate a hex scaling factor for it that's cool too

poc_iot_verifier/src/main.rs
jeffgrunewald: mostly just my erlang/elixir "communicating processes" pattern that made sense to me for an independent process operating within the boundaries of the verifier but ultimately doing separate side work. also, conceptually it lets me develop it more independently of the verifier itself and could theoretically be independently tested more easily if i had a decent way to mock the gateway info stream from the node follower

poc_iot_verifier/src/poc.rs
```diff
@@ -163,16 +169,18 @@ impl Poc {
             GatewayStakingMode::Light => (),
         }
 
-        // TODO: insert hex scale lookup here
-        //       value hardcoded to 1.0 temporarily
+        // beaconer location is guaranteed to unwrap as we've already checked and returned early above when it's `None`
+        let scaling_factor = density_queries
+            .query(beaconer_info.location.unwrap().to_string())
```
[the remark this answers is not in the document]
jeffgrunewald: good point; i don't think there's a good sensible default to put here and we've already validated the gw has an asserted location so potentially throwing it back as an error seems like it would read as just confusing so i refactored it to use your earlier check for a location and just pull the value out then when it's available.

poc_mobile_verifier/src/subnetwork_rewards.rs
```diff
@@ -43,11 +42,11 @@ impl SubnetworkRewards {
             .map(|(pubkey, mut shares)| {
                 let speedmultiplier = speedtests
                     .get_average(&pubkey)
-                    .map_or(dec!(0.0), |avg| avg.reward_multiplier());
+                    .map_or(Decimal::default(), |avg| avg.reward_multiplier());
                 shares *= speedmultiplier;
                 (pubkey, shares)
             })
-            .filter(|(_pubkey, shares)| shares > &dec!(0.0))
+            .filter(|(_pubkey, shares)| shares > &Decimal::default())
```
JayKickliter, line 49: did clips suggest this change? I think the comparison against 0 is more clear. Or better yet (untested):

```suggestion
            .filter(|(_pubkey, shares)| shares > &Decimal::ZERO)
```

poc_mobile_verifier/src/subnetwork_rewards.rs
```diff
@@ -43,11 +42,11 @@ impl SubnetworkRewards {
             .map(|(pubkey, mut shares)| {
                 let speedmultiplier = speedtests
                     .get_average(&pubkey)
-                    .map_or(dec!(0.0), |avg| avg.reward_multiplier());
+                    .map_or(Decimal::default(), |avg| avg.reward_multiplier());
```
JayKickliter, line 45:
```suggestion
                    .map_or(Decimal::ZERO, |avg| avg.reward_multiplier());
```

poc_mobile_verifier/src/subnetwork_rewards.rs
jeffgrunewald: no clippy suggestion; i agree the zero constant is a clearer version of the same

jeffgrunewald: fixed [0d6e939](https://github.com/helium/oracles/pull/191/commits/0d6e93943e8c0f7ad74905f4b2a8e50eae0ea4b8)

poc_iot_verifier/src/poc.rs
andymck: My original concern here was mostly in the scenario whereby the query fails.  In any error scenario we default to 1.0.  An alternative is to fail the verification and allow it to be retried next tick but if a default to 1.0 is effectively a no-op then it should be fine as is

file_store/src/lora_valid_poc.rs
```diff
@@ -120,7 +123,7 @@ impl From<LoraValidBeaconReport> for LoraValidBeaconReportV1 {
                 .location
                 .map(|l| l.to_string())
                 .unwrap_or_else(String::new),
-            hex_scale: v.hex_scale,
+            hex_scale: (v.hex_scale * dec!(100)).to_u32().unwrap_or(0),
```
vihu, line 126: Maybe make a `SCALE_MULTIPLIER` const with `dec!(100)` with a docstring?

poc_iot_verifier/src/poc.rs
```diff
@@ -215,11 +225,14 @@ impl Poc {
                     let gw_info: GatewayInfo = witness_result.gateway_info.ok_or_else(|| {
                         Error::not_found("invalid FollowerGatewayResp for witness")
                     })?;
-                    // TODO: perform hex density check here for a valid witness
+                    let scaling_factor = density_queries
+                        .query(gw_info.location.unwrap_or_default().to_string())
+                        .await?
+                        .unwrap_or(dec!(1.0));
```
vihu, line 231:
```suggestion
                        .unwrap_or(Decimal::ONE);
```

bbalser (approved):

poc_iot_verifier/src/poc.rs
jeffgrunewald: Nice, i missed this constant

jeffgrunewald: [3a99e29](https://github.com/helium/oracles/pull/191/commits/3a99e2995a48b124f65812a2e7cb9b90a425b969)

file_store/src/lora_valid_poc.rs
jeffgrunewald: [3a99e29](https://github.com/helium/oracles/pull/191/commits/3a99e2995a48b124f65812a2e7cb9b90a425b969)

vihu (approved):

PR #195, opened 2022-11-09
Convert verifier output to amount per radio reward

maplant:
**Important note:** This change subtly changes the output of rewards (see the corresponding assert changes in this PR) as the method of calculating rewards is done per radio as opposed to per owner. I tried using a higher precision floating point library and got the same results.


poc_mobile_verifier/src/ingest.rs
```diff
@@ -14,13 +14,9 @@ pub async fn ingest_heartbeats(
...
     Ok(file_store
         .source(
-            stream::iter(
-                file_store
-                    .list_all(FileType::CellHeartbeatIngestReport, epoch.start, epoch.end)
-                    .await?,
-            )
-            .map(Ok)
-            .boxed(),
+            file_store
```
jeffgrunewald, line 17: no longer streaming these results?

mobile_rewards/src/server.rs
```diff
@@ -267,48 +247,98 @@ impl Server {
         Ok(())
     }
 
-    async fn handle_rewards(
-        &mut self,
-        reward_period: Range<u64>,
```
jeffgrunewald, line 272: this reward period is still needed to correctly set the start and end _block height_ in the subnetwork reward transaction.

mobile_rewards/src/server.rs
```diff
@@ -267,48 +247,98 @@ impl Server {
...
-                .await?;
-        } else {
-            tracing::error!("cannot continue; unable to determine reward period!");
-            return Err(Error::NotFound("invalid reward period".to_string()));
+        for (epoch, rewards) in owner_rewards.into_iter() {
+            let rewards = rewards
+                .into_iter()
+                .map(|(account, amount)| SubnetworkReward { account, amount })
+                .collect();
+            self.issue_rewards(rewards, epoch.0..epoch.1).await?;
```
jeffgrunewald, line 336: this is where you still need to pass in the "reward_period" range of block heights to construct the txn. this range is retrieved at the top of `process_clock_tick` by querying the follower which gets it from the chain

poc_mobile_verifier/src/owner_shares.rs
```diff
@@ -1,94 +1,170 @@
...
+pub struct ResolveError(#[from] tonic::Status);
+
+#[async_trait::async_trait]
+pub trait OwnerResolver: Send {
+    async fn resolve_owner(
+        &mut self,
+        address: &PublicKey,
+    ) -> Result<Option<PublicKey>, ResolveError>;
+
+    async fn owner_shares(
```
jeffgrunewald, line 123: i'm not sure i like hanging this off the OwnerResolver impl b/c in practice it means we're calling it against the follower client which just reads as confusing; it implies the follower is calculating the owner shares. since this is just calling the `resolve_owner` method of the impl this feels like it should just be an independent function with the impl of the `OwnerResolver` as one of the parameters

poc_mobile_verifier/src/cli/reward_from_db.rs
```diff
@@ -32,18 +31,24 @@ impl Cmd {
...
-        let follower = settings.follower.connect_follower()?;
+        let mut follower = settings.follower.connect_follower()?;
         let pool = settings.database.connect(10).await?;
 
         let heartbeats = Heartbeats::validated(&pool, epoch.start).await?;
         let speedtests = SpeedtestAverages::validated(&pool, epoch.end).await?;
-        let rewards =
-            SubnetworkRewards::from_epoch(follower, &epoch, heartbeats, speedtests.clone()).await?;
+        let owner_shares = follower
+            .owner_shares(heartbeats, speedtests.clone())
```
jeffgrunewald, line 40: see below; i think implementing this method against the `OwnerResolver` makes this looks like the follower is doing the calculations

file_store/src/file_sink.rs
```diff
@@ -384,6 +407,28 @@ impl FileSink {
     }
 }
 
+/// List of all files written by a file sink
+#[derive(Clone, Debug, Default)]
+pub struct Manifest {
+    written_files: Arc<Mutex<Vec<String>>>,
```
bbalser, line 413: We could avoid the `Arc<Mutex` if we retrieve the manifest list by sending a message to the file_sink, which then could send the results back in a oneshot.  This should also mean we don't need the flush message to send anything back on a oneshot.

poc_mobile_verifier/src/cli/server.rs
```diff
@@ -53,14 +53,28 @@ impl Cmd {
...
+        let (radio_rewards_tx, radio_rewards_rx) = file_sink::message_channel(50);
         let mut subnet_rewards = file_sink::FileSinkBuilder::new(
-            FileType::SubnetworkRewards,
+            FileType::RadioRewardShare,
             store_base_path,
-            subnet_rewards_rx,
+            radio_rewards_rx,
         )
         .deposits(Some(file_upload_tx.clone()))
+        .manifest(reward_manifest.clone())
```
bbalser, line 65: Do we not want the same 15 roll time on this sink?

poc_mobile_verifier/src/ingest.rs
bbalser: file_store.list returns a FileInfoStream, so before we were turning a stream into a Vec to turn it back into a stream.

poc_mobile_verifier/src/owner_shares.rs
```diff
@@ -1,94 +1,170 @@
...
-            subnet_rewards_tx,
-            proto::SubnetworkRewards {
-                start_epoch: self.epoch.start.timestamp() as u64,
-                end_epoch: self.epoch.end.timestamp() as u64,
-                rewards: self.rewards,
-            }
+        epoch: &'_ Range<DateTime<Utc>>,
+    ) -> impl Iterator<Item = proto::RadioRewardShare> + '_ {
+        let total_shares = self.total_shares();
+        let total_rewards = get_scheduled_tokens(epoch.start, epoch.end - epoch.start).unwrap();
```
bbalser, line 67: seems like we should do something more than just `unwap` here

poc_mobile_verifier/src/owner_shares.rs
bbalser: It does feel weird to have `OwnerResolver` calculate the owner shares.  Maybe it is just a naming thing?

mobile_rewards/src/server.rs
```diff
@@ -267,48 +247,98 @@ impl Server {
...
-        while let Some(Ok(msg)) = stream.next().await {
-            let subnet_rewards = SubnetworkRewards::decode(msg);
-            if let Ok(subnet_rewards) = subnet_rewards {
-                // We only care that the verified epoch ends in our rewards epoch,
-                // so that we include verified epochs that straddle two rewards epochs.
-                if time_range.contains(&Utc.timestamp(subnet_rewards.end_epoch as i64, 0)) {
-                    rewards.extend(subnet_rewards.rewards);
-                }
-            }
+            .source_unordered(5, manifests)
```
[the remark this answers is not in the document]
bbalser: I don't understand the need for concurrency will reading the manifest files, the majority of the time, there will be one file.

mobile_rewards/src/server.rs
maplant: I literally keep forgetting this every single time

mobile_rewards/src/server.rs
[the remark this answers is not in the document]
maplant: This is a good point. I'm going to change this logic to only deal with one manifest at a time

file_store/src/file_sink.rs
```diff
@@ -347,6 +367,9 @@ impl FileSink {
         if let Some(deposits) = &self.deposits {
             file_upload::upload_file(deposits, &target_path).await?;
         }
+        if let Some(ref manifest) = self.manifest {
+            manifest.add_file(&target_path);
```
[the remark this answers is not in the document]
maplant: yeah, fixed that with the new implementation

mobile_rewards/src/server.rs
```diff
@@ -267,48 +247,98 @@ impl Server {
...
+            })
+            .filter_map(|msg| async move {
+                RadioRewardShare::decode(msg).map_or_else(
+                    |err| {
+                        tracing::error!("Could not decode reward share: {:?}", err);
+                        None
+                    },
+                    Some,
+                )
+            });
```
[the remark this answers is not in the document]
maplant: It can - I have an MR that never got anywhere to do that

poc_mobile_verifier/src/owner_shares.rs
maplant: Fixed, move this into `OwnerShares::aggregate`

poc_mobile_verifier/src/owner_shares.rs
maplant: Maybe an `expect`? This error can only occur if the system clock is broken

mobile_rewards/src/server.rs
```diff
@@ -267,48 +247,98 @@ impl Server {
...
+                RewardManifest::decode(msg).map_or_else(
+                    |err| {
+                        tracing::error!("Could not decode reward manifest: {:?}", err);
+                        None
+                    },
+                    Some,
+                )
+            })
+            .filter(move |manifest| {
+                future::ready(manifest.start_timestamp >= (last_reward_manifest as u64))
```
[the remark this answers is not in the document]
maplant: None now!

file_store/src/file_sink.rs
```diff
@@ -274,8 +301,19 @@ impl FileSink {
...
-                    Some(Message::Flush) => self.flush().await?,
+                    }
+                    Some(Message::Flush(on_flush_tx)) => {
+                        let res = self.flush().await;
+                        let _ = on_flush_tx.send(res);
+                    }
+                    Some(Message::FetchManifest(on_fetch_tx)) => {
+                        let _ = self.flush().await;
+                        if let Some(ref mut manifest) = self.manifest {
+                            let _ = on_fetch_tx.send(Ok(mem::take(manifest)));
```
bbalser, line 312: NIT: Can you use `manifest.take()` instead?

poc_mobile_verifier/src/verifier.rs
```diff
@@ -106,10 +109,31 @@ impl VerifierDaemon {
...
+        let rewards = self.verifier.reward_epoch(heartbeats, speedtests).await?;
+
+        for reward_share in rewards.into_radio_shares(&scheduler.reward_period) {
+            file_sink_write!("radio_reward_shares", &self.radio_rewards_tx, reward_share)
+                .await?
+                // Await the returned one shot to ensure that we wrote the file
+                .await??;
+        }
+
+        let written_files = file_sink::fetch_manifest(&self.radio_rewards_tx)
```
bbalser, line 121: Don't you need to flush the radio rewards sink before you fetch the manifest?

poc_mobile_verifier/src/owner_shares.rs
bbalser: `expect` feels better, at least we make sure the string clearly tells us what happened.

madninja: why have either? It _crashes_ the current thread, with unknown side effects.. Why not return a Result of an Impl?

poc_mobile_verifier/src/owner_shares.rs
```diff
@@ -240,21 +315,15 @@ mod test {
...
-        .map(|p| (PublicKey::try_from(p.account).unwrap(), p.amount))
-        .collect();
+        let owner_rewards = OwnerShares::aggregate(&mut resolver, heartbeats, speedtest_avgs)
+            .await
+            .expect("Could not generate rewards");
 
         // The owner with two hotspots gets more rewards
-        assert!(owner_rewards.get(&owner1).unwrap() > owner_rewards.get(&owner2).unwrap());
+        assert!(
+            owner_rewards.shares.get(&owner1).unwrap().total_shares()
```
madninja, line 324: can you replace these with expects to make the test clearer when it fails?

madninja (changes requested):

poc_mobile_verifier/src/verifier.rs
bbalser: Nevermind, I see it now that we are auto flushing when we fetch manifests

file_store/src/file_sink.rs
maplant: Nope - `take` doesn't do the same thing as `Option::take` for Vecs

poc_mobile_verifier/src/cli/server.rs
maplant: Fixed

poc_mobile_verifier/src/owner_shares.rs
maplant: yup

poc_mobile_verifier/src/owner_shares.rs
maplant: yeah might as well make it an error

jeffgrunewald (approved):
Not sure we still need the follower/owner resolver/share aggregator thing to be mutable but otherwise looks good

poc_mobile_verifier/src/owner_shares.rs
```diff
@@ -1,94 +1,175 @@
...
-impl SubnetworkRewards {
-    pub async fn from_epoch(
-        mut follower_service: impl OwnerResolver,
-        epoch: &Range<DateTime<Utc>>,
+const REWARDS_PER_SHARE_PREC: u32 = 9;
+const MOBILE_SCALE: u32 = 100_000_000;
+
+impl OwnerShares {
+    pub async fn aggregate(
+        resolver: &mut impl OwnerResolver,
```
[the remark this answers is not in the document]
maplant: we could take it by owner - but that would require a clone. `find_gateway` take a mutable reference

poc_mobile_verifier/src/owner_shares.rs
```diff
@@ -1,94 +1,175 @@
...
+        let mut owner_shares = Self::default();
         for heartbeat in heartbeats.into_iter() {
-            *hotspot_shares
-                .entry(heartbeat.hotspot_key.clone())
-                .or_default() += heartbeat.reward_weight;
+            if let Some(owner) = resolver.resolve_owner(&heartbeat.hotspot_key).await? {
+                let speedmultiplier = speedtests
+                    .get_average(&heartbeat.hotspot_key)
+                    .as_ref()
+                    .map_or(dec!(0.0), Average::reward_multiplier);
```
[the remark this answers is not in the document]
maplant: 👍

bbalser (approved):

madninja (approved):

PR #196, opened 2022-11-10
Fix time comparison when determining if file should roll based on time

bbalser:


madninja (approved):

dpezely (approved):

jeffgrunewald (approved):

file_store/src/file_sink.rs
```diff
@@ -294,7 +294,7 @@ impl FileSink {
 
     pub async fn maybe_roll(&mut self) -> Result {
         if let Some(active_sink) = self.active_sink.as_mut() {
-            if active_sink.time + self.roll_time > Utc::now() {
+            if active_sink.time + self.roll_time < Utc::now() {
```
jeffgrunewald, line 297: i assume these timers aren't so precise that we'd need this check to be `<=`? i.e. it's not likely to hit directly on `active_sink.time + self.roll_time == Utc::now()` and keep cycling?

vihu, line 297:
```suggestion
            if (active_sink.time + self.roll_time) < Utc::now() {
```

jeffgrunewald (approved):

file_store/src/file_sink.rs
```diff
@@ -16,6 +16,12 @@ use tokio::{
 use tokio_util::codec::{length_delimited::LengthDelimitedCodec, FramedWrite};
 
 pub const DEFAULT_SINK_ROLL_MINS: i64 = 3;
+
+#[cfg(not(test))]
+pub const DEFAULT_SINK_CHECK_MILLIS: i64 = 60_000;
```
madninja, line 21: Rename without DEFAULT prefix since this is not looked up anywhere

bbalser: Fixed 536f3c1

PR #199, opened 2022-11-10
Handle oracle packaging differently

jadeallenx:
Problem to solve: The current build environment is slow and has bit rotted to the point of not working or hardly working. 

Solution: Use a different solution for our debian packaging pipeline - `fpm` and rearrange certain build tasks before packaging to avoid "double building." Finally rust cache GHA seems to speed up anything and breaks portions of packaging. Eliminate it.


.github/scripts/make_debian.sh
```diff
@@ -0,0 +1,114 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+cd $GITHUB_WORKSPACE
```
dpezely, line 5: Unless $HOME an acceptable default for `cd` here, test for empty var?

dpezely (approved):
> Problem to solve: The current build environment is slow
[...]
> `sudo apt install --yes ruby`

There's the next culprit.

If we need to shave more time off this, assembling a .deb from its component files, all that's missing to call `dpkg-deb --build ...` directly is a slight modification to the generated subdirectory structure and creating `DEBIAN/control` text file.

(While my earlier effort for router unwittingly omitted `{erts, true}` from relx config, the .deb portions worked fine.  Just in case-- that's all.)

.github/scripts/make_debian.sh
jadeallenx: Since like 99% of GHA use this construction I'm reasonably confident it will exist in the context of GHA - if you're suggesting we ought generalize the script to make it work on non GHA environments?

jadeallenx:
> If we need to shave more time off this, assembling a .deb from its component files, all that's missing to call `dpkg-deb --build ...` directly is a slight modification to the generated subdirectory structure and creating `DEBIAN/control` text file.

I don't want to use the native debian packaging tools because they require a lot of overhead and control files we don't want or need to generate.  It's true that stuffing fpm as a gem takes a little time but we're still at least 3x faster than the old way we were doing.

.github/scripts/make_debian.sh
dpezely: No need to change, then.  Just double-checking as nuances of GHA are still fairly new to me.  Thanks!

Cargo.toml
```diff
@@ -1,11 +1,11 @@
 [workspace]
 members = [
-    "ingest",
+    "poc_ingest",
     "db_store",
     "file_store",
     "mobile_rewards",
     "poc_mobile_verifier",
-    "entropy",
+    "poc_entropy",
```
xandkar, line 8: yay :tada:

xandkar (approved):

PR #202, opened 2022-11-11
Remove Default lookup delay from mobile rewards server

bbalser:


maplant (approved):

mobile_rewards/src/server.rs
```diff
@@ -405,12 +403,12 @@ async fn reward_period(client: &mut follower::Client<Channel>) -> Result<Range<u
...
-    let after_utc = Utc.timestamp(last_reward_end_time, 0);
-    let now = Utc::now();
-    let stop_utc = now - Duration::minutes(DEFAULT_LOOKUP_DELAY);
-    let start_utc = after_utc.min(stop_utc);
-    (start_utc, stop_utc)
+pub fn get_time_range(
+    last_reward_end_time: i64,
+    now: DateTime<Utc>,
+) -> (DateTime<Utc>, DateTime<Utc>) {
+    let last_timestamp_utc = Utc.timestamp(last_reward_end_time, 0);
```
madninja, line 410: So wouldn’t a second accuracy for file time stamp maybe loose the millisecond accuracy file time?

mobile_rewards/src/server.rs
```diff
@@ -442,3 +440,28 @@ fn sign_txn(txn: &BlockchainTxnSubnetworkRewardsV1, keypair: &Keypair) -> Result
...
+mod tests {
+    use super::*;
+
+    #[test]
+    fn get_time_range_returns_24_hour_period() {
+        let last_timestamp = Utc.ymd(2022, 11, 1).and_hms(0, 0, 0).timestamp();
+        let now = Utc.ymd(2022, 11, 2).and_hms(0, 0, 0);
+        let (start, stop) = get_time_range(last_timestamp, now);
+
+        assert_eq!(start, Utc.ymd(2022, 11, 1).and_hms(0, 0, 0));
```
xandkar, line 454: If this date is meant to be the same as `last_timestamp`, I'd rather it be made explicit by naming the value and reusing, than duplicating.

mobile_rewards/src/server.rs
bbalser: This timestamp is not from a file, it is just a timestamp in seconds stored in the db

mobile_rewards/src/server.rs
bbalser: But last_timestamp is a timestamp not a `DateTime<Utc>>`

bbalser: Fixed 9c82ebf

xandkar (approved):

PR #204, opened 2022-11-13
Set start after when calling aws s3 list-objects-v2

bbalser:


file_store/src/file_store.rs
```diff
@@ -72,14 +72,13 @@ impl FileStore {
         let before = before.into();
         let after = after.into();
 
-        let stream = self
-            .client
+        self.client
             .list_objects_v2()
             .bucket(&self.bucket)
             .prefix(prefix)
+            .set_start_after(after.map(|dt| FileInfo::from((file_type.into(), dt)).into()))
```
madninja, line 79: no need to `file_type.into()` twice.. Once for the prefix and again here.. maybe have an intermediate file_type and do `.prefix(file_type.to_string())`

xandkar: Also please leave a comment explaining why this is done.

jeffgrunewald (approved):
+1 for the file type into but otherwise looks good

jeffgrunewald (approved):

andymck (approved):

PR #205, opened 2022-11-14
move reward unit calculations to iot verifier

jeffgrunewald:
We need to continue to calculate the total IoT reward shares as a factor (in part) of the number of witnesses to a PoC event but should be doing so in the IoT Verifier so as to be able to keep the logic in the (temporary) IoT receipt Injector light and to allow for easily calculating the full reward shares within the verifier when submitting to the reward indexer for Solana integration.

This change migrates the calculation of the reward unit value from the injector to the verifier. It adds the value as a new field on the valid beacon and witness structs as well as the accompanying proto definitions as part of this dependent PR https://github.com/helium/proto/pull/234

### DIFF
--- a/poc_iot_verifier/src/runner.rs
+++ b/poc_iot_verifier/src/runner.rs
@@ -25,6 +25,8 @@ use helium_proto::{
     Message,
 };
 use node_follower::follower_service::FollowerService;
+use rust_decimal::{Decimal, MathematicalOps};
+use rust_decimal_macros::dec;
 use sqlx::PgPool;
 use std::path::Path;
 use tokio::time;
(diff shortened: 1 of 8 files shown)


andymck (approved):
this generally lgtm.

I will note however I am not convinced performing rewards in the verifier is a good design choice.  I suspect it will make the task of horizontally scaling the verifier more difficult

poc_iot_verifier/src/runner.rs
```diff
@@ -423,3 +436,23 @@ impl Runner {
...
+            let unnormalized = Decimal::TWO - to_sub;
+            Ok(normalize_reward_unit(unnormalized))
+        } else {
+            Err(Error::InvalidExponent(exp.to_string()))
+        }
+    }
+}
+
+fn normalize_reward_unit(unnormalized: Decimal) -> Decimal {
+    std::cmp::min(HIP15_TX_REWARD_UNIT_CAP, unnormalized)
```
vihu, line 457: nit: Just move this to the above function directly, there's no need to have `normalized_reward_unit` function separately if we're just calling it from a single place.

jeffgrunewald: fixed [d613888](https://github.com/helium/oracles/pull/205/commits/d61388894263863c2d4eb13c992e96ca7330698f)

dpezely (approved):
Since I'm still relatively new to internals of rewards calculations and those bits of code were *moved*, I didn't scrutinize the calculations.

Mainly, I focused on the Rust code and library semantics-- both of which look good.

PR #206, opened 2022-11-14
purge db using streams

andymck:
Uses concurrent streams to purge the DB of stale beacon and witness reports

It also actually fixes the purging, as previously the purge did not take place as i misunderstood the lazy quality of map when originally implementing this.  Also fixes other bugs relating to the DB report ID

### DIFF
--- a/poc_iot_verifier/src/purger.rs
+++ b/poc_iot_verifier/src/purger.rs
@@ -196,7 +196,7 @@ impl Purger {
         )
         .await?;
         // delete the report from the DB
-        _ = Report::delete_report(&self.pool, &beacon_id).await;
+        Report::delete_report(&self.pool, &beacon_id).await?;
         Ok(())
     }
 
--- a/poc_iot_verifier/src/purger.rs
+++ b/poc_iot_verifier/src/purger.rs
@@ -16,7 +16,7 @@ use helium_proto::Message;
 use sqlx::PgPool;
 use tokio::time;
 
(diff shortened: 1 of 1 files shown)


poc_iot_verifier/src/purger.rs
```diff
@@ -198,22 +220,13 @@ impl Purger {
...
             "invalid_witness_report",
-            lora_invalid_witness_tx,
+            &lora_invalid_witness_tx,
             invalid_witness_report_proto
         )
         .await?;
 
         // delete the report from the DB
-        self.delete_db_report(public_key, packet_data.clone()).await;
+        _ = Report::delete_report(&self.pool, &witness_id).await;
```
vihu, line 229:
```suggestion
        Report::delete_report(&self.pool, &witness_id).await?;
```

poc_iot_verifier/src/purger.rs
```diff
@@ -166,27 +191,24 @@ impl Purger {
...
             "invalid_beacon",
-            lora_invalid_beacon_tx,
+            &lora_invalid_beacon_tx,
             invalid_beacon_proto
         )
         .await?;
         // delete the report from the DB
-        let public_key = beacon_report.report.pub_key.to_vec();
-        self.delete_db_report(public_key, packet_data.clone()).await;
+        _ = Report::delete_report(&self.pool, &beacon_id).await;
```
vihu, line 199:
```suggestion
        Report::delete_report(&self.pool, &beacon_id).await?;
```

jeffgrunewald (approved):

PR #209, opened 2022-11-16
Only create txn_service if do_submission is true

vihu:
# Problem

poc-injector server would inconsistently crash trying to connect to blockchain-node for no reason even when `do_submission` is set to false.

# Solution

Potentially bypass the error by not creating `TxnService` at all unless `do_submission` is set to true.


bbalser (approved):
Still not sure why we have to do this, but I don't have a better idea.

jeffgrunewald (approved):

PR #210, opened 2022-11-16
Remove delete from heartbeats and change to on conflict overwrite hot…

bbalser:
…spot_key

### DIFF
--- a/poc_mobile_verifier/src/heartbeats.rs
+++ b/poc_mobile_verifier/src/heartbeats.rs
@@ -209,18 +209,12 @@ impl Heartbeat {
             return Ok(false);
         }
 
-        sqlx::query("DELETE FROM heartbeats WHERE cbsd_id = $1 AND hotspot_key != $2 ")
-            .bind(&self.cbsd_id)
-            .bind(&self.hotspot_key)
-            .execute(&mut *exec)
-            .await?;
-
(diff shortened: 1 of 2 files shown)


maplant (approved):

poc_mobile_verifier/src/heartbeats.rs
```diff
@@ -209,18 +209,12 @@ impl Heartbeat {
             return Ok(false);
         }
 
-        sqlx::query("DELETE FROM heartbeats WHERE cbsd_id = $1 AND hotspot_key != $2 ")
```
jeffgrunewald, line 212: does this still do what the original change intended it to do? if we're not clearing the table from records of the last heartbeats that were registered to a hotspot aren't we just building up two sets of records in parallel here (one radio attached to two hotspots)? couldn't a game just keep switching them back and forth between hotspots throughout the day?

maplant: I believe it should because we change the constraint from being unique (hotspot_key, cbsd_id) to just cbsd_id (see the new migration)

bbalser: the insert statement now matches only on cbsd_id and updates hotspot_key on conflict

jeffgrunewald (approved):

PR #212, opened 2022-11-17
Explicitly flush receipt sender

vihu:
This seems like a correct thing to do after file_sink_write for a receipt txn is done.


bbalser:
why do you want to flush the file sink in this case? 
I don't think you should do it in the `try_each_concurrent`.

vihu:
> why do you want to flush the file sink in this case? I don't think you should do it in the `try_each_concurrent`.

Yeah, fair point. I had hoped this would fix an error I had noticed on the injector crashing. Will rethink.

PR #214, opened 2022-11-18
Load testing optimisations, fixes and tweaks

andymck:
Optimisations, fixes and tweaks originating from load testing


poc_iot_verifier/src/main.rs
```diff
@@ -57,27 +59,23 @@ impl Server {
...
+        let gateway_cache = GatewayCache::from_settings(settings).await?;
         let mut loader = loader::Loader::from_settings(settings).await?;
         let mut runner = runner::Runner::from_settings(settings).await?;
         let purger = purger::Purger::from_settings(settings).await?;
-        let mut density_scaler = DensityScaler::from_settings(settings.density_scaler.clone())?;
+        let mut density_scaler = DensityScaler::from_settings(settings.density_scaler.clone()).await?;
         tokio::try_join!(
-            runner.run(density_tx, &shutdown),
-            loader.run(&shutdown),
+            runner.run(&shutdown, &gateway_cache, density_scaler.hex_density_map()),
```
andymck, line 75: ideally we could pass the hex density map by reference ?

poc_iot_verifier/src/runner.rs
```diff
@@ -122,7 +118,8 @@ impl Runner {
                     match self.handle_db_tick(  shutdown.clone(),
                                                 lora_invalid_beacon_tx.clone(),
                                                 lora_invalid_witness_tx.clone(),
-                                                lora_valid_poc_tx.clone()).await {
+                                                lora_valid_poc_tx.clone(),
+                                                gateway_cache, hex_density_map.clone()).await {
```
andymck, line 122: if hex_density_map could be a reference this cloning could be avoided

poc_iot_verifier/src/runner.rs
```diff
@@ -154,129 +155,155 @@ impl Runner {
...
-                        self.handle_valid_poc(
-                            beacon,
-                            valid_beacon_report,
-                            verified_witnesses_result,
-                            &lora_valid_poc_tx,
-                            &lora_invalid_witness_tx,
-                        )
+        // verify POC beacon
+        let beacon_verify_result = poc
+            .verify_beacon(hex_density_map.clone(), gateway_cache, &self.pool)
```
andymck, line 230: if hex_density_map could be a reference this cloning could be avoided.  its especially bad here as we clone for *every* beacon we are verifying

poc_iot_verifier/src/runner.rs
```diff
@@ -154,129 +155,155 @@ impl Runner {
...
+            VerificationStatus::Valid => {
+                tracing::debug!(
+                    "valid beacon. entropy: {:?}, addr: {:?}",
+                    beacon.data,
+                    beacon.pub_key.to_string()
+                );
+                // beacon is valid, verify the POC witnesses
+                if let Some(beacon_info) = beacon_verify_result.gateway_info {
+                    let verified_witnesses_result = poc
+                        .verify_witnesses(&beacon_info, hex_density_map, gateway_cache)
```
andymck, line 242: if hex_density_map could be a reference this cloning could be avoided. its especially bad here as we clone for every witness we are verifying

poc_iot_verifier/src/main.rs
bbalser: I we have to pass it in as is and clone it throughout because it is an [Arc](https://doc.rust-lang.org/std/sync/struct.Arc.html) and has to track where the pointer to the underlying map.

poc_iot_verifier/src/poc.rs
```diff
@@ -221,23 +210,27 @@ impl Poc {
...
                         Error::not_found("invalid FollowerGatewayResp for witness")
                     })?;
-                    let scaling_factor = density_queries
-                        .query(gw_info.location.unwrap_or_default().to_string())
-                        .await?
+                    let scaling_factor = hex_density_map
+                        .get(&gw_info.location.unwrap_or_default().to_string())
+                        .await
                         .unwrap_or(Decimal::ONE);
+                    // let scaling_factor = Decimal::ONE;
```
jeffgrunewald, line 233:
```suggestion
```

poc_iot_verifier/src/runner.rs
```diff
@@ -104,6 +100,7 @@ impl Runner {
             lora_valid_poc_rx,
         )
         .deposits(Some(file_upload_tx.clone()))
+        .roll_time(ChronoDuration::minutes(10))
```
jeffgrunewald, line 103: is this intended to be a different number from the two invalid ones?

jeffgrunewald (commented):
minor nits; i wonder if in future we should move the gateway_cache to the node_follower crate instead and just have the entire api be optionally behind a cache you can enable or disable as needed.

looks like it also needs a rebase to grab the fixes for these unrelated clippy complaints

andymck:
> minor nits; i wonder if in future we should move the gateway_cache to the node_follower crate instead and just have the entire api be optionally behind a cache you can enable or disable as needed.

As discussed offline, the issue with moving the cache to the node follower is that in order to achieve concurrency from a follower client, it has to be cloned for each channel/concurrent stream.  Having the cache as part of the follower struct would result in having to clone it each time which would not be pretty.

poc_iot_verifier/migrations/3_poc_report.sql
```diff
@@ -22,3 +22,26 @@ create table poc_report (
...
+ON poc_report(status);
+
+CREATE INDEX idx_poc_report_created_at
+ON poc_report(created_at);
+
+CREATE INDEX idx_poc_report_attempts
+ON poc_report(attempts);
+
+CREATE INDEX idx_poc_report_report_type_status_created_at
+ON poc_report (report_type,status,created_at);
```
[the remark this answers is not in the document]
andymck: As we are still in dev, the DB is just recreated each time we change anything related

poc_iot_verifier/src/meta.rs
```diff
@@ -42,7 +42,7 @@ impl Meta {
 
     pub async fn last_timestamp<'c, E>(
         executor: E,
-        file_type: FileType,
+        file_type: &str,
```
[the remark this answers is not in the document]
andymck: I dont think its an issue.  The meta table is a key value store and shouldnt really have had any dep on file types anyway.    This branch adds a non filetype key called 'report' which necessitates this change

poc_iot_verifier/src/gateway_cache.rs
```diff
@@ -0,0 +1,52 @@
+use crate::{Error, Result, Settings};
+use helium_crypto::PublicKey;
+use node_follower::{
+    follower_service::FollowerService,
+    gateway_resp::{GatewayInfo, GatewayInfoResolver},
+};
+use retainer::*;
```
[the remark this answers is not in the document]
andymck: fixed here: 74a52995f5c218cde3d4d4fe538f810833ba5a64

poc_iot_verifier/src/loader.rs
```diff
@@ -129,76 +113,113 @@ impl Loader {
...
+                        gateway_cache,
+                        after,
+                        before,
+                    )
+                    .await;
+            })
+            .await;
+
+        stream::iter(&[FileType::LoraWitnessIngestReport])
+            .map(|file_type| (file_type))
```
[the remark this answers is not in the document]
andymck: This was a carry over from when multiple file types were handled in the same stream:iter and processed concurrently.  However I am now processing each file type serially, and have removed the stream iter wrapper here:  74a52995f5c218cde3d4d4fe538f810833ba5a64

poc_iot_verifier/src/runner.rs
andymck: this is intentional, the valid POCs will be uploaded to s3 more frequent as we have a known consumer of these ( iot injector currently )

jeffgrunewald (approved):

bbalser (approved):

PR #215, opened 2022-11-18
Remove usage of Custom error type

vihu:
- Introduce more specific errors when handling reports

### DIFF
--- a/file_store/src/error.rs
+++ b/file_store/src/error.rs
@@ -6,8 +6,6 @@ pub type Result<T = ()> = std::result::Result<T, Error>;
 pub enum Error {
     #[error("io error")]
     Io(#[from] std::io::Error),
-    #[error("custom error")]
-    Custom(String),
     #[error("encode error")]
     Encode(#[from] EncodeError),
     #[error("dencode error")]
@@ -24,6 +22,12 @@ pub enum Error {
     Config(#[from] config::ConfigError),
     #[error("mpsc channel error")]
     Channel,
+    #[error("unsupported datarate {0}")]
+    UnsupportedDataRate(String),
+    #[error("unsupported invalid_reason {0}")]
+    UnsupportedInvalidReason(String),
+    #[error("unsupported participant_side {0}")]
+    UnsupportedParticipantSide(String),
 }
 
 #[derive(Error, Debug)]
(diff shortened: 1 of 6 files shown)


file_store/src/error.rs
```diff
@@ -24,6 +22,12 @@ pub enum Error {
     Config(#[from] config::ConfigError),
     #[error("mpsc channel error")]
     Channel,
+    #[error("unsupported datarate {0}")]
+    UnsupportedDataRate(String),
+    #[error("unsupported invalid_reason {0}")]
```
madninja, line 27: may want to make this more explicit.. reason for what?

file_store/src/error.rs
```diff
@@ -24,6 +22,12 @@ pub enum Error {
...
+    UnsupportedInvalidReason(String),
+    #[error("unsupported participant_side {0}")]
+    UnsupportedParticipantSide(String),
```
madninja, line 30: May want to make this more explicit..

madninja (commented):
I think this is good, but I'd suggest 1. pulling them into DecodeError since you're decoding the proto types into internal types and 2. making the names more explicit to about what they are.. at least in the string description for them

vihu:
> I think this is good, but I'd suggest 1. pulling them into DecodeError since you're decoding the proto types into internal types and 2. making the names more explicit to about what they are.. at least in the string description for them

Check if [c688769](https://github.com/helium/oracles/pull/215/commits/c688769c8b664bac808b180d82c0e7adb3cdd09f) looks reasonable. Made it more verbose with specific type of proto and placed it in DecodeError.

madninja (commented):
Better! I'd actually have `DecodeError::unsupported_datarate` as a function too which takes `Into<String>` for both the parameters to shorten the calling sites

vihu:
> Better! I'd actually have `DecodeError::unsupported_datarate` as a function too which takes `Into<String>` for both the parameters to shorten the calling sites

Pushed [17f8f83](https://github.com/helium/oracles/pull/215/commits/17f8f835d46ab4e50d93f996417bb22b48acd534)

file_store/src/error.rs
vihu: [c688769](https://github.com/helium/oracles/pull/215/commits/c688769c8b664bac808b180d82c0e7adb3cdd09f)

file_store/src/error.rs
vihu: [c688769](https://github.com/helium/oracles/pull/215/commits/c688769c8b664bac808b180d82c0e7adb3cdd09f)

madninja (approved):

jeffgrunewald (approved):

PR #216, opened 2022-11-18
Change Hex Density to wrap a hashmap with and Arc<RwLock<>> so update…

bbalser:
… to map don't block reads for very long


density_scaler/src/server.rs
```diff
@@ -47,19 +46,6 @@ impl Server {
...
-                                   .get(&hex)
-                                   .map(|scale| scale.to_owned());
-                        tx.send(resp)
-                    },
-                    None => {
-                        tracing::warn!("query channel closed");
-                        return Ok(())
-                    }
-                },
                 // _ = trigger_timer.tick() => self.refresh_scaling_map().await?,
```
maplant, line 49: should this be uncommented now?

bbalser: This is from the base branch, Jeff is testing something this weekend

maplant (approved):

density_scaler/src/hex.rs
```diff
@@ -45,20 +45,29 @@ static HIP17_RES_CONFIG: [Option<HexResConfig>; 11] = [
...
+pub trait HexDensityMap: Clone {
+    async fn get(&self, hex: &str) -> Option<Decimal>;
+    async fn swap(&self, new_map: HashMap<String, Decimal>);
+}
+
+#[derive(Debug, Clone)]
+pub struct SharedHexDensityMap(Arc<RwLock<HashMap<String, Decimal>>>);
 
-impl ScalingMap {
+impl SharedHexDensityMap {
```
JayKickliter, line 57: perhaps outside of the scope of this PR, but I so new reason for this trait to exists. And having an `fn new()`  in a trait is almost never needed, if at all

density_scaler/src/hex.rs
```diff
@@ -45,20 +45,29 @@ static HIP17_RES_CONFIG: [Option<HexResConfig>; 11] = [
...
-#[derive(Debug, Serialize, Eq, PartialEq)]
-pub struct ScalingMap(HashMap<String, Decimal>);
+#[async_trait::async_trait]
+pub trait HexDensityMap: Clone {
+    async fn get(&self, hex: &str) -> Option<Decimal>;
+    async fn swap(&self, new_map: HashMap<String, Decimal>);
+}
+
+#[derive(Debug, Clone)]
+pub struct SharedHexDensityMap(Arc<RwLock<HashMap<String, Decimal>>>);
```
JayKickliter, line 55: If performance is a (the?) goal here, then using String for the map key when the actual key is an H3 cell is not going to help. Cells impl Hash directly. Cells also impl Ord and can be used in a BTreeMap (this is a case where I think a b-tree could make sense over a hash map)

bbalser: @jeffgrunewald

density_scaler/src/hex.rs
bbalser: As for the `HexDensityMap` trait, I was thinking having the trait would help with testing later, so we could swap in a different implementation if need be.

As for the `new` function, I like how the code outside this module does not know about the Arc/RowLock

Is there a reason not to have the `new` function?

JayKickliter: > Is there a reason not to have the new function?

Sorry, my fault. I misread and thought the `HexDensityMap` trait had a `new`.

density_scaler/src/hex.rs
```diff
@@ -45,20 +45,29 @@ static HIP17_RES_CONFIG: [Option<HexResConfig>; 11] = [
...
+impl SharedHexDensityMap {
     pub fn new() -> Self {
-        Self(HashMap::new())
+        Self(Arc::new(RwLock::new(HashMap::new())))
     }
+}
 
-    pub fn insert(&mut self, hex: &str, scale_factor: Decimal) -> Option<Decimal> {
-        self.0.insert(hex.to_string(), scale_factor)
+#[async_trait::async_trait]
```
jeffgrunewald, line 63: do you actually need to annotate the trait _and_ the impl of the trait with `async_trait`? this seems like something the compiler should be able to figure out from the annotation on the trait itself

density_scaler/src/hex.rs
```diff
@@ -45,20 +45,29 @@ static HIP17_RES_CONFIG: [Option<HexResConfig>; 11] = [
...
+impl HexDensityMap for SharedHexDensityMap {
+    async fn get(&self, hex: &str) -> Option<Decimal> {
+        self.0.read().await.get(hex).cloned()
```
jeffgrunewald, line 66: do you need `cloned()` here on the get or can you get away with returning the `*self.0.read().await.get(hex)`?

node_follower/src/gateway_resp.rs
```diff
@@ -4,6 +4,7 @@ use helium_crypto::PublicKey;
 use helium_proto::{
     services::follower::GatewayInfo as GatewayInfoProto, GatewayStakingMode, Region,
 };
+use retainer::*;
```
jeffgrunewald, line 7: this needed here or down in the gateway_cache?

poc_iot_verifier/src/loader.rs
```diff
@@ -18,22 +19,24 @@ use helium_proto::{
...
 };
-use node_follower::{follower_service::FollowerService, gateway_resp::GatewayInfoResolver};
 use sha2::{Digest, Sha256};
 use sqlx::PgPool;
 use std::time::Duration;
 use tokio::time;
 
 /// cadence for how often to look for new beacon and witness reports from s3 bucket
-const REPORTS_POLL_TIME: time::Duration = time::Duration::from_secs(60);
+const REPORTS_POLL_TIME: time::Duration = time::Duration::from_secs(60 * 5 + 10);
```
jeffgrunewald, line 28:
```suggestion
const REPORTS_POLL_TIME: time::Duration = time::Duration::from_secs((60 * 5) + 10);
```
not sure what the math should be here? should it be 310 or 900?

poc_iot_verifier/src/runner.rs
```diff
@@ -50,20 +49,17 @@ impl Runner {
             pool,
             settings: settings.clone(),
             // follower_cache,
-            density_queries: None,
```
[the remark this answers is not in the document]
bbalser: I wanted to the verifier to use an `impl HexDensityMap` and if I put that in a struct, it has to a trait object which enforces dynamic dispatch on the method calls, and rather than jump through those hoops, I just passed it in function arguments which doesn't require any of that.

density_scaler/src/hex.rs
bbalser: I had to unless I am missing something.

density_scaler/src/hex.rs
bbalser:
```
error[E0614]: type `std::option::Option<&Decimal>` cannot be dereferenced
  --> density_scaler/src/hex.rs:67:9
   |
67 |         *self.0.read().await.get(hex)
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   ```
   If I got what you wanted correct, compiler doesn't like it.

jeffgrunewald (commented):
should be able to clean up the commented out lines now that we know the query/process API is a bottleneck; otherwise looks good

density_scaler/src/hex.rs
JayKickliter: Yeah, `HashMap::get()` returns `Option<&V>`. `.cloned()` (moreso `.copied()`, but no real difference) is the right choice here.

density_scaler/src/hex.rs
JayKickliter: Rust doesn't currently support async trait functions. `async_trait` is a procedural macro that gets around this by turning all `async fn` into normal `fn` that return a future.

density_scaler/src/hex.rs
JayKickliter: > the location format over the wire that the verifier has

Got it. It's unfortunate that we don't keep indices `u64`, but given that constraint this makes sense.

jeffgrunewald (commented):
also needs to re-enable scheduling the refresh of the scaling map/tick interval handler

andymck (approved):

bbalser:
> also needs to re-enable scheduling the refresh of the scaling map/tick interval handler

Fixed afea922

jeffgrunewald (approved):

PR #217, opened 2022-11-19
Send incoming buf on write if there is no sink

vihu:
I _think_ this is more correct, we don't have to return an `Error` if there's no sink when we're trying to write an incoming buf. We just make a new one, send the incoming data and attach that to `FileSink`.

### DIFF
--- a/file_store/src/file_sink.rs
+++ b/file_store/src/file_sink.rs
@@ -4,7 +4,7 @@ use bytes::Bytes;
 use chrono::{DateTime, Duration, Utc};
 use futures::SinkExt;
 use std::{
-    io, mem,
+    mem,
     path::{Path, PathBuf},
 };
 use tokio::{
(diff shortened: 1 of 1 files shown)


file_store/src/file_sink.rs
```diff
@@ -363,24 +360,28 @@ impl FileSink {
...
-                self.active_sink = Some(self.new_sink().await?);
+                let mut new_sink = self.new_sink().await?;
+                assert!(new_sink.size == 0);
+                if buf_len < self.max_size {
+                    new_sink.transport.send(buf).await?;
+                    new_sink.size += buf_len;
+                    self.active_sink = Some(new_sink);
+                } else {
+                    // XXX: How can we directly deposit here since the incoming
+                    // buf is already too big?
```
madninja, line 379: This indicates this is not correct also.. The whole point was to deposit a sink and make a new one if the new buffer will overrun the sink. Then write to the sink. 

I still don't get how you have an empty sink at all. I'm missing the logic error in the original implementation and this patch adds incorrect behavior

file_store/src/file_sink.rs
```diff
@@ -363,24 +360,28 @@ impl FileSink {
                     let prev_path = active_sink.path.clone();
                     self.deposit_sink(&prev_path).await?;
                     self.active_sink = Some(self.new_sink().await?);
+                } else {
```
madninja, line 363: This is still not correct.. if active sink size + buf_len is more than max_size it's creating a new sink but then you're not writing the data to the new sink, leading to a lost buffer

madninja (changes requested):

file_store/src/file_sink.rs
vihu: 4f82d9a  addresses this particular concern I think. We basically flush the current active sink which is full. And then create a new one, send the incoming buf and attach it. Regardless, this full patch may still not be correct. More to dig through.

vihu:
Closing this, for now at least, since the error we've seen originates in iot-verifier's `loader.rs` module as well, which does not use file_sink at all. We have a suspicion that it may be coming from either `source_unordered` or `get_byte_stream` but haven't been able to narrow it down yet.

PR #218, opened 2022-11-21
Debug each error explicitly for txn submission

vihu:
- This changes the poc-iot-injector server's stream processing back to `for_each_concurrent` with explicitly tracing each error, primarily for debugging purposes. Hopefully we can catch the exact place at which the injector fatally fails and debug it further if we can. Otherwise, the hope is that at the minimum this change wouldn't bring down the whole system.

### DIFF
--- a/poc_iot_injector/src/server.rs
+++ b/poc_iot_injector/src/server.rs
@@ -6,7 +6,6 @@ use chrono::{DateTime, NaiveDateTime, TimeZone, Utc};
 use db_store::MetaValue;
 use file_store::{file_sink, file_sink_write, FileStore, FileType};
 use futures::stream::{self, StreamExt};
-use futures::TryStreamExt;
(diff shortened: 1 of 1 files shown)


vihu:
Can confirm that this change, at the very minimum, does not crash the whole system over when we see a `ConnectionReset` error. It's been running for 24 hrs at this point, we can wait another day to see what happens and then merge this.

jeffgrunewald (approved):
have we not encountered the point where the error was previously crashing the service yet?

vihu:
> have we not encountered the point where the error was previously crashing the service yet?

So far we haven't, I started it roughly 30 hrs ago. Admittedly there have been some iot-verifier crashes/fixes so the influx of valid-poc reports is not huge (yet). Regardless, injector has been running with this branch, so we can wait longer and see what happens.

PR #220, opened 2022-11-22
Stop using aws s3 built in paginator due to errors, handle pagination…

bbalser:
… with streams::unfold

### DIFF
--- a/file_store/src/file_store.rs
+++ b/file_store/src/file_store.rs
@@ -72,34 +72,53 @@ impl FileStore {
         let before = before.into();
         let after = after.into();
 
-        self.client
+        let request = self
+            .client
             .list_objects_v2()
             .bucket(&self.bucket)
             .prefix(file_type.to_string())
-            .set_start_after(after.map(|dt| FileInfo::from((file_type, dt)).into()))
-            .into_paginator()
-            .send()
-            .flat_map(move |entry| match entry {
-                Ok(output) => {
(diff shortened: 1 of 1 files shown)


file_store/src/file_store.rs
```diff
@@ -72,34 +72,53 @@ impl FileStore {
...
+                    .into_iter()
+                    .filter_map(|obj| {
+                        if FileInfo::matches(obj.key().unwrap_or_default()) {
+                            Some(FileInfo::try_from(&obj).unwrap())
+                        } else {
+                            None
+                        }
+                    })
+                    .filter(move |info| after.map_or(true, |v| info.timestamp > v))
+                    .filter(move |info| before.map_or(true, |v| info.timestamp <= v))
```
andymck, line 115: for this to be a 'true' before we should drop the equal sign, in the same way as the 'after' above

andymck (approved):
lgtm and has been running on mainnet verifier for past two days

file_store/src/file_store.rs
bbalser: So you would rather have `before` be exclusive?  I think we need one of them to be inclusive and given how `start_after` works in the s3api, I think it makes sense that before is inclusive.

bbalser: @madninja you okay with this?

madninja (approved):

PR #221, opened 2022-11-22
cleanup old cargo deb settings

jeffgrunewald:
since moving debian packaging to use `fpm` invoked on the build host from a script, these settings are no longer valid. this change cleans them up as well as migrating the settings.toml template files to all be within an oracle-level `pkg` directory since the build script simply searches for a file named `settings-template.toml` from within the build root without targeting a specific directory


jadeallenx (approved):

jadeallenx:
Meant to get around to this but thank you for taking care of it

bbalser (approved):

PR #223, opened 2022-11-23
Add dumps for radio reward share and reward manifest

maplant:


file_store/src/cli/dump.rs
```diff
@@ -132,6 +132,25 @@ impl Cmd {
                         "validity": heartbeat.validity,
                     }))?;
                 }
+                FileType::RadioRewardShare => {
+                    let reward = RadioRewardShare::decode(msg)?;
+                    print_json(&json!({
+                        "owner_key": PublicKey::try_from(reward.owner_key)?,
+                        "hotpost_key": PublicKey::try_from(reward.hotspot_key)?,
```
jeffgrunewald, line 139: do these need to be base64 encoded?

jeffgrunewald: nvm, i see that's the default display behavior

jeffgrunewald (approved):

bbalser (approved):

PR #226, opened 2022-11-29
Require multiple heartbeats to give reward to hotspot

maplant:
@meowshka currently this does not change any of the reports that we write to S3. Do we want to write anything that includes the heartbeat count?

### DIFF
--- a/poc_mobile_verifier/src/heartbeats.rs
+++ b/poc_mobile_verifier/src/heartbeats.rs
@@ -188,7 +188,7 @@ impl Heartbeat {
             r#"
             INSERT INTO heartbeats (hotspot_key, cbsd_id, reward_weight, timestamps)
             VALUES ($1, $2, $3, ARRAY[$4])
-            ON CONFLICT cbsd_id DO UPDATE SET
+            ON CONFLICT (cbsd_id) DO UPDATE SET
             timestamps = CASE WHEN heartbeats.hotspot_key = EXCLUDED.hotspot_key THEN
                              CASE WHEN date_trunc('hour', $4) != date_trunc('hour', heartbeats.timestamps[1]) THEN
                                  array_prepend($4, heartbeats.timestamps)
@@ -199,7 +199,7 @@ impl Heartbeat {
                             ARRAY[$4]
                          END,
             hotspot_key = EXCLUDED.hotspot_key,
-            reward_weight = EXCLUDED.reward_weight,
+            reward_weight = EXCLUDED.reward_weight
             RETURNING (xmax = 0) as inserted;
             "#,
         )


poc_mobile_verifier/src/heartbeats.rs
```diff
@@ -193,23 +188,30 @@ impl Heartbeat {
 
         Ok(sqlx::query_as::<_, HeartbeatSaveResult>(
             r#"
-            insert into heartbeats (hotspot_key, cbsd_id, reward_weight, timestamp)
-            values ($1, $2, $3, $4)
-            on conflict (cbsd_id) do update set
-            hotspot_key = EXCLUDED.hotspot_key, reward_weight = EXCLUDED.reward_weight, timestamp = EXCLUDED.timestamp
```
bbalser, line 199: I believe this will leave the hotspot under the old hotspot_key instead of moving it to the new hotspot_key in the case of a moved radio.  Also in that scenario, should we blank out all the timestamps?

Also, shouldn't we update the reward_weight still?

poc_mobile_verifier/src/heartbeats.rs
```diff
@@ -91,16 +97,19 @@ impl Extend<Heartbeat> for Heartbeats {
...
-                    timestamp: heartbeat.timestamp,
-                },
-            );
+                })
+                .or_default();
+            entry.reward_weight = heartbeat.reward_weight;
+            if entry.timestamps.is_empty()
+                || entry.timestamps[0].hour() != heartbeat.timestamp.hour()
+            {
+                entry.timestamps.insert(0, heartbeat.timestamp);
```
jeffgrunewald, line 111: should we be concerned with any performance implications of having to shift all the elements of the list right here vs. pushing newer elements onto the vector and then grabbing the last index instead of the first?

poc_mobile_verifier/src/heartbeats.rs
```diff
@@ -91,16 +97,19 @@ impl Extend<Heartbeat> for Heartbeats {
...
-                HeartbeatValue {
-                    reward_weight: heartbeat.reward_weight,
-                    timestamp: heartbeat.timestamp,
-                },
-            );
+                })
+                .or_default();
+            entry.reward_weight = heartbeat.reward_weight;
+            if entry.timestamps.is_empty()
+                || entry.timestamps[0].hour() != heartbeat.timestamp.hour()
```
jeffgrunewald, line 109: maybe overkill but why not take the difference in timestamps and insert if it's equal to a duration >= 1 hour? i'm picturing a radio that's been offline for 12 hours and registering its next heartbeat in the same "hour" as the last recorded one but i suppose if they're consistently online they should never go so long between successful heartbeats that would be a likely scenario

poc_mobile_verifier/src/heartbeats.rs
```diff
@@ -110,25 +119,9 @@ impl FromIterator<Heartbeat> for Heartbeats {
...
-                    HeartbeatValue {
-                        reward_weight: hb.reward_weight,
-                        timestamp: hb.timestamp,
-                    },
-                ))
-            })
-            .collect();
-        Self { heartbeats }
+        let mut heartbeats = Self::default();
+        heartbeats.extend(iter);
```
jeffgrunewald, line 123: clever; are there any appreciable performance implications to this over just `collect` ing into the collection since I have to imagine the _intention_ behind the Extend trait is to add to a collection that already has elements in it?

poc_mobile_verifier/src/owner_shares.rs
```diff
@@ -303,7 +303,12 @@ mod test {
             },
         ];
 
-        let heartbeats: Heartbeats = heartbeats.into_iter().collect();
+        let mut heartbeats: Heartbeats = heartbeats.into_iter().collect();
+
+        // Populate with the minimum required number of timestamps
+        for (_, hb) in heartbeats.heartbeats.iter_mut() {
+            hb.timestamps = std::iter::repeat(timestamp).take(12).collect();
```
jeffgrunewald, line 310: should these timestamps each be from different hours to capture the actual requirement? unless i'm misunderstanding this it's just going to get twelve of the _same_ timestamp

poc_mobile_verifier/src/heartbeats.rs
```diff
@@ -193,11 +186,21 @@ impl Heartbeat {
...
             r#"
-            insert into heartbeats (hotspot_key, cbsd_id, reward_weight, timestamp)
-            values ($1, $2, $3, $4)
-            on conflict (cbsd_id) do update set
-            hotspot_key = EXCLUDED.hotspot_key, reward_weight = EXCLUDED.reward_weight, timestamp = EXCLUDED.timestamp
-            returning (xmax = 0) as inserted;
+            INSERT INTO heartbeats (hotspot_key, cbsd_id, reward_weight, timestamps)
+            VALUES ($1, $2, $3, ARRAY[$4])
+            ON CONFLICT (cbsd_id) DO UPDATE SET
+            timestamps = CASE WHEN heartbeats.hotspot_key = EXCLUDED.hotspot_key THEN
```
jeffgrunewald, line 192: if heartbeats are against the radio and not the hotspot why are we gauging these based on whether or not the hotspot key is the same from one heartbeat to the next? shouldn't swapping a radio between different hotspots only affect the speedtest records?

jeffgrunewald (commented):
this all seems to work and the only _real_ question I have that might impact functionality is the one about matching current and incoming hotspot key in the sql update statement but i feel like the data structures are getting a little redundant and could be consolidated a bit. the way i remember it, the different `HeartbeatKey` and `HeartbeatValue` structs were in place to mirror the original requirement of treating the combination of the `cbsd_id` and `hotspot_key` as a composite primary key in the database, but now that we're scrapping that and _just_ making the `cbsd_id` the primary key on its own it seems like you could just collapse the different data structures into a single one that encompases the pubkey, radio id, reward shares and timestamps and implement the necessary `FromRow` sqlx trait on that single struct. My 2 cents, but I know we want to get this deployed soon.

poc_mobile_verifier/src/heartbeats.rs
maplant: This is only used when we do the generate cli so I don't think so

poc_mobile_verifier/src/heartbeats.rs
maplant: It was easier to do it this way, and again this is only used for the generate cli

poc_mobile_verifier/src/heartbeats.rs
maplant: There shouldn't be - this is what we would do in FromIterator anyway

poc_mobile_verifier/src/owner_shares.rs
maplant: Since the logic is entirely in the sql query I would argue no, that would give the impression that we are doing such checks on the client side, which we are not. The only thing that the client does is check that the number of timestamps is greater than 12

poc_mobile_verifier/src/heartbeats.rs
maplant: We have to reset the number of heartbeats seen because of gamification concerns. Otherwise, a person could swap a radio between two hotspots every thirty minutes and get double the rewards

maplant:
> this all seems to work and the only _real_ question I have that might impact functionality is the one about matching current and incoming hotspot key in the sql update statement but i feel like the data structures are getting a little redundant and could be consolidated a bit. the way i remember it, the different `HeartbeatKey` and `HeartbeatValue` structs were in place to mirror the original requirement of treating the combination of the `cbsd_id` and `hotspot_key` as a composite primary key in the database, but now that we're scrapping that and _just_ making the `cbsd_id` the primary key on its own it seems like you could just collapse the different data structures into a single one that encompases the pubkey, radio id, reward shares and timestamps and implement the necessary `FromRow` sqlx trait on that single struct. My 2 cents, but I know we want to get this deployed soon.

You are entirely correct that at this point this code no longer really reflects the database. My intention was fixing that in a separate PR in order to reduce the amount of changes for Monday as much as possible.

madninja:
> this all seems to work and the only _real_ question I have that might impact functionality is the one about matching current and incoming hotspot key in the sql update statement but i feel like the data structures are getting a little redundant and could be consolidated a bit. the way i remember it, the different `HeartbeatKey` and `HeartbeatValue` structs were in place to mirror the original requirement of treating the combination of the `cbsd_id` and `hotspot_key` as a composite primary key in the database, but now that we're scrapping that and _just_ making the `cbsd_id` the primary key on its own it seems like you could just collapse the different data structures into a single one that encompases the pubkey, radio id, reward shares and timestamps and implement the necessary `FromRow` sqlx trait on that single struct. My 2 cents, but I know we want to get this deployed soon.

The main thing to keep in mind is that we need to avoid rewarding two different hotspots with the same radio (people move radios around to get multiple rewards)

poc_mobile_verifier/src/heartbeats.rs
jeffgrunewald: yep, makes sense

poc_mobile_verifier/src/owner_shares.rs
jeffgrunewald: makes sense; i missed that we were _only_ tracking the hourly differences in the db

jeffgrunewald:
> > [...] seems like you could just collapse the different data structures into a single one that encompases the pubkey, radio id, reward shares and timestamps and implement the necessary `FromRow` sqlx trait on that single struct. My 2 cents, but I know we want to get this deployed soon.
> 
> You are entirely correct that at this point this code no longer really reflects the database. My intention was fixing that in a separate PR in order to reduce the amount of changes for Monday as much as possible.

Totally agree here; best to push this through as-is since it's been tested and works correctly and change it after Monday

jeffgrunewald (approved):

bbalser (approved):

PR #227, opened 2022-11-29
Remove subnetwork rewards

maplant:


poc_mobile_verifier/src/cli/server.rs
```diff
@@ -61,23 +61,10 @@ impl Cmd {
             radio_rewards_rx,
         )
         .deposits(Some(file_upload_tx.clone()))
-        .roll_time(Duration::minutes(15))
```
jeffgrunewald, line 64: does this default to a more frequent interval without this line?

jeffgrunewald (approved):

jeffgrunewald:
It looks like there's a fair amount of unused code (the `mobile` and the `reward_share` modules for example)

bbalser (approved):

PR #233, opened 2022-12-01
mobile index wired up to reward manifests

madninja:
* this also fixes up the TimestampDecode to avoid using the deprecated timestamp_*() calls on chrono::Utc


mobile_index/src/reward_index.rs
```diff
@@ -0,0 +1,37 @@
...
+
+    sqlx::query(
+        r#"
+        insert info reward_index (
+                address, 
+                rewards, 
+                last_reward
+            ) values ($1, $2, $3)
+            on conflict(address) do update set
+                rewards = rewards + EXCLUDED.rewards,
```
bbalser, line 26: this will need to be `rewards = reward_index.rewards + EXCLUDED.rewards,`

bbalser (approved):

jeffgrunewald (approved):

PR #238, opened 2022-12-05
Fuse the ordered and unordered sources

madninja:


bbalser (approved):
Is this a theory that the stream is being called after it is complete, or have you seen something that indicates this behavior?

jeffgrunewald (approved):
looking at the description of `fuse()` this feels like it should be the default behavior and you should call some `continue_to_poll()` function on a stream that may return `None` but could potentially return results after that

andymck:
testing this atm on mainnet

PR #239, opened 2022-12-05
Add support for constructing receipt txn from valid_poc file

vihu:
# Summary

This adds a new CLI `construct` for basically constructing poc receipts txns from an input lora_valid_poc file. This is largely for debugging since we have the `generate` CLI as well but that has dependency on setting up local minio for FileStore.

### DIFF
--- a/poc_iot_injector/src/cli/construct.rs
+++ b/poc_iot_injector/src/cli/construct.rs
@@ -36,7 +36,7 @@ async fn process_msg(
     msg: prost::bytes::BytesMut,
     shared_key_clone: Arc<Keypair>,
 ) -> anyhow::Result<TxnDetails> {
-    if let Ok(txn_details) = handle_report_msg(msg.clone(), shared_key_clone) {
+    if let Ok(txn_details) = handle_report_msg(msg, shared_key_clone) {
         tracing::debug!("txn_bin: {:?}", txn_details.txn.encode_to_vec());
         Ok(txn_details)
     } else {
--- a/poc_iot_injector/src/cli/construct.rs
+++ b/poc_iot_injector/src/cli/construct.rs
@@ -1,0 +1,45 @@
+use crate::{
+    receipt_txn::{handle_report_msg, TxnDetails},
+    Settings,
+};
+use anyhow::bail;
+use file_store::file_source;
+use futures::stream::StreamExt;
+use helium_crypto::Keypair;
+use helium_proto::Message;
+use std::{path::PathBuf, sync::Arc};
+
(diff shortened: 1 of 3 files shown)


poc_iot_injector/src/cli/construct.rs
```diff
@@ -0,0 +1,50 @@
...
+async fn process_msg(
+    msg: prost::bytes::BytesMut,
+    shared_key_clone: Arc<Keypair>,
+    before_ts: i64,
+) -> anyhow::Result<TxnDetails> {
+    if let Ok(txn_details) = handle_report_msg(msg.clone(), shared_key_clone, before_ts) {
+        tracing::debug!("txn_bin: {:?}", txn_details.txn.encode_to_vec());
+        Ok(txn_details)
+    } else {
+        bail!("unable to construct txn for msg {:?}", msg);
```
jeffgrunewald, line 48: how big are these messages likely to be if they're just a byte array when they're being output in the panic message here? we talking "swallow the terminal window" size?

poc_iot_injector/src/cli/construct.rs
```diff
@@ -0,0 +1,50 @@
...
+}
+
+async fn process_msg(
+    msg: prost::bytes::BytesMut,
+    shared_key_clone: Arc<Keypair>,
+    before_ts: i64,
+) -> anyhow::Result<TxnDetails> {
+    if let Ok(txn_details) = handle_report_msg(msg.clone(), shared_key_clone, before_ts) {
+        tracing::debug!("txn_bin: {:?}", txn_details.txn.encode_to_vec());
+        Ok(txn_details)
```
jeffgrunewald, line 46: if this is a cli command, what happens to this result? is this for debugging purposes in which case it would be helpful to output a more structured version of the txn than the `Vec<u8>` in the tracing message or should this write out the receipt(s) to a file where they can be further validated/worked with?

poc_iot_injector/src/cli/construct.rs
vihu: hmm, fair point. I'll remove the msg itself, it's not that useful in its raw form anyway.

poc_iot_injector/src/cli/construct.rs
vihu: Yeah, primarily for debugging. And actually outputting the Vec<u8> form is more useful for debugging cuz it's relatively easy to put that on a running node's remote console for inspection. Decoding it to an actual txn is not _really_ super helpful.

bbalser (approved):

jeffgrunewald (approved):

andymck (approved):

PR #242, opened 2022-12-06
Revert back to previous method of storing seen heartbeats

maplant:
This method does not require that the timestamps of heartbeats are seen in order.

### DIFF
--- a/poc_mobile_verifier/src/heartbeats.rs
+++ b/poc_mobile_verifier/src/heartbeats.rs
@@ -29,7 +29,15 @@ pub struct HeartbeatKey {
 #[derive(Default)]
 pub struct HeartbeatValue {
     pub reward_weight: Decimal,
-    pub timestamps: Vec<NaiveDateTime>,
+    pub hours_seen: [bool; 24],
+}
+
+impl HeartbeatValue {
+    fn heartbeat_count(&self) -> usize {
+        self.hours_seen
+            .iter()
+            .fold(0, |hbs, seen| *seen as usize + hbs)
+    }
 }
 
 pub struct HeartbeatReward {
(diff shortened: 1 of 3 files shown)


poc_mobile_verifier/src/heartbeats.rs
```diff
@@ -29,7 +29,15 @@ pub struct HeartbeatKey {
...
     pub reward_weight: Decimal,
-    pub timestamps: Vec<NaiveDateTime>,
+    pub hours_seen: [bool; 24],
+}
+
+impl HeartbeatValue {
+    fn heartbeat_count(&self) -> usize {
+        self.hours_seen
+            .iter()
+            .fold(0, |hbs, seen| *seen as usize + hbs)
```
jeffgrunewald, line 39: casting a `boolean` to a numeric type seems very anti-rust, even if you're doing it explicitly; just when i thought i had you figured out rust...

poc_mobile_verifier/src/heartbeats.rs
```diff
@@ -53,33 +61,35 @@ impl Heartbeats {
...
-                    },
-                    HeartbeatValue {
-                        reward_weight: hb.reward_weight,
-                        timestamps: hb.timestamps,
-                    },
-                )
-            })
-            .collect();
+        let heartbeats = sqlx::query_as::<_, HeartbeatRow>(
+            "SELECT (hotspot_key, cbsd_id, reward_weight, hours_seen) FROM heartbeats",
```
jeffgrunewald, line 68: this is the same as `*` right? this diff looks like it's changing more than i think it really is; effectively you're only actually accessing the `hours_seen` field on the result instead of the `timestamps` field previously?

bbalser (approved):

jeffgrunewald (approved):

poc_mobile_verifier/src/heartbeats.rs
maplant: It's quite rusty! https://doc.rust-lang.org/std/primitive.bool.html

PR #247, opened 2022-12-07
Anyhow mobile_rewards

maplant:
This is PR is part of the larger conversion of our oracles to use anyhow for application level errors. This converts `mobile_rewards` to use Anyhow


mobile_rewards/src/pending_txn.rs
```diff
@@ -84,24 +82,36 @@ pub struct PendingTxn {
     pub updated_at: Option<DateTime<Utc>>,
 }
 
+#[derive(thiserror::Error, Debug)]
+pub enum PendingTxnError {
+    #[error("{0} not found")]
```
jeffgrunewald, line 87:
```suggestion
    #[error("{not found: {0}")]
```

mobile_rewards/src/pending_txn.rs
```diff
@@ -134,11 +144,15 @@ impl PendingTxn {
         .bind(&self.hash)
         .bind(&self.txn_bin)
         .fetch_one(executor)
-        .map_err(Error::from)
-        .await
+        .await?)
```
jeffgrunewald, line 147: functionally, what conversion is this performing for us that we "unwrap" the result and then re-wrap it in an `Ok()` on success? don't we get the same result by just returning the outcome of the method chain ending in `.await` without the `?` and still get the error thrown or else an `Ok(result)`?

mobile_rewards/src/pending_txn.rs
```diff
@@ -148,17 +162,16 @@ impl PendingTxn {
             .bind(hash)
             .bind(timestamp.into())
             .execute(executor)
-            .map_ok(|_| ())
-            .map_err(Error::from)
-            .await
+            .await?;
+        Ok(())
```
jeffgrunewald, line 166: is there a reason we're wrapping the previous function/method chain in an `Ok()` and here we're just executing the chain and then returning an empty `Ok(())`? neither sql statement is a "get" request so i could see an argument both for and against returning the record being "upserted" in either case so these feel like they could be aligned

jeffgrunewald (approved):
i'd still like to understand better at what level we should be returning a "local" error struct type that derives `thiserror::Error` and where we should be returning `anyhow::Error`; here we're returning an anyhow error in the main.rs and the server.rs and i don't yet see the pattern for why those two modules return `anyhow::Error` and everything else returns `thiserror::Error`

mobile_rewards/src/token_type.rs
```diff
@@ -16,12 +15,17 @@ impl Display for BlockchainTokenTypeV1 {
...
 }
 
+#[derive(thiserror::Error, Debug)]
+#[error("{0} is not a blockchain token type")]
+pub struct NoSuchBlockchainTokenType(i32);
+
 impl TryFrom<i32> for BlockchainTokenTypeV1 {
-    type Error = Error;
-    fn try_from(value: i32) -> Result<Self> {
+    type Error = NoSuchBlockchainTokenType;
```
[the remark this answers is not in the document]
maplant: This isn't an alias - it's an associated type

maplant:
> i'd still like to understand better at what level we should be returning a "local" error struct type that derives `thiserror::Error` and where we should be returning `anyhow::Error`; here we're returning an anyhow error in the main.rs and the server.rs and i don't yet see the pattern for why those two modules return `anyhow::Error` and everything else returns `thiserror::Error`

`main.rs` and `server.rs` get anyhow because they  are _applications_ everything else gets local error structs since they are _APIs_

mobile_rewards/src/pending_txn.rs
maplant: While I think you can consider nil a valid return value from an expression and therefore it should go in the Ok, I think that the returning of nil usually signals state effects and therefore should be written sequentially

mobile_rewards/src/pending_txn.rs
maplant: It's a different error - this is a convenient way of converting an Error type into the desired one. 
The reason I do this as opposed to mapping is that try blocks will mirror something like this, as try blocks have an implicit Ok mapping their return values.

jeffgrunewald:
> `main.rs` and `server.rs` get anyhow because they are _applications_ everything else gets local error structs since they are _APIs_

i get why `main.rs` is an application if that's got the `main` or `tokio::main` annotation but how do we determine what else is an application vs. an api, for instance what makes `server.rs` an application and not an api called by main? in the case of the iot verifier would that make the `loader.rs`, `runner.rs`, and `purger.rs` all applications b/c they're long-running threaded processes?

maplant:
> > `main.rs` and `server.rs` get anyhow because they are _applications_ everything else gets local error structs since they are _APIs_
> 
> i get why `main.rs` is an application if that's got the `main` or `tokio::main` annotation but how do we determine what else is an application vs. an api, for instance what makes `server.rs` an application and not an api called by main? in the case of the iot verifier would that make the `loader.rs`, `runner.rs`, and `purger.rs` all applications b/c they're long-running threaded processes?

Indeed. Specifically, the `run` functions return `anyhow` results, while the constructions return `New{Loader, Runner, Purger}Error`

JayKickliter:
> +82 −487

As much as I love deleted code, I think we need to summarize our motivations (emphasize the 'why's) a bit better when opening PRs with this many changes.

JayKickliter (changes requested):
> +82 −487

As much as I **love deleted code**, I think we need to summarize our motivations (emphasize the 'why's) when opening PRs with this many changes. I don't comments/discord discussions are substitutes either.

mobile_rewards/src/token_type.rs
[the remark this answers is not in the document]
jeffgrunewald: I guess what I mean is more whether or not it’s correct to reference the associated type in the scope of the impl as `Self::Error` rather than explicitly by name? Or else why make it an associates type at all? Can you not use the error in that scope otherwise?

bbalser (approved):

mobile_rewards/src/token_type.rs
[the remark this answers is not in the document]
maplant: Neither is more correct, as long as that type `== Self::Error`. I tend to use `Self::Error` to match the trait definition

JayKickliter (approved):

PR #251, opened 2022-12-07
refactor of poc module and unit tests for each verification

andymck:
Refactor of the `poc` module to break up long winded functions and to break out the beacon and witness verifications to independently testable functions.  It then adds unit tests for each such verification.

I was planning on also adding tests for the poc struct and confirm the input and outputs of an overall poc but i need more planning for those and how to work around the lack of a follower etc.  That can come later as a sep PR, this is already long enough as it is in any case.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1877,7 +1877,7 @@ dependencies = [
 [[package]]
 name = "helium-proto"
 version = "0.1.0"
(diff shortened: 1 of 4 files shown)


bbalser (approved):
❤️

jeffgrunewald (approved):

PR #252, opened 2022-12-07
Anyhow poc_iot_injector

maplant:


poc_iot_injector/src/server.rs
```diff
@@ -208,7 +216,7 @@ async fn handle_txn_submission(
                     "txn submission failed!, hash: {:?}",
                     txn_details.hash_b64_url
                 );
-                Err(Error::TxnSubmission(txn_details.hash_b64_url))
+                Err(HandleTxnSubmissionError(txn_details.hash_b64_url))
```
vihu, line 219: I had a patch locally where I wasn't encapsulating this into its own error since it reduces visibility, let's just propogate the actual Error originating from the node follower on txn submission and just return `Err(Error:from(err))` instead?

poc_iot_injector/src/server.rs
```diff
@@ -56,7 +66,7 @@ impl Server {
         Ok(result)
     }
 
-    pub async fn run(&mut self, shutdown: &triggered::Listener) -> Result {
+    pub async fn run(&mut self, shutdown: &triggered::Listener) -> anyhow::Result<()> {
```
vihu, line 69: nit: Hoist anyhow at the top

poc_iot_verifier/src/loader.rs
```diff
@@ -64,7 +74,7 @@ impl Loader {
         &mut self,
         shutdown: &triggered::Listener,
         gateway_cache: &GatewayCache,
-    ) -> Result {
+    ) -> anyhow::Result<()> {
```
andymck, line 77: why not import anyhow result and avoid the need for all the explicit type annotations here and in other modules ?

poc_iot_verifier/src/poc.rs
```diff
@@ -1,13 +1,15 @@
 use crate::{
-    entropy::ENTROPY_LIFESPAN, gateway_cache::GatewayCache, last_beacon::LastBeacon, Error, Result,
+    entropy::ENTROPY_LIFESPAN,
```
andymck, line 2: i have heavily refactored this module here: https://github.com/helium/oracles/pull/251

I'll sort out the conflicts and carry across the anyhow changes after #252 gets merged

andymck (approved):

poc_iot_verifier/src/loader.rs
jeffgrunewald: Since the `std::result::Result` is the default when we’re not importing our own thiserror module based result type and we’re using that one more often throughout the app’s other modules, I appreciate the explicit call out

poc_iot_injector/src/server.rs
maplant: I'm not sure what you mean, this basically does the same thing as before

vihu: It's fine. I can patch it in a smaller PR. Basically I would like to remove the `HandleTxnSubmissionError` entirely and just propagate the error from node_follower up.

vihu (approved):

PR #254, opened 2022-12-08
file store and file ggsink improvements

bbalser:
file_store: Add file `stream_files` which will return a stream FileData which contains the stream of bytes in the file.
file_sink: Added optional transaction semantics

Everything should be backwards compatible.

### DIFF
--- a/file_store/src/file_sink.rs
+++ b/file_store/src/file_sink.rs
@@ -435,8 +435,6 @@ impl FileSink {
                     if self.auto_commit {
                         self.commit().await?;
                     }
-                    // let prev_path = active_sink.path.clone();
-                    // self.deposit_sink(&prev_path).await?;
                     self.new_sink().await?;
                 }
             }
--- a/file_store/src/file_sink.rs
+++ b/file_store/src/file_sink.rs
@@ -40,8 +40,8 @@ fn transport_sink(transport: &mut Transport) -> &mut Sink {
 #[derive(Debug)]
 pub enum Message {
     Data(oneshot::Sender<Result>, Vec<u8>),
-    Flush(oneshot::Sender<Result>),
-    FetchManifest(oneshot::Sender<Result<Vec<String>>>),
+    Commit(oneshot::Sender<Result<Vec<String>>>),
+    Rollback(oneshot::Sender<Result<Vec<String>>>),
 }
 
 pub type MessageSender = mpsc::Sender<Message>;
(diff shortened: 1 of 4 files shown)


file_store/src/file_store.rs
```diff
@@ -198,6 +204,25 @@ impl FileStore {
             })
             .boxed()
     }
+
+    pub fn stream_files<A, B, F>(&self, file_type: F, after: A, before: B) -> Stream<FileData>
```
andymck, line 208: I note the `listing` of files is handled internally here.  I think it might be better to allow the caller to list the files and then stream_files take this as an input.  It allows simple things like the caller to choose to log the number of files found, which is currently what the iot verifier does ( and is helpful to get an idea of whats going on ).  

Obviously you could log it here too but then that enforces the same behaviour for all consumers of the file store

file_store/src/file_store.rs
```diff
@@ -198,6 +204,25 @@ impl FileStore {
...
+    where
+        F: Into<FileType> + Copy,
+        A: Into<Option<DateTime<Utc>>> + Copy,
+        B: Into<Option<DateTime<Utc>>> + Copy,
+    {
+        let client = self.client.clone();
+        let bucket = self.bucket.clone();
+
+        self.list(file_type, after, before)
```
andymck, line 217: is this thing going to return an error and the file on which the error occurred ?

file_store/src/file_store.rs
bbalser: I added `stream_file` to make it up to caller how to loop over files

maplant (approved):

file_store/src/file_sink.rs
```diff
@@ -40,8 +40,8 @@ fn transport_sink(transport: &mut Transport) -> &mut Sink {
 #[derive(Debug)]
 pub enum Message {
     Data(oneshot::Sender<Result>, Vec<u8>),
-    Flush(oneshot::Sender<Result>),
-    FetchManifest(oneshot::Sender<Result<Vec<String>>>),
+    Commit(oneshot::Sender<Result<Vec<String>>>),
```
jeffgrunewald, line 43: Maybe a `Vec<String>` wants to be it’s own type or at least a type Alia’s here for clarity?

jeffgrunewald (approved):

file_store/src/file_sink.rs
bbalser: Address 7301c64

andymck (approved):

PR #256, opened 2022-12-11
PublicKeyBinary

madninja:
This switches internal structs to use PublicKeyBinary as mostly a drop in replacement for PublicKey. 

It also optimizes a number of clones away for report and ingest_id since that's just not needed. 

Requires https://github.com/helium/helium-crypto-rs/pull/41 to land, and versioned, and Cargo.toml to be updated here


andymck (approved):

bbalser (approved):

andymck:
included now with #245

PR #258, opened 2022-12-13
Switch txn submission to use JoinSet

vihu:
- Simplify submission error handling
- Fix function calls
- Also removes the file_sink_write when `do_submission` is set to `true`.

### DIFF
--- a/poc_iot_injector/src/server.rs
+++ b/poc_iot_injector/src/server.rs
@@ -175,6 +175,11 @@ async fn submit_txns(
         }
     }
 
+    // Make sure the tasks are finished to completion even when we run out of stream items
+    while !set.is_empty() {
+        set.join_next().await;
+    }
+
     Ok(())
 }
 
--- a/poc_iot_injector/src/server.rs
+++ b/poc_iot_injector/src/server.rs
@@ -1,6 +1,6 @@
 use crate::{
     receipt_txn::{handle_report_msg, TxnDetails},
-    Settings, LOADER_WORKERS, STORE_WORKERS,
+    Settings, LOADER_WORKERS,
 };
 use chrono::{DateTime, NaiveDateTime, TimeZone, Utc};
 use db_store::MetaValue;
@@ -8,10 +8,13 @@ use file_store::{file_sink, file_sink_write, FileStore, FileType};
 use futures::stream::{self, StreamExt};
 use helium_crypto::Keypair;
 use node_follower::txn_service::TransactionService;
+use prost::bytes::BytesMut;
 use sqlx::{Pool, Postgres};
 use std::sync::Arc;
 use std::time::Duration as StdDuration;
-use tokio::time;
+use tokio::{task::JoinSet, time};
+
+const MAX_CONCURRENT_SUBMISSIONS: usize = 16;
 
 pub struct Server {
     pool: Pool<Postgres>,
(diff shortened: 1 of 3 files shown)


poc_iot_injector/src/server.rs
```diff
@@ -139,85 +142,84 @@ async fn submit_txns(
...
+                if set.len() > MAX_CONCURRENT_SUBMISSIONS {
+                    tracing::info!("Processing {MAX_CONCURRENT_SUBMISSIONS} submissions");
+                    set.join_next().await;
                 }
             }
-        })
-        .await;
+        }
+    }
+
```
maplant, line 177: Best to add the following here: 

```rust
while !set.is_empty() { 
     set.join_next().await;
 }
 ```
 
 `JoinSet`  aborts all of the threads it contains on drop

Comment (vihu) [reply poc_iot_injector/src/server.rs:177]: Update [5069b47](https://github.com/helium/oracles/pull/258/commits/5069b47707ccfcbe6fec0de12a4714b312e9a738)

**Review by maplant — APPROVED** [review]

File: poc_iot_injector/src/server.rs [hunk 2]
```diff
@@ -139,85 +142,89 @@ async fn submit_txns(
...
+                        do_submission,
+                        &receipt_sender_clone,
+                        &mut shared_txn_service,
+                    )
+                    .await;
+                });
+
+                if set.len() > MAX_CONCURRENT_SUBMISSIONS {
+                    tracing::info!("Processing {MAX_CONCURRENT_SUBMISSIONS} submissions");
+                    set.join_next().await;
```
jeffgrunewald, line 172: so will this effectively hold within the body of this `while let` loop until the stream has exhausted, and keep processing (`join` ing) the next message every time the queue exceeds the maximum length until it drops back down again? then when the stream _is_ exhausted it'll iterate through the `while` loop below until all of the outstanding tasks have been completed?

poc_iot_injector/src/server.rs
```diff
@@ -139,85 +142,89 @@ async fn submit_txns(
...
+    do_submission: bool,
+    receipt_sender: &file_sink::MessageSender,
+    txn_service: &mut Option<TransactionService>,
+) -> anyhow::Result<()> {
+    match handle_report_msg(buf, shared_key) {
+        Ok(txn_details) => {
+            tracing::debug!("txn_details: {:?}", txn_details);
+            if do_submission {
+                tracing::info!("submitting txn: {:?}", txn_details.hash_b64_url);
+                match handle_txn_submission(txn_details.clone(), txn_service).await {
```
jeffgrunewald, line 198: do we not want to continue to write the receipts to the output bucket (for testing or auditing at least) when we're submitting them to the chain?

poc_iot_injector/src/server.rs
vihu: Yeah, that's the idea at least.

poc_iot_injector/src/server.rs
vihu: Two reasons:
1. I wasn't really sure where it was bottle-necking so I removed the file_sink_write.
2. Its kinda pointless anyway since the receipts would already be on chain and whatever audit, testing, observability is required can be done from chain. Note that we can only be best-effort here as far as submission is concerned.

That said, if you think its still valuable to put the receipts txns on chain when submission is happening, I can add that back in.

cc: @Vagabond

jeffgrunewald: i wonder if it would be helpful for correlating submitted txns vs. minted txns but we could always infer the expected submissions from the other upstream data in the pipeline that fed into those receipts.

jeffgrunewald (approved):

PR #259, opened 2022-12-13
move hex scaling factor to validation check

jeffgrunewald:
This makes hex scaling factor another validation check, requiring for the density scaler to have found a valid scaling factor decimal for the asserted gateway's hex location. Depends on https://github.com/helium/proto/pull/241 being merged to clean up the branch reference in the root cargo file.

Also performs some minor error handling cleanup related to the previous channel-based implementation of the density scaler.

### DIFF
--- a/poc_iot_verifier/src/poc.rs
+++ b/poc_iot_verifier/src/poc.rs
@@ -58,6 +58,7 @@ pub struct VerifyWitnessResult {
     result: VerificationStatus,
     invalid_reason: Option<InvalidReason>,
     pub gateway_info: Option<GatewayInfo>,
+    hex_scale: Option<Decimal>,
 }
 
 pub struct VerifyWitnessesResult {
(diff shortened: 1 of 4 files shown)


andymck (commented):
I had been looking at adding an explicit validation for the hex scale, ie verifying if we do have a value and that is is between 0 and 1.  But maybe that is overkill ?   That as it is can work too

poc_iot_verifier/src/poc.rs
```diff
@@ -307,23 +326,18 @@ impl Poc {
...
-                    .ok_or(VerificationError::NotFound("invalid beacon_location"))?,
-            )
-            .await
-            .unwrap_or(Decimal::ONE);
         Ok(LoraValidWitnessReport {
             received_timestamp: witness_report.received_timestamp,
             location: gw_info.location,
-            hex_scale: scaling_factor,
+            // witnesses should no longer validate if scaling factor not found, so this should be changed
+            // to a `ValidationError(NoScalingFactor)` or similar post-anyhow migration
```
jeffgrunewald, line 337: can probably delete this comment now to avoid confusion

PR #260, opened 2022-12-13
Allow configuring multiple followers

vihu:
# Summary

This allows configuring multiple follower nodes effectively choosing a client at random to submit txns and querying responses.

# TODO:
- [ ] Testing
- [ ] Change injector to actually use multiple followers and submit txns to them

# Questions

- I'm not 100% certain whether this is actually the right approach but its one way at least. We can discuss internally if we need to do things differently.


node_follower/src/follower_service.rs
```diff
@@ -57,16 +69,19 @@ impl FollowerService {
             txn_hash: txn_hash.to_vec(),
             txn_types: txn_types.iter().map(|e| e.to_string()).collect(),
         };
-        let res = self.client.txn_stream(req).await?.into_inner();
+        let mut client = self.random_client()?;
+        let res = client.txn_stream(req).await?.into_inner();
```
jeffgrunewald, line 73: what happens if the client disconnects and goes to reconnect, gets a different client, and this one is either ahead of or behind where the previous client was in the txn stream? the other APIs should all be safe to run against more or less any one of a set of nodes but this seems like it could cause inconsistency if any of the nodes are not perfectly in sync as to where they are in following the chain

jeffgrunewald (commented):
what's the primary reason for this change? if it's to have redundancy/horizontally scale how many nodes are available to receive receipt transactions for distributing the load to the helium chain across multiple nodes could you avoid the concern i raised in the follower service module by _only_ implementing this change for the `transcation_service`? that only provides two unary RPCs so it should be safe for the injector to hop from one node to the next on disconnects whereas for the follower service that seems like it could get trickier

node_follower/src/follower_service.rs
Vagabond: as long as the nodes you're submitting are in the current epoch it doesn't matter, and even if they're behind an epoch or two, enough of the consensus group should overlap that they can still get the txn to a group member.

Vagabond: And these poc receipt transactions are largely unrelated to each other, the ordering is not important.

jeffgrunewald: submitting is still a unary request; we don't (currently) stream transactions to the txn_manager for submission. i'm talking about the txn feed coming _out of_ the consensus group that this `txn_stream` rpc is providing. this is how we notify the oracle in question that a txn (typically rewards txn) has indeed been accepted by the CG and we update the database to mark the txn as settled. we checkpoint the blocks the node has seen so when it disconnects it can reconnect to the same point in the chain and not miss any txns, so this specific rpc seems like it would need to be more precise than just within the current epoch

node_follower/src/txn_service.rs
```diff
@@ -26,13 +35,16 @@ impl TransactionService {
...
+
+        let mut client = self.random_client()?;
+        let res = client.submit(req).await?.into_inner();
         Ok(res)
     }
 
     pub async fn query(&mut self, key: &[u8]) -> Result<TxnQueryRespV1> {
         let req = TxnQueryReqV1 { key: key.to_vec() };
-        let res = self.client.query(req).await?.into_inner();
+        let mut client = self.random_client()?;
```
jeffgrunewald, line 46: since you're querying the txn manager you submitted the txn to, i don't think randomly selecting a client to query is likely to get you back the same client that has the txn you're looking for in its queue

vihu:
Closing in favor of https://github.com/helium/oracles/pull/278

PR #264, opened 2022-12-15
Change poc_iot_verifier purger to delete beacons/witnesses in a db transaction

bbalser:


poc_iot_verifier/src/poc_report.rs
```diff
@@ -130,10 +130,10 @@ impl Report {
         Ok(())
     }
 
-    pub async fn delete_report<'c, 'q, E>(executor: E, id: &'q Vec<u8>) -> Result<(), ReportError>
-    where
-        E: sqlx::Executor<'c, Database = sqlx::Postgres> + Clone,
-    {
+    pub async fn delete_report(
+        executor: impl sqlx::PgExecutor<'_>,
```
andymck, line 134: doesnt this break the existing patten applied post the anyhow changes?

andymck (approved):
lgtm, one comment re the poc report function definition pattern

poc_iot_verifier/src/poc_report.rs
bbalser: so I could just remove clone from the `E` type so it would become
```
        E: sqlx::Executor<'c, Database = sqlx::Postgres>,
```
but I just thought that `executor: impl sqlx::PgExecutor<'_>` read better.  Do you know the reason for preferring the previous pattern.

jeffgrunewald (approved):

PR #267, opened 2022-12-15
IoT Packet Verifier

maplant:


iot_packet_verifier/src/balances.rs
```diff
@@ -0,0 +1,130 @@
...
+        let mut balances = HashMap::new();
+
+        while let Some(Burn {
+            payer,
+            amount: burn_amount,
+            ..
+        }) = burns.next().await.transpose()?
+        {
+            // Look up the current balance of the payer
+            let payer = PublicKey::try_from(payer).unwrap();
```
andymck, line 50: I am closing my eyes to multiple unwraps in here...see no evil
Also noting on the PublicKey vs PublicKeyBinary again

iot_packet_verifier/src/daemon.rs
```diff
@@ -0,0 +1,215 @@
...
+                    last_verified = self.handle_tick(last_verified.clone()).await?;
+                }
+            }
+        }
+
+        Ok(())
+    }
+
+    async fn handle_tick(&mut self, mut last_verified: NaiveDateTime) -> Result<NaiveDateTime> {
+        let reports =
```
andymck, line 53: I note the daemon will process all reports from the last_verified timestamp up until the current time.  Is there a need to consider a sliding window here as per the POC verifier loader or even a simpler max age for last timestamp ?  For example what if the verifier is down for an extended period of time, do we want to process all queued packets 
 from when it was last up ?  

It may well be fine to just process all queued packets but if network usage takes off it could be a lot of data to get through.   That could in turn result in more pressure on the burner side as more accounts build up to the threshold at which solana accounts are required to be updated.

On the POC side we have taken a lossy approach, ie we are allowed to drop reports in order to ensure the verifier remains current.  But maybe the volume on the packet side is not of a concern, at least currently.

iot_packet_verifier/src/daemon.rs
```diff
@@ -0,0 +1,215 @@
...
+
+        let mut org_cache = HashMap::<u64, PublicKey>::new();
+
+        while let Some(report) = reports.next().await {
+            let report_timestamp =
+                NaiveDateTime::from_timestamp_millis(report.gateway_timestamp_ms as i64).unwrap();
+
+            // Since the report timestamp will always be behind the timestamp for the while,
+            // the first few reports we see upon restarting will have already been processed.
+            if report_timestamp <= last_verified {
```
andymck, line 66: why would this be the case, ie that we will reload reports that were previously verified ?

iot_packet_verifier/src/balances.rs
```diff
@@ -0,0 +1,130 @@
...
+use solana_sdk::program_pack::Pack;
+use solana_sdk::pubkey::Pubkey;
+use sqlx::{Pool, Postgres};
+use std::collections::HashMap;
+use std::sync::Arc;
+use tokio::sync::Mutex;
+
+pub struct Balances {
+    pub provider: Arc<RpcClient>,
+    pub balances: Arc<Mutex<HashMap<PublicKey, Balance>>>,
```
andymck, line 16: Do we need to use full public key here as the hashmap key ? PublicKeyBinary much more efficient

andymck (commented):
Some initial comments, still digesting...

iot_packet_verifier/src/daemon.rs
maplant: I'm not sure what the best approach is. At the very least, the value is configurable by modifying the DB so if we crash we should be able to set which packet we start at

iot_packet_verifier/src/balances.rs
maplant: Sure, PublicKeyBinary works

iot_packet_verifier/src/daemon.rs
maplant: Because we're storing the timestamp of the packet and not the file. Files should always have a later timestamp than the packets they contain

file_store/src/file_info.rs
```diff
@@ -106,6 +106,9 @@ pub const VALIDATED_HEARTBEAT: &str = "validated_heartbeat";
 pub const SIGNED_POC_RECEIPT_TXN: &str = "signed_poc_receipt_txn";
 pub const RADIO_REWARD_SHARE: &str = "radio_reward_share";
 pub const REWARD_MANIFEST: &str = "reward_manifest";
+pub const IOT_PACKET_REPORT: &str = "packet_report";
+pub const VALID_PACKET: &str = "valid_packet";
```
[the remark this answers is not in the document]
maplant: Yup, I certainly can

iot_packet_verifier/Cargo.toml
```diff
@@ -0,0 +1,42 @@
+ [package]
+name = "iot-packet-verifier"
+version = "0.1.0"
+edition = "2021"
+description = "Packet verification for IOT"
+license = "Apache-2.0"
```
jeffgrunewald, line 6:
```suggestion
name = "iot-packet-verifier"
version = "0.1.0"
description = "Packet verification for IOT"
edition.workspace = true
authors.workspace = true
license.workspace = true
```

iot_packet_verifier/Cargo.toml
```diff
@@ -0,0 +1,42 @@
...
+tonic = {workspace = true}
+tracing = {workspace = true}
+tracing-subscriber = {workspace = true}
+triggered = {workspace = true}
+sha256 = "1.1.1"
+
+data-credits = {git = "https://github.com/helium/helium-program-library.git"}
+helium-sub-daos = {git = "https://github.com/helium/helium-program-library.git"}
+
+
```
jeffgrunewald, line 42:
```suggestion
```

iot_packet_verifier/Cargo.toml
```diff
@@ -0,0 +1,42 @@
...
+thiserror = {workspace = true}
+tokio = {workspace = true}
+tonic = {workspace = true}
+tracing = {workspace = true}
+tracing-subscriber = {workspace = true}
+triggered = {workspace = true}
+sha256 = "1.1.1"
+
+data-credits = {git = "https://github.com/helium/helium-program-library.git"}
+helium-sub-daos = {git = "https://github.com/helium/helium-program-library.git"}
```
jeffgrunewald, line 40: how does this one work with them being the same repo?

iot_packet_verifier/src/ingest.rs
```diff
@@ -0,0 +1,35 @@
+use chrono::{DateTime, Utc};
+use file_store::{FileStore, FileType};
+use futures::{Stream, StreamExt};
+use helium_proto::services::router::PacketRouterPacketReportV1;
```
jeffgrunewald, line 4:
```suggestion
use helium_proto::services::packet_router::PacketRouterPacketReportV1;
```
that proto message isn't defined in the `router.proto` service; looks like that one is just a stub at this point

iot_packet_verifier/Cargo.toml
maplant: When the git repo is a workspace, it will use the package name to find the source

jeffgrunewald (commented):
i had a really hard time sorting out everything the balance sufficiency enum was doing at first and i think it's a combination of the fact that on the surface it seems redundant that you would have a balance that's both sufficient/insufficient _and_ enabled/disabled (and it looks like you should be able to derive the desired behavior from just the enum variant or the boolean), but also that the call to mem::replace is swapping out the value of the balance struct's `enabled` field.

now that i feel i have a handle on what's going on i think what bothers me is how the enum function is reaching into the balance struct and mutating its state remotely. i think it would be a lot more straightforward to follow if we just treated the balance struct as a state machine and pass _it_ around as needed, and then update the `enable` field or its equivalent directly against the balance struct and take appropriate action against the org (enable/disable) based on the result of that operation (i.e. we could handle enabling/disabling the org within the same function that flips the boolean in the balance state machine as a result of the operation to validate the available balance all atomically)

maplant:
> i had a really hard time sorting out everything the balance sufficiency enum was doing at first and i think it's a combination of the fact that on the surface it seems redundant that you would have a balance that's both sufficient/insufficient _and_ enabled/disabled (and it looks like you should be able to derive the desired behavior from just the enum variant or the boolean), but also that the call to mem::replace is swapping out the value of the balance struct's `enabled` field.
> 
> now that i feel i have a handle on what's going on i think what bothers me is how the enum function is reaching into the balance struct and mutating its state remotely. i think it would be a lot more straightforward to follow if we just treated the balance struct as a state machine and pass _it_ around as needed, and then update the `enable` field or its equivalent directly against the balance struct and take appropriate action against the org (enable/disable) based on the result of that operation (i.e. we could handle enabling/disabling the org within the same function that flips the boolean in the balance state machine as a result of the operation to validate the available balance all atomically)

I think that the `BalanceSufficiency` enum is pretty terrible and I'm planning on removing entirely in my next push. The plan is to have whether or not we previously enabled or disabled an org to be cached separately, so that all we get back from balance is whether or not the we were able to debit successfully. Hopefully that fixes things up a bit

node_follower/src/settings.rs
```diff
@@ -1,5 +1,7 @@
 use crate::txn_service::{TransactionClient, TransactionClients};
-use helium_proto::services::{follower, transaction, Channel, Endpoint};
+use helium_proto::services::{
+    follower, iot_config::org_client::OrgClient, transaction, Channel, Endpoint,
```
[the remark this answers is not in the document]
maplant: I agree, this seemed like the easiest way to get things compiled but I'll move in into the packet verifier code.

iot_packet_verifier/migrations/3_burns.sql
```diff
@@ -0,0 +1,6 @@
+CREATE TABLE pending_burns (
+       id SERIAL PRIMARY KEY,
```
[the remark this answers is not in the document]
maplant: I think it's fine, there aren't that many payers I don't think.

iot_packet_verifier/src/daemon.rs
```diff
@@ -0,0 +1,168 @@
...
+                    &self.invalid_packets,
+                )
+                .await?;
+            meta::store(&mut transaction, "last_verified_report", last_verified).await?;
+            transaction.commit().await?;
+            self.valid_packets.commit().await?;
+            self.invalid_packets.commit().await?;
+        }
+
+        Ok(last_verified)
```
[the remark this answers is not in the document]
maplant: yeah, I think that's probably cleaner

iot_packet_verifier/migrations/3_burns.sql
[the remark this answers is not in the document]
jeffgrunewald: But it’s per txn entries, right? Will there just be 1 per org that lives forever and only ever increases or will there be one per burn txn?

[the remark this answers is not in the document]
maplant: Nah, these are per payer. Perhaps the name is a little confusing. When we complete a burn transaction, we subtract the amount burned from the column - we don't delete it.

iot_packet_verifier/src/pdas.rs
```diff
@@ -0,0 +1,15 @@
+use helium_crypto::PublicKeyBinary;
+use solana_sdk::pubkey::Pubkey;
+
+pub fn delegated_data_credits(sub_dao: &Pubkey, payer: &PublicKeyBinary) -> Pubkey {
```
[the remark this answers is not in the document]
maplant: I think a comment on the function and the module to explain that this is returning the Program Derived Address of the DelegatedDataCredits for the payer

andymck (commented):
This could do with a readme that outlines the operation of the verifier from a high level perspective and gives sufficient context to get the basic intent of the verifier.

maplant:
> This could do with a readme that outlines the operation of the verifier from a high level perspective and gives sufficient context to get the basic intent of the verifier.

That's a good idea, I'll add one today

iot_packet_verifier/src/verifier.rs
```diff
@@ -0,0 +1,490 @@
...
+            let signature = self.keypair.sign(&req.encode_to_vec())?;
+            req.signature = signature;
+            let _ = self.client.disable(req).await?;
+        }
+        Ok(())
+    }
+}
+
+#[async_trait]
+pub trait Burner {
```
jeffgrunewald, line 259: are some of these traits not in use yet? the Burner trait at least seems to be ignored by the Burner struct when it runs its loop

maplant: This is in use, it's different from the `Burner` struct. This trait is for inserting the burn amount into `pending_burns`, I just wasn't sure what to name it other than giving it the same name. Perhaps I'll rename the struct to `BurnDaemon` or something. This trait is used so that I can mock the database in unit tests

iot_packet_verifier/src/burner.rs
```diff
@@ -0,0 +1,240 @@
...
+            provider,
+            keypair: keypair.to_bytes(),
+        })
+    }
+
+    pub async fn run(mut self, shutdown: &triggered::Listener) -> Result<(), BurnError> {
+        let burn_service = task::spawn(async move {
+            loop {
+                self.burn().await?;
+                tokio::time::sleep(Duration::from_secs(1)).await;
```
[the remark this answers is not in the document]
maplant: I'm not sure how quickly the solana transaction will take, but it shouldn't really matter, unless the concern is missing the target of 1TPS. We reset the balance in the cache to force a re-check, and only subtract from the burn amount what we actually burned

iot_packet_verifier/src/verifier.rs
```diff
@@ -0,0 +1,489 @@
...
+
+    async fn debit_if_sufficient(
+        &self,
+        payer: &PublicKeyBinary,
+        amount: u64,
+    ) -> Result<bool, Self::Error>;
+}
+
+#[async_trait]
+impl Debiter for Arc<Mutex<HashMap<PublicKeyBinary, u64>>> {
```
bbalser, line 156: This impl is only used for tests? If so, can we put it in a `#[cfg(test)]`

iot_packet_verifier/src/verifier.rs
```diff
@@ -0,0 +1,489 @@
...
+        .bind(amount as i64)
+        .bind(Utc::now().naive_utc())
+        .fetch_one(&mut **self)
+        .await?;
+        Ok(())
+    }
+}
+
+#[async_trait]
+impl Burner for Arc<Mutex<HashMap<PublicKeyBinary, u64>>> {
```
bbalser, line 288: Is this impl just used for tests?

maplant: Yup, they both can go in there

iot_packet_verifier/src/settings.rs
```diff
@@ -0,0 +1,73 @@
...
+    pub cache: String,
+    /// Solana RpcClient URL:
+    pub solana_rpc: String,
+    /// Path to the keypair for signing burn transactions
+    pub burn_keypair: PathBuf,
+    /// Path to the keypair for signing config changes
+    pub config_keypair: PathBuf,
+    pub dc_mint: String,
+    pub dnt_mint: String,
+    pub hnt_mint: String,
```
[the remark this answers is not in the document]
maplant: Mint is a bs58 encoded string, unfortunately I can't use solana pubkeys as those deserialize from arrays.

iot_packet_verifier/src/settings.rs
```diff
@@ -0,0 +1,73 @@
...
+
+    pub fn dnt_mint(&self) -> Result<Pubkey, ParsePubkeyError> {
+        self.dnt_mint.parse()
+    }
+
+    pub fn hnt_mint(&self) -> Result<Pubkey, ParsePubkeyError> {
+        self.hnt_mint.parse()
+    }
+
+    pub fn config_keypair(&self) -> Result<helium_crypto::Keypair, Box<helium_crypto::Error>> {
```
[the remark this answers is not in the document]
maplant: Unfortunately it's not super useful because the keypair functions that solana uses returns a `Box<dyn Error>`, which can't be converted into an anyhow::Error

iot_packet_verifier/src/ingest.rs
maplant: router includes packet_router: https://github.com/helium/proto/blob/master/src/lib.rs#L36

iot_packet_verifier/src/burner.rs
```diff
@@ -0,0 +1,242 @@
...
+            UPDATE pending_burns SET
+              amount = amount - $1,
+              last_burn = $2
+            WHERE id = $3
+            "#,
+        )
+        .bind(amount)
+        .bind(Utc::now().naive_utc())
+        .bind(id)
+        .execute(&self.pool)
```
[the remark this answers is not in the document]
maplant: I don't think so, there's no other sql that needs to be run before the transaction has occurred

iot_packet_verifier/src/burner.rs
```diff
@@ -0,0 +1,242 @@
...
+#[derive(FromRow, Debug)]
+pub struct Burn {
+    pub id: i32,
+    pub payer: PublicKeyBinary,
+    pub amount: i64,
+}
+
+/// Cached pubkeys for the burn program
+pub struct BurnProgramCache {
+    pub account_payer: Pubkey,
```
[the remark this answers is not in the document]
maplant: I agree with you, but that's the name used in the solana program :-/

iot_packet_verifier/src/daemon.rs
maplant: This has been changed to use `continuous_source`, which should be far more robust.

bbalser (approved):

iot_packet_verifier/Cargo.toml
```diff
@@ -37,7 +37,7 @@ tracing-subscriber = {workspace = true}
 triggered = {workspace = true}
 http = {workspace = true}
 http-serde = {workspace = true}
-sha256 = "1.1.1"
+sha2 = "0.10"
```
madninja, line 40: why not {workspace = true}?

maplant: Just added that dep. Will change

iot_packet_verifier/src/burner.rs
```diff
@@ -0,0 +1,252 @@
...
+            "Successfully burned data credits. Transaction: {}",
+            signature
+        );
+
+        // Now that we have successfully executed the burn and are no long in
+        // sync land, we can remove the amount burned.
+        sqlx::query(
+            r#"
+            UPDATE pending_burns SET
+              amount = amount - $1,
```
jeffgrunewald, line 180: in theory this should just be $0 right, since you got the `amount` value from a query at the beginning of this function? if there's any chance this value could be updated out-of-band from this burn function should we enforce some guard that ensures that `amount` in the DB never goes into the negative?

iot_packet_verifier/src/verifier.rs
```diff
@@ -0,0 +1,496 @@
...
+        oui: u64,
+        cache: &mut HashMap<u64, PublicKeyBinary>,
+    ) -> Result<PublicKeyBinary, Self::Error> {
+        if let Entry::Vacant(e) = cache.entry(oui) {
+            let req = OrgGetReqV1 { oui };
+            let pubkey =
+                PublicKeyBinary::from(self.client.get(req).await?.into_inner().org.unwrap().owner);
+            e.insert(pubkey);
+        }
+        Ok(cache.get(&oui).unwrap().clone())
```
jeffgrunewald, line 214: stray unwraps?

iot_packet_verifier/src/verifier.rs
```diff
@@ -0,0 +1,496 @@
...
+
+#[async_trait]
+pub trait Burner {
+    type Error;
+
+    async fn burn(&mut self, payer: &PublicKeyBinary, amount: u64) -> Result<(), Self::Error>;
+}
+
+#[async_trait]
+impl Burner for &'_ mut Transaction<'_, Postgres> {
```
jeffgrunewald, line 254: i'm still not a fan of the overloading of the name `Burner` since we've got a `Burner` struct that routinely does burn operations and then we've got a `Burner` trait in the same crate, both of which define a `burn()` function that writes to the same database table but other than the table there's nothing that actually links these two parts of the dc burn process so i think this could be ripe for a refactor bug in the future with these being defined in separate files/modules and the names implying they're more related than two separate processes writing to the same table.

at the very least i think this should be renamed to something distinct from the `Burner` struct in the `burner.rs` module like `BurnMeter` with a `add_burn()` function so long as this whole thing can't be threaded through the `Burner` struct/process itself and have _all_ of the account burn access to the database be managed by a single process (such as with a message channel or similar)

iot_packet_verifier/src/verifier.rs
maplant: This is guaranteed to never panic, an insert always occurs before this if no entry was found

iot_packet_verifier/src/burner.rs
maplant: It _may_ be zero or greater than. It is possible for the verification process to add burn amounts during this - or at least ideally it can. No other process can subtract from the burn amount besides this one though, so it will never become negative

iot_packet_verifier/src/verifier.rs
maplant: You're right, I'll rename this to `PendingBurns`

iot_packet_verifier/src/burner.rs
jeffgrunewald: 👍

jeffgrunewald (approved):

PR #275, opened 2022-12-20
correctly convert decimals back and forth

jeffgrunewald:
this change corrects a mismatch in the amount of precision being applied to hex scaling factors and reward units when they are serialized to protobuf as u32s for transit over the wire and then back again to decimals on the other side.

we are applying 4 places of decimal precision when creating the decimal value but when converting to a u32 we have not been multiplying by the correct factor to re-initialize the desired decimal value on the other side of the report proto conversion


vihu (approved):

madninja (commented):
I guess there was no good way to actually define a Decimal type alias here? sorta like we have for `Hnt` and `Hst` and `Dbi` but then add one for `RewardUnit`? or something?

bbalser (approved):

PR #277, opened 2023-01-03
dedup packet data hashes

andymck:
Dedup the list of packet data hashes, derived from processed beacons, before initializing the xor filter.  The xor filter library requires a deduped list and will hang if dups are present


poc_iot_verifier/Cargo.toml
```diff
@@ -51,3 +51,4 @@ rust_decimal = {workspace = true, features = ["maths"]}
 rust_decimal_macros = {workspace = true}
 humantime = {workspace = true}
 twox-hash = {workspace = true}
+itertools = "*"
```
bbalser, line 54: We are also using `itertools` in `density_scalar` so we should probably move `itertools` up to the main `Cargo.toml`.

andymck: fixed here: db51b68a7b91454cc309fa0e92ea7e75463c3e01

bbalser (approved):

poc_iot_verifier/src/loader.rs
```diff
@@ -184,7 +185,7 @@ impl Loader {
             ),
         }
         tracing::info!("creating beacon xor filter");
-        let beacon_packet_data = xor_data.into_inner();
+        let beacon_packet_data: Vec<u64> = xor_data.into_inner().into_iter().unique().collect();
```
JayKickliter, line 188: This should be slightly faster, and doesn't require hashing like the itertools's version.

```suggestion
        let mut beacon_packet_data = xor_data.into_inner();
        beacon_packet_data.sort();
        beacon_packet_data.dedup();
```

JayKickliter (approved):

JayKickliter:
Unrelated, but while reviewing this PR I found a potential _minor_ optimization point: move the  `filter_key_hash` call to before the lock
https://github.com/helium/oracles/blob/9d416328d379c3b5942a6d551f5ac7826cc93118/poc_iot_verifier/src/loader.rs#L350

PR #279, opened 2023-01-04
Fix witness error handling

andymck:
This addresses an issue with how the refactored POC module handled errors after the prior merged `anyhow` changes.

If the handling of any one of N witnesses for a beacon resulted in an error ( such as when a witness has no asserted location ) then that would halt any further processing of remaining witnesses and also result in the beacon itself failing to complete verification.

The DB would also not be updated to bump the `attempts` count on either the failing witness or associated beacon and thus the POC would get stuck in a loop of being verified and failing each tick until they end up being purged by the purger.

Thus this PR makes the following logic assumptions:

1.   `poc::verify_beacon` and `poc::verify_witnesses` will return Result<VerifyResult, VerificationError>.  VerificationError is only returned in the event the verification cannot proceed/complete.  In such a scenario, the caller which in this case is the runner will update the attempts count for the beacon in the DB.  Once it exceeds MAX_ATTEMPTS it will not be reselected again for verification
2. An error which occurs whilst verifying any one individual witness will not block verification of subsequent witnesses and will not propagate an error back to the caller ( ie the runner )
3. If an individual verification ( such as verify_gw_location or verify_witness_distance etc ) cannot derive the data it requires in order to complete verification ( such as the gateway location ) then it will not throw an error but instead return an expected result which renders the verification result as `InValid` and in handled within the poc::verify_witnesses API itself.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1877,7 +1877,7 @@ dependencies = [
 [[package]]
 name = "helium-proto"
 version = "0.1.0"
-source = "git+https://github.com/helium/proto?branch=master#c7094737694d22245f9db6977868355ae99b6e74"
+source = "git+https://github.com/helium/proto?branch=andymck/extend-invalid-reason-generic-error#08a19674a2dc9daafa4c0d6495e4f7b46652262b"
 dependencies = [
  "bytes",
  "prost",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1877,7 +1877,7 @@ dependencies = [
 [[package]]
 name = "helium-proto"
 version = "0.1.0"
-source = "git+https://github.com/helium/proto?branch=andymck/extend-invalid-reason-generic-error#08a19674a2dc9daafa4c0d6495e4f7b46652262b"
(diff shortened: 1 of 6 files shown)


poc_iot_verifier/src/poc.rs
```diff
@@ -228,28 +228,13 @@ impl Poc {
...
-                witness_report,
-                beaconer_info,
-                witness_distance,
-            )
+            .do_witness_verifications(&witness_info, witness_report, beaconer_info)
             .await
         {
             Ok(()) => {
-                if let Some(hex_scale) = hex_density_map.get(witness_location).await {
+                if let Some(hex_scale) = hex_density_map.get(witness_info.location.unwrap()).await {
```
bbalser, line 237: Don't we need to handle this `unwrap` better?

andymck: so at this point the location has already been verified and thus we know unwrap will work.

i did attempt to not use unwrap like this but i end up doing something elaborate and it feels like just for the sake of it.

Open to suggestions

bbalser (approved):

bbalser (approved):

jeffgrunewald (approved):

PR #280, opened 2023-01-04
Update mobile ingest to accept data transfer session requests

bbalser:


poc_ingest/src/server_5g.rs
```diff
@@ -105,6 +109,36 @@ impl poc_mobile::PocMobile for GrpcServer {
...
+            .verify_public_key(event.pub_key.as_ref())
+            .and_then(|public_key| self.verify_network(public_key))
+            .and_then(|public_key| self.verify_signature(public_key, event))
+            .map(|(_, event)| DataTransferSessionIngestReportV1 {
+                received_timestamp: timestamp,
+                report: Some(event),
+            })?;
+
+        _ = file_sink_write!(
+            "data_transfer_session_report",
```
jeffgrunewald, line 130: any chance these metrics could get mixed up with iot network data transfer metrics without namespacing them like `iot_data_transfer_session_report` or will the metrics from the two networks be kept totally separate within the monitoring environment? references to heartbeats and speedtests seem pretty unique to the mobile network so i get those not needing it but this should be a common metrics across all networks

madninja: The grafana config has a network field or something equivalent to distinguish them right?

madninja (approved):

jeffgrunewald (approved):

PR #283, opened 2023-01-04
Hold on to locks as little as necessary

JayKickliter:
I'm pretty sure that moving the call to `filter_key_hash` to before lock acquire time makes sense, if perhaps with little runtime benefit.

The bigger question is how locks are used in purger. From what I can see, the current use of [`for_each_concurrent`]() is not concurrent at all since the closures are all serialized on [acquiring a lock]()


bbalser (approved):

andymck (approved):
lgtm, @bbalser best placed to comment on the purger side as he implemented the locks there.

jeffgrunewald (approved):

PR #286, opened 2023-01-05
Rename poc_ingest to ingest

bbalser:


jeffgrunewald (commented):
2 questions since this will likely come up eventually:
- do we want to rename the operating modes for the ingest server from `5g` and `lora` to `mobile` and `iot`?
- do these python test scripts still work such that they're worth renaming or are they so outdated they should just be deleted instead?

madninja:
> 2 questions since this will likely come up eventually:
> 
> * do we want to rename the operating modes for the ingest server from `5g` and `lora` to `mobile` and `iot`?

that’d be great. Yes please

ingest/src/settings.rs
```diff
@@ -49,7 +49,7 @@ pub fn default_sink() -> String {
 #[derive(Debug, Deserialize)]
 #[serde(rename_all = "lowercase")]
 pub enum Mode {
-    Lora,
+    IOT,
```
madninja, line 52: I'm not sure the all caps `IOT` works. I think we should just use `Iot` to avoid  enum case clippy issues

jeffgrunewald: + 1

bbalser: Fixed d9660c6

jeffgrunewald (approved):

madninja (approved):

PR #287, opened 2023-01-05
Initial IoT Config Service

jeffgrunewald:
This PR implements the initial IoT Config oracle service that will be used by helium network users for organization and routing management as well as supply hotspots with region information upon request, all via grpc apis. It implements:
- Basic oracle server structure, run loop, config management and packaging requirements
- Implementations of the grpc traits defined in the proto repo for the `Org` and `Route` services (minus the route disable api)

This is the first of several PRs to break up the review and development of the config service. The following changes are _not_ yet implemented and will follow in subsequent PRs:
- Refactor service backends behind org/route traits for easier testing and separation of database access from data manipulation
- Authentication and authorization checks against registered admin and org management keys
- Additional validations of devaddr range constraints, nonce incrementation, and timestamp replay prevention
- Session Key Filter management
- Gateway region lookup


iot_config/src/main.rs
```diff
@@ -0,0 +1,98 @@
...
+impl Cmd {
+    pub async fn run(&self, settings: Settings) -> Result<()> {
+        match self {
+            Self::Server(cmd) => cmd.run(&settings).await,
+        }
+    }
+}
+
+#[derive(Debug, clap::Args)]
+pub struct Daemon {}
```
JayKickliter, line 51: FYI, field less structs do not need the `{}`:
```rust
pub struct Daemon;
```

iot_config/src/region.rs
```diff
@@ -0,0 +1,128 @@
+use helium_proto::Region as ProtoRegion;
+use serde::{de, Deserialize, Deserializer, Serialize};
+use std::{fmt, str::FromStr};
+
+#[derive(Clone, Debug, Eq, Hash, Ord, PartialEq, PartialOrd)]
+pub enum Region {
```
JayKickliter, line 6: It kills me that we define these regions defined in so many places. I'm not sure what the best solution is.

JayKickliter: can we use the photo region enum as the single source of truth?

iot_config/src/lora_field.rs
```diff
@@ -0,0 +1,516 @@
+use serde::{Deserialize, Deserializer, Serialize, Serializer};
+use std::{fmt::Display, str::FromStr};
+
+#[derive(Clone, Copy, Debug, Default, Eq, PartialEq)]
+pub struct LoraField<const WIDTH: usize>(pub u64);
```
JayKickliter, line 5: A quick scan through this code doesn't make it obvious what this is for, but I'm guessing it has to do with parsing?

iot_config/pkg/settings-template.toml
```diff
@@ -0,0 +1,23 @@
+# log settings for the application (RUST_LOG format). Default below
+# 
+# log = "poc-iot-verifier=debug,poc_store=info"
```
bbalser, line 3: comment needs updated

iot_config/src/region.rs
vihu: +1 ^

iot_config/src/main.rs
jeffgrunewald: good to know; surprised that's not something clippy would complain about

iot_config/src/lora_field.rs
jeffgrunewald: correct; this is for bit twiddling NetIDs, DevAddrs, and Euis; just below that i implement each of these:
```
pub type NetIdField = LoraField<6>;
pub type DevAddrField = LoraField<8>;
pub type EuiField = LoraField<16>;
```
and then further down have impl blocks that add specific behavior on those three type. this is liberally stolen from work @michaeldjeffrey did in the rust CLI

iot_config/pkg/settings-template.toml
jeffgrunewald: good eye; fixed

iot_config/src/region.rs
jeffgrunewald: full disclosure, i cargo culted (ha!) this module as well before i was really using it but we can't use the proto enum out of the box b/c we need the proto region to implement `Deserialize` in order to implement serde for the `GwmpMap` type in the route module but if i'm able to update this module to just wrap the proto enum in a new type or barring that, implement Serialize and Deserialize for the region _in_ the proto repo then we can at least simplify this

madninja: We should be able to derive serialize for the region struct.

madninja: there are pattern matches and directives for Prost to declare attributes for structs

jeffgrunewald: Lemme give that a shot then. That would certainly be the cleanest solution

jeffgrunewald: unless there's a more succinct way of defining the derive for deserialize on the region proto, https://github.com/helium/proto/pull/259 does appear to do the trick; i've eliminated the local region module entirely as a result in this latest commit

bbalser (approved):

iot_config/migrations/2_organizations.sql
```diff
@@ -0,0 +1,29 @@
...
+
+    inserted_at timestamptz not null default now(),
+    updated_at timestamptz not null default now()
+);
+
+select trigger_updated_at('organizations');
+
+create table organization_devaddr_constraints (
+    oui bigint not null references organizations(oui) on delete cascade,
+    nwk_id int not null,
```
michaeldjeffrey, line 21: Is the `nwk_id` meant to be the full `net_id`?
The same `nwk_id` can appear across the different `types` of `net_id`.

iot_config/migrations/3_routes.sql
```diff
@@ -0,0 +1,36 @@
...
+    inserted_at timestamptz not null default now(),
+    updated_at timestamptz not null default now()
+);
+
+select trigger_updated_at('route_eui_pairs');
+
+create table route_devaddr_ranges (
+    route_id uuid not null references routes(id) on delete cascade,
+    start_nwk_addr int not null,
+    end_nwk_addr int not null,
```
michaeldjeffrey, line 30: Are these only the `nwk_addr` bits of a DevAddr?

jeffgrunewald: crap; no, that should be bigint and encompass the full integer of the devaddr; it looks like they originally started out as bigints in the _original_ route migrations for the elixir service and were downgraded to integers for some reason and that's what i carried over here

iot_config/migrations/2_organizations.sql
jeffgrunewald: this was just translated over from the elixir service so you tell me if this should be the full net_id or if the nwk_id would suffice? the only place we're currently using it here is in filtering db results to get the next available helium devaddr. in the elixir service it looks like it's primarily used for address parsing (which we don't need here) and later in route validation.

is it fair to say that using it for route validation is still a valuable use case or do you think we should switch our use of `nwk_id` here for `net_id` in general? would all of the devaddr constraints that make up the available helium address space share a nwk_id or would they all be different?

michaeldjeffrey: I think using only `nwk_id` for route validation would be a bug. 

That would mean a Route could contain a valid devaddr range across all 7 types of NetIDs and we wouldn't necessarily be able to know which is correct.

> would all of the devaddr constraints that make up the available helium address space share a nwk_id or would they all be different?

A DevAddr contraint spanning multiple `nwk_id` would cross multiple NetIDs, which I believe should not be allowed.

jeffgrunewald: switched the use of `nwk_id` to `net_id` as the 64-bit integer version of the type

iot_config/src/lora_field.rs
```diff
@@ -0,0 +1,516 @@
...
+    fn from(field: LoraField<WIDTH>) -> Self {
+        field.0 as u32
+    }
+}
+
+impl<const WIDTH: usize> From<LoraField<WIDTH>> for i32 {
+    fn from(field: LoraField<WIDTH>) -> Self {
+        field.0 as i32
+    }
+}
```
michaeldjeffrey, line 120: Would it make sense to only provide `64` conversions for `EuiField`, `32` conversions for `DevAddrField` and `NetIdField`?

In hopes if disallowing conversions that would remove necessary bits.

jeffgrunewald: That’s a good call; I think most of those were a holdover from morphing db requirements that are no longer relevant

jeffgrunewald: fixed [4c94984](https://github.com/helium/oracles/pull/287/commits/4c949841518190a56205e8e3ccb88b1138ed09b4)

iot_config/src/route_service.rs
```diff
@@ -0,0 +1,200 @@
...
+}
+
+#[tonic::async_trait]
+impl iot_config::Route for RouteService {
+    async fn list(&self, request: Request<RouteListReqV1>) -> GrpcResult<RouteListResV1> {
+        let request = request.into_inner();
+
+        let proto_routes: Vec<RouteV1> = route::list_routes(request.oui, &self.pool)
+            .await
+            .map_err(|_| Status::internal("route list failed"))?
```
andymck, line 49: i notice all error response are of `internal` type.  Should we not be more descriptive where possible ?  For example if the user attempts to list or get an invalid route, maybe a not found response

jeffgrunewald: Yes, they should probably be more descriptive and specific

jeffgrunewald: so i went back over them and made a few updates but for the most part these all amount to "something went wrong in the db update" so i could be a little more explicit in that regard with the message, but i wouldn't have a better suggestion for which _type_ of status error each one should have than the current ones (options defined here: https://docs.rs/tonic/latest/tonic/struct.Status.html); i'm open to arguments for/against and in particular.

i also think as the API is fleshed out in subsequent changes we'll have alternative failure scenarios like `unauthorized` that just aren't relevant with this initial build

andymck (approved):
lgtm

One note is some of the sql queries look intensive, are any indexes required ? i suspect the volume of data is very low to perhaps not important at this point

maplant (approved):
Looks good to me!

jeffgrunewald:
> One note is some of the sql queries look intensive, are any indexes required ? i suspect the volume of data is very low so perhaps not important at this point

Good point; it took a while to load up a suitable sample of data in a test db but after playing with the queries, particularly the route-related ones, I've added a few indexes https://github.com/helium/oracles/blob/272de7e98718f0c276f6fc6b44d66dbc5b318515/iot_config/migrations/3_routes.sql that really improved performance on all but the insert and delete operations on the `route_devaddr_ranges` table and those only increased by 0.4-0.5ms

PR #289, opened 2023-01-09
Add reason to metrics for invalid beacon and witnesses

bbalser:


andymck (approved):
lgtm, assuming no other code needs updated to use the updated file sink write

bbalser:
before this change we were not using the optional 4th parameter to `file_sink_write!` anywhere, so changing it affecting no other oracles

jeffgrunewald (approved):

PR #291, opened 2023-01-11
Add support to squish witnesses per receipt

vihu:
- Add `max_witnesses_per_receipt` setting. Default = 14.
- Maybe squish poc_witnesses if the length of witnesses is > max_witnesses_per_receipt. Seed a random number from the poc_id for shuffling the witnesses and truncating the witnesses

### DIFF
--- a/poc_iot_injector/src/receipt_txn.rs
+++ b/poc_iot_injector/src/receipt_txn.rs
@@ -72,11 +72,8 @@ fn maybe_squish_witnesses(
     }
 
     // Seed a random number from the poc_id for shuffling the witnesses
-    let mut hasher = Sha256::new();
-    hasher.update(poc_id);
-    let mut seed: [u8; 32] = [0; 32];
-    seed.copy_from_slice(&hasher.finalize());
-    let mut rng = StdRng::from_seed(seed);
+    let seed = Sha256::digest(poc_id);
+    let mut rng = StdRng::from_seed(seed.into());
 
     // Shuffle and truncate witnesses
     poc_witnesses.shuffle(&mut rng);
--- a/poc_iot_injector/src/receipt_txn.rs
+++ b/poc_iot_injector/src/receipt_txn.rs
@@ -7,6 +7,7 @@ use helium_proto::{
     blockchain_txn::Txn, BlockchainPocPathElementV1, BlockchainPocReceiptV1,
     BlockchainPocWitnessV1, BlockchainTxn, BlockchainTxnPocReceiptsV2, Message,
 };
+use rand::{rngs::StdRng, seq::SliceRandom, SeedableRng};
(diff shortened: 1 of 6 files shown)


jeffgrunewald (approved):
impl looks good to me. i'm curious where the 14 default count came from (looks like `poc_per_hop_max_witnesses` chain var but more wondering where that came from) and if we want to bother truncating the total per-poc witness count if it makes sense to do so only temporarily or else to move this operation into the verifier and keep it beyond the solana migration?

andymck (approved):
Surprised to see we are adding this limit back in and with an equally low count to the current chain value.   I thought we have headroom now for a much higher count ?

Other than that tho lgtm

madninja:
> Surprised to see we are adding this limit back in and with an equally low count to the current chain value. I thought we have headroom now for a much higher count ?
> 
> Other than that tho lgtm

The problem isn't just the chain but

1.  the downstream impacts of this. With 100s of witnesses for some there are likely going to be issues with, for example ETL
2. If unlimited witnesses are allowed a large number of witnesses can drown out the witness rewards for that report right? If I can get/game 1000 witnesses against any other 5/10 ones I'd basically earn all the witness rewards?

andymck:
My surprise was more the limit being so low again, ie 14.  That is no improvement over what we have already, despite the extra headroom ( at least on the chain )

poc_iot_injector/src/receipt_txn.rs
```diff
@@ -54,6 +61,28 @@ pub fn handle_report_msg(
...
+    poc_witnesses: &mut Vec<BlockchainPocWitnessV1>,
+    poc_id: &Vec<u8>,
+    max_witnesses_per_receipt: usize,
+) {
+    if poc_witnesses.len() <= max_witnesses_per_receipt {
+        return;
+    }
+
+    // Seed a random number from the poc_id for shuffling the witnesses
+    let mut hasher = Sha256::new();
```
madninja, line 75: Use the Digest trait and the fact that GenericArray works for slices just fine:

```rust
    let seed = Sha256::digest(poc_id);
    let mut rng = StdRng::from_seed(seed.into());
```

madninja:
> My surprise was more the limit being so low again, ie 14. That is no improvement over what we have already, despite the extra headroom ( at least on the chain )

Let's powwow on today, please. While more is certainly awesome it has reward implications we should understand

poc_iot_injector/src/receipt_txn.rs
vihu: Oh nice, updated [6fc793e](https://github.com/helium/oracles/pull/291/commits/6fc793ede9854eecc9e7a197ad3c6675c1145e3d)

madninja (approved):

PR #292, opened 2023-01-11
Timing related tweaks and conflict support for bulk inserts

andymck:
- Adds conflict strategy for bulk inserts. If the verifier is abruptly shutdown or crashes there is a potential to reload reports which were previously loaded.  Without the `conflict` strategy we get a lot of noisy errors polluting the logs
- Extends expected beacon interval from 10 mins to 6 hours and define beacon intervals as duration rather than i64s
- Explicitly define beacon interval tolerance and enhance related test
- Set interval before a stale beacon or witness report is purged to the same value

### DIFF
--- a/iot_verifier/src/poc.rs
+++ b/iot_verifier/src/poc.rs
@@ -16,6 +16,7 @@ use helium_proto::{
     services::poc_iot::{InvalidParticipantSide, InvalidReason},
     GatewayStakingMode, Region,
 };
+use lazy_static::lazy_static;
 use node_follower::gateway_resp::GatewayInfo;
 use rust_decimal::Decimal;
 use sqlx::PgPool;
@@ -28,9 +29,13 @@ pub const C: f64 = 2.998e8;
 /// R is the (average) radius of the earth
 pub const R: f64 = 6.371e6;
 
-/// the cadence in seconds at which hotspots are permitted to beacon
-/// 6 hours with 10 min tolerance
-const BEACON_INTERVAL: i64 = (60 * 60 * 6) - (10 * 60);
+lazy_static! {
+    /// the cadence at which hotspots are permitted to beacon
+    static ref BEACON_INTERVAL: Duration = Duration::hours(6);
+    /// a tolerance applied to beacon intervals within which beacons
+    /// will be accepted
+    static ref BEACON_INTERVAL_TOLERANCE: Duration = Duration::minutes(10);
+}
 /// max permitted distance of a witness from a beaconer measured in KM
 const POC_DISTANCE_LIMIT: i32 = 100;
 
@@ -377,7 +382,7 @@ fn verify_beacon_schedule(
     match last_beacon {
         Some(last_beacon) => {
             let interval_since_last_beacon = beacon_received_ts - last_beacon.timestamp;
-            if interval_since_last_beacon < Duration::seconds(BEACON_INTERVAL) {
+            if interval_since_last_beacon < (*BEACON_INTERVAL - *BEACON_INTERVAL_TOLERANCE) {
                 tracing::debug!(
                     "beacon verification failed, reason:
                         IrregularInterval. Seconds since last beacon {:?}",
@@ -772,19 +777,43 @@ mod tests {
     fn test_verify_beacon_schedule() {
         let now = Utc::now();
         let id: &str = "test_id";
-        let last_beacon_a = Some(LastBeacon {
+        let last_beacon = Some(LastBeacon {
             id: id.as_bytes().to_vec(),
-            timestamp: now - Duration::seconds(60 * 12),
+            timestamp: now - *BEACON_INTERVAL,
         });
-        // last beacon was 12 mins in the past, expectation pass
-        assert_eq!(Ok(()), verify_beacon_schedule(&last_beacon_a, now));
-        //we dont have any last beacon data, expectation pass
-        assert_eq!(Ok(()), verify_beacon_schedule(&None, now));
-        // too soon after our last beacon, expectation fail
+        // last beacon was BEACON_INTERVAL in the past, expectation pass
+        assert_eq!(Ok(()), verify_beacon_schedule(&last_beacon, now));
+        // last beacon was BEACON_INTERVAL + 1hr in the past, expectation pass
+        assert_eq!(
+            Ok(()),
+            verify_beacon_schedule(&last_beacon, now + Duration::minutes(60))
+        );
+        // last beacon was BEACON_INTERVAL - 1 hr, too soon after our last beacon,
+        // expectation fail
         assert_eq!(
             Err(InvalidReason::IrregularInterval),
-            verify_beacon_schedule(&last_beacon_a, now - Duration::seconds(60 * 5))
+            verify_beacon_schedule(&last_beacon, now - Duration::minutes(60))
         );
+        // last beacon was just outside of our tolerance period by 2 mins
+        // therefore beacon too soon, expectation fail
+        assert_eq!(
+            Err(InvalidReason::IrregularInterval),
+            verify_beacon_schedule(
+                &last_beacon,
+                now - (*BEACON_INTERVAL_TOLERANCE + Duration::minutes(2))
+            )
+        );
+        // last beacon was just inside of our tolerance period by 2 mins
+        // expectation pass
+        assert_eq!(
+            Ok(()),
+            verify_beacon_schedule(
+                &last_beacon,
+                now - (*BEACON_INTERVAL_TOLERANCE - Duration::minutes(2))
+            )
+        );
+        //we dont have any last beacon data, expectation pass
+        assert_eq!(Ok(()), verify_beacon_schedule(&None, now));
     }
 
     #[test]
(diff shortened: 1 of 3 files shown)


andymck:
> So if there was a transaction around the inserts for a file this would (a) likely speed up inserts and (b) avoid man conflict errors?

the inserts are already as part of a txn, see here: https://github.com/helium/oracles/blob/main/iot_verifier/src/loader.rs#L303

Doesnt avoid conflict errors tho

madninja:
> > So if there was a transaction around the inserts for a file this would (a) likely speed up inserts and (b) avoid man conflict errors?
> 
> the inserts are already as part of a txn, see here: https://github.com/helium/oracles/blob/main/iot_verifier/src/loader.rs#L303
> 
> Doesnt avoid conflict errors tho

Not completely, no but 

> If the verifier is abruptly shutdown or crashes there is a potential to reload reports 

it should avoid all of those right?

andymck:
the txn will not avoid potential for conflicts.  Each incoming file from s3 is processed concurrently and there are multiple batches per file...each batch wrapped in its own txn.  As such its is certainly possible to process one or more batches within a loading window and then for the app to be abruptly stopped.  Potentially neither any one full file or all files within the window may have been processed.  The time for the last file may not have been advanced and as such when we start back up again we process the same files again and thus the resulting duplicates

andymck:
Correction, we have a single txn which handles all batches

madninja:
> Correction, we have a single txn which handles all batches
> 
> But same principle applies, we do not update the last window time until all files in the window is complete. As such terminating the app may result in that window time not being updated and we may already have processed some files

Ah right.. We're re-processing the files, but the transaction guarantees that any inserts that were in the aborted transaction were not inserted, right? Hence fewer conflicts on reprocessing the same files?

andymck:
Yes, the txn should guarantee that any inserts from an aborted file dont make it to the DB.  Conflicts will only occur with files that were processed to completion and for which the txn was committed.

madninja:
> Yes, the txn should guarantee that any inserts from an aborted file dont make it to the DB. Conflicts will only occur with files that were processed to completion and for which the txn was committed.

Ah ok.. your first bullet read like it was implying something else

iot_verifier/src/poc.rs
```diff
@@ -29,7 +29,10 @@ pub const C: f64 = 2.998e8;
 pub const R: f64 = 6.371e6;
 
 /// the cadence in seconds at which hotspots are permitted to beacon
-const BEACON_INTERVAL: i64 = (10 * 60) - 10; // 10 mins ( minus 10 sec tolerance )
+const BEACON_INTERVAL: i64 = 60 * 60 * 6;
+/// a tolerance applied to beacon intervals within which beacons
+/// will be accepted
+const BEACON_INTERVAL_TOLERANCE: i64 = 10 * 60;
```
madninja, line 35: and that's in seconds? Milis? add a comment?

andymck: fixed here: 2900708ca135d5b2a0f7781274b0e1f8483b48c1

iot_verifier/src/poc.rs
```diff
@@ -29,7 +29,10 @@ pub const C: f64 = 2.998e8;
 pub const R: f64 = 6.371e6;
 
 /// the cadence in seconds at which hotspots are permitted to beacon
-const BEACON_INTERVAL: i64 = (10 * 60) - 10; // 10 mins ( minus 10 sec tolerance )
+const BEACON_INTERVAL: i64 = 60 * 60 * 6;
```
madninja, line 32: Use `Duration::hours`

iot_verifier/src/poc.rs
```diff
@@ -29,7 +29,10 @@ pub const C: f64 = 2.998e8;
...
+/// a tolerance ( in seconds ) applied to beacon intervals within which beacons
+/// will be accepted
+const BEACON_INTERVAL_TOLERANCE: i64 = 10 * 60;
```
madninja, line 35: Use `Duration::minutes`

iot_verifier/src/poc.rs
```diff
@@ -662,19 +667,46 @@ mod tests {
...
-        assert_eq!(Ok(()), verify_beacon_schedule(&last_beacon_a, now));
-        //we dont have any last beacon data, expectation pass
-        assert_eq!(Ok(()), verify_beacon_schedule(&None, now));
-        // too soon after our last beacon, expectation fail
+        // last beacon was BEACON_INTERVAL in the past, expectation pass
+        assert_eq!(Ok(()), verify_beacon_schedule(&last_beacon, now));
+        // last beacon was BEACON_INTERVAL + 1hr in the past, expectation pass
+        assert_eq!(
+            Ok(()),
+            verify_beacon_schedule(&last_beacon, now + Duration::seconds(60 * 60))
```
madninja, line 679: Use `Duration::minutes`?

madninja (commented):
Can you update to use Duration for the const and anywhere you use Duration constructors with multipliers use a higher level Duration constructor?

bbalser (approved):

bbalser (approved):

iot_verifier/src/poc.rs
```diff
@@ -28,8 +29,13 @@ pub const C: f64 = 2.998e8;
...
 
-/// the cadence in seconds at which hotspots are permitted to beacon
-const BEACON_INTERVAL: i64 = (10 * 60) - 10; // 10 mins ( minus 10 sec tolerance )
+lazy_static! {
+    /// the cadence at which hotspots are permitted to beacon
+    static ref BEACON_INTERVAL: Duration = Duration::hours(6);
+    /// a tolerance applied to beacon intervals within which beacons
+    /// will be accepted
+    static ref BEACON_INTERVAL_TOLERANCE: Duration = Duration::minutes(10);
+}
```
jeffgrunewald, line 38: are these values we will want to tweak often enough for testing purposes such that they make more sense as a configurable `Setting`? not sure if we're still expecting to eventually implement a "chain var"-style network variable configuration feature but these seem like they would be good candidates for that type of setting as well if we do

if these are just statically defined constants i don't see the point in having them be separate values as effectively we're just hard-coding the acceptable beacon interval to 5 hours and 5 min. i think these make more sense as independent values when it's capable of tuning them from outside the app without a code release

iot_verifier/src/poc_report.rs
```diff
@@ -16,6 +16,7 @@ const REPORT_INSERT_SQL: &str = "insert into poc_report (
     report_type,
     status
 ) ";
+const REPORT_INSERT_CONFLICT_SQL: &str = " on conflict (id) do nothing ";
```
jeffgrunewald, line 19: if the poc reports table is the only thing db table we're interacting with in this module and since these consts aren't public it's probably okay, but having this independent query snippet without a table reference seems like it could be a footgun lying in wait for us; maybe ?
```suggestion
const REPORT_INSERT_CONFLICT_SQL: &str = " on conflict (poc_report.id) do nothing ";
```

iot_verifier/src/poc.rs
andymck: These are unlikely to be modified often.  The value in having them separate is that it avoids any confusion or questions as to why the server side value does not match that of the client side value

iot_verifier/src/poc_report.rs
andymck: The alternative is to not use a const and just have a str on the query itself.  I think its better as a const, you would have to go to lengths to misuse this, no?

jeffgrunewald: I’m strictly advocating adding the context to the snippet in case this query ever gets extended in the future to join on any other tables, particularly since the compiler can’t check the use of any of these queries

iot_verifier/src/poc.rs
```diff
@@ -771,19 +777,43 @@ mod tests {
...
-            timestamp: now - Duration::seconds(60 * 12),
+            timestamp: now - *BEACON_INTERVAL,
         });
-        // last beacon was 12 mins in the past, expectation pass
-        assert_eq!(Ok(()), verify_beacon_schedule(&last_beacon_a, now));
-        //we dont have any last beacon data, expectation pass
-        assert_eq!(Ok(()), verify_beacon_schedule(&None, now));
-        // too soon after our last beacon, expectation fail
+        // last beacon was BEACON_INTERVAL in the past, expectation pass
+        assert_eq!(Ok(()), verify_beacon_schedule(&last_beacon, now));
```
maplant, line 785: instead of `assert_eq` here, you might want to do `assert!(verify_beacon_schedule(&last_beacon, now).is_ok())`

iot_verifier/src/poc.rs
jeffgrunewald: i'm thinking more from the perspective of a devnet deployment or an end-to-end integration test where we want to be able to simulate the full pipeline in a manner of 10s of minutes instead of adjusting them for the live mainnet; makes sense to keep them separate from a conceptual standpoint though

jeffgrunewald (approved):

iot_verifier/src/poc.rs
andymck: updated all similar here: f26a924836c10144187fe55e19c28a3a9f1827c0

PR #293, opened 2023-01-11
Remove unused metrics and add gauge for number of beacons in verifier…

bbalser:
… ready to process

### DIFF
--- a/iot_verifier/src/loader.rs
+++ b/iot_verifier/src/loader.rs
@@ -1,6 +1,7 @@
 use crate::{
     gateway_cache::GatewayCache,
     meta::Meta,
+    metrics::LoaderMetricTracker,
     poc_report::{InsertBindings, LoraStatus, Report, ReportType},
     Settings,
 };
@@ -277,6 +278,7 @@ impl Loader {
     ) -> anyhow::Result<()> {
         let file_type = file_info.file_type;
         let tx = Mutex::new(self.pool.begin().await?);
+        let metrics = LoaderMetricTracker::new();
         store
             .stream_file(file_info.clone())
             .await?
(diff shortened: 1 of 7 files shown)


iot_verifier/src/loader.rs
```diff
@@ -372,9 +376,7 @@ impl Loader {
                                         report_type: ReportType::Witness,
                                         status: LoraStatus::Ready,
                                     };
-                                    metrics::increment_counter!(
-                                        "oracles_iot_verifier_loader_witness"
-                                    );
+                                    metrics.increment_witnesses();
```
andymck, line 379: Witnesses are being incremented here but not being decremented afaics

bbalser: That is actually to increment the witness counter, so we never need to decrement.  maybe a better function name will eliminate the confusion?

andymck: Ah that's fine then. Reading on my phone so missed that

andymck (approved):

iot_verifier/src/metrics.rs
```diff
@@ -0,0 +1,81 @@
...
+    pub fn count_loader_witnesses(count: u64) {
+        metrics::counter!(LOADER_WITNESS_COUNTER, count);
+    }
+
+    pub fn count_loader_invalid_witnesses(count: u64, labels: &[(&'static str, &'static str)]) {
+        metrics::counter!(INVALID_WITNESS_COUNTER, count, labels);
+    }
+
+    pub fn num_beacons(count: f64) {
+        metrics::gauge!(BEACON_GUAGE, count);
```
jeffgrunewald, line 24: Maybe hide the f64 requirement for gauges by moving the casting into this function? I think it’s weird enough that a gauge requires a float but now that we’re obscuring the gauge behind our own function here it looks even more arbitrary that youre tracking a gauge with a float

bbalser: Good idea, fixed 857006f

andymck (approved):

PR #297, opened 2023-01-17
Register file sink counters on boot

bbalser:


maplant (approved):
This looks good to me

PR #299, opened 2023-01-19
org and route request auth on config service

jeffgrunewald:
adds basic admin validation of requests to create organization entities; lays foundation for further admin pubkey validation of routing and session key filter management requests.

also consolidates proper deserialization and network validation for submitted owner/payer and delegate keys

implements http2 keepalive interval and timeout for grpc server to prevent long-lived streams from being culled by the load balancer

### DIFF
--- a/iot_config/src/main.rs
+++ b/iot_config/src/main.rs
@@ -8,7 +8,7 @@ use iot_config::{
     gateway_service::GatewayService, org_service::OrgService, route_service::RouteService,
     session_key_service::SessionKeyFilterService, settings::Settings,
 };
-use std::{path::PathBuf, time::Duration};
+use std::{path::PathBuf, sync::Arc, time::Duration};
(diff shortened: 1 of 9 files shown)


iot_config/src/main.rs
```diff
@@ -71,14 +71,18 @@ impl Daemon {
             shutdown_trigger.trigger()
         });
 
+        let admin_pubkey = Arc::new(settings.admin_pubkey()?);
```
andymck, line 74: Is an arc necessary here ?  Why not clone the key itself given it is a relatively small payload

andymck (approved):

iot_config/src/main.rs
jeffgrunewald: removed; changed the strategy here

bbalser (approved):

PR #301, opened 2023-01-19
Change all protobuf references for iot back to lora

bbalser:


andymck (approved):
this looks grand in itself, but not sure i get the reasoning as to why.  lets chat when online

madninja (approved):

PR #302, opened 2023-01-20
add metrics to cover denied and unknown gateways dropped during loading

andymck:
Adds metrics to track the number of reports we drop to the floor due to:

1. Gateway being on the deny list 
2. Gateway not found 
3. Witness report not having any associated beacon data

### DIFF
--- a/iot_verifier/src/loader.rs
+++ b/iot_verifier/src/loader.rs
@@ -54,6 +54,12 @@ pub enum NewLoaderError {
     DenyListError(#[from] denylist::Error),
 }
 
+pub enum ValidGatewayResult {
+    Valid,
+    Denied,
+    Unknown,
+}
+
 impl Loader {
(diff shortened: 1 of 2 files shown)


bbalser (approved):
❤️

jeffgrunewald (approved):

iot_verifier/src/loader.rs
```diff
@@ -54,6 +54,12 @@ pub enum NewLoaderError {
     DenyListError(#[from] denylist::Error),
 }
 
+pub enum ValidGatewayResult {
```
maplant, line 57: It might be more readable to have this called `GatewayValidity`, and then return an `Option<GatewayValidity>` with the None matching unknown

andymck: i gave this a try and tbh i think it is less readable.  Unknown here is an explicit result, ie this gateway has been checked and is not found.  It is not the absence of a result.

PR #305, opened 2023-01-23
Ensure that snr and signal are of correct units

vihu:
For a witness, this change would effectively give back:

```
                                timestamp: 1673546588,
                                signal: -120,
                                packet_hash: [...],
                                signature: [],
                                snr: -5.2,
                                frequency: 905.1,
                                channel: 0,
                                datarate: "SF12BW125",
                                reward_shares: 200,

```

previously:

```
                                timestamp: 1673546588,
                                signal: -1200,
                                packet_hash: [...],
                                signature: [],
                                snr: -52,
                                frequency: 905.1,
                                channel: 0,
                                datarate: "SF12BW125",
                                reward_shares: 200,

```

### DIFF
--- a/poc_iot_injector/src/receipt_txn.rs
+++ b/poc_iot_injector/src/receipt_txn.rs
@@ -14,6 +14,8 @@ use sha2::{Digest, Sha256};
 use std::sync::Arc;
 
 const REWARD_SHARE_MULTIPLIER: Decimal = Decimal::ONE_HUNDRED;
+const SIGNAL_MULTIPLIER: i32 = 10;
+const SNR_MULTIPLIER: f32 = 10.0;
 type PocPath = Vec<BlockchainPocPathElementV1>;
 
 #[derive(Clone, Debug)]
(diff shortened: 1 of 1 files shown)


poc_iot_injector/src/receipt_txn.rs
```diff
@@ -142,10 +144,10 @@ fn construct_poc_witnesses(
             let poc_witness = BlockchainPocWitnessV1 {
                 gateway: witness_report.report.pub_key.into(),
                 timestamp: witness_report.report.timestamp.timestamp() as u64,
-                signal: witness_report.report.signal,
+                signal: witness_report.report.signal / SIGNAL_MULTIPLIER,
```
jeffgrunewald, line 147: i assume this will always cleanly divide into a whole number without underflow or fractional components?

jeffgrunewald (approved):

poc_iot_injector/src/receipt_txn.rs
vihu: I believe it should, from what I could find the incoming signal and snr are both `*10` so this is just dividing that.

vihu: Err, it's i32/i32 but the info after decimal point will be lost, yes. But I think that's okay? So in the witness_report both snr and signal are i32 but for the chain, snr is f32 and signal is i32 (in blockchain_poc_witness_v1) . The more critical bit is that both of these are multiplied by a factor of 10 when they come in, so at the very last step we do the div by 10.

Still, if you guys think we should add some protection around the division, please to be suggesting.

madninja: I'm more worried about the div 10 losing information that should really be available to chain? If the div loss is expected, I'm all good

PR #306, opened 2023-01-23
Refactor LoraValidPocV1 to LoraPocV1

andymck:
Verifier will now output a LoraPocV1 proto containing a list of selected witnesses and a list of unselected witnesses.

The selected witnesses list length will be capped at a max value determined via settings.

Both the selected and unselected lists will potentially contain a mix of valid and invalid witnesses, each represented as a `lora_verified_witness_report_v1` or an 'IotVerifiedWitnessReport` internally.  The status of a witness ( either valid or invalid ) can be determined from the IotVerifiedWitnessReport.status field.


bbalser (approved):
NIT: maybe change `use` statement so everything from `helum_proto` starts with proto, example: `proto::LoraPocV1`.

Makes it clearer which structures are protos

iot_verifier/src/runner.rs
```diff
@@ -458,3 +430,24 @@ fn poc_challengee_reward_unit(num_witnesses: u32) -> anyhow::Result<Decimal> {
     };
     Ok(reward_units.round_dp(SCALING_PRECISION))
 }
+
+/// maybe shuffle & split the verified witness list
+// if the number of witnesses exceeds our cap
+// split the list into two
+// one representing selected witnesses
+// the other representing unselected witnesses
+fn shuffle_and_split_witnesses(
```
vihu, line 439: Can you add a unit test for this please to ensure that the incoming witnesses list is split correctly? There's a `max_witnesses_per_receipt_test` in receipt_txn module in the injector.

andymck: test added here: e103d9b81792652d7f391e99a42b78d7254e19b5

bbalser (approved):

vihu (approved):

PR #309, opened 2023-01-24
Make beacon_interval & beacon_interval_tolerance configurable

vihu:
This removes the constant `BEACON_INTERVA`L and `BEACON_INTERVAL_TOLERANCE` and makes it configurable via settings.

### DIFF
--- a/iot_verifier/src/settings.rs
+++ b/iot_verifier/src/settings.rs
@@ -34,17 +34,17 @@ pub struct Settings {
     pub reward_offset_minutes: i64,
     #[serde(default = "default_max_witnesses_per_poc")]
     pub max_witnesses_per_poc: u64,
-    /// beacon_interval (in seconds)
+    /// The cadence at which hotspots are permitted to beacon (in seconds)
     #[serde(default = "default_beacon_interval")]
     pub beacon_interval: i64,
-    /// beacon_interval_tolerance (in seconds)
+    /// Tolerance applied to beacon intervals within which beacons will be accepted (in seconds)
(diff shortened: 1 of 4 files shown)


iot_verifier/src/settings.rs
```diff
@@ -34,6 +34,22 @@ pub struct Settings {
...
+    #[serde(default = "default_beacon_interval")]
+    pub beacon_interval: i64,
+    /// beacon_interval_tolerance (in seconds)
+    #[serde(default = "default_beacon_interval_tolerance")]
+    pub beacon_interval_tolerance: i64,
+}
+
+// Default: 10 minutes
+pub fn default_beacon_interval_tolerance() -> i64 {
+    10 * 60 * 60
```
bbalser: This default should be `10 * 60`

bbalser (approved):

jeffgrunewald (approved):
lgtm, but I’m becoming a fan of hanging a method off the app’s Settings impl block that takes the raw setting value and returns the converted value you wanna use so you’d call something like `settings.beacon_interval()` and get back a Chrono Duration

vihu:
> lgtm, but I’m becoming a fan of hanging a method off the app’s Settings impl block that takes the raw setting value and returns the converted value you wanna use so you’d call something like `settings.beacon_interval()` and get back a Chrono Duration

Yeah, I like that too. Added [29b4ddf](https://github.com/helium/oracles/pull/309/commits/29b4ddf0b4d05906bc90273c8dde08be3f834d1d)

PR #311, opened 2023-01-25
IoT Config service gateway region params rpc

jeffgrunewald:
Implements the gRPC APIs for IoT gateways to connect and request their asserted location's region and region params, as well as the API for the admin user of the Config service to (re)load region params and region hex lists into the config server, persisting them to the DB

- Loading region params changes and region hex index lists to the service requires the request being signed by the registered "admin" keypair
- [x] final approval and merging of https://github.com/helium/proto/pull/270, followed by updating the oracles cargo file to revert the proto reference back to master branch is a pre-requisite to merging
- Region params and H3idx lists of region hex res 7 indexes are persisted to the DB as bytea fields, in-memory they are wrapped together in a `RegionMap` struct that consists of a `HexTreeMap<Region, EqCompactor>` (for efficient mapping of each asserted location's lorawan region), and a `HashMap<Region, BlockchainRegionParamsV1>`

### DIFF
--- a/iot_config/src/gateway_service.rs
+++ b/iot_config/src/gateway_service.rs
@@ -1,28 +1,45 @@
-use crate::{GrpcResult, Settings};
+use crate::{region_map, GrpcResult, Settings};
 use anyhow::Result;
 use file_store::traits::MsgVerify;
 use helium_crypto::{Keypair, PublicKey, Sign};
 use helium_proto::{
-    services::iot_config::{self, GatewayRegionParamsReqV1, GatewayRegionParamsResV1, LoadRegionReqV1, LoadRegionResV1},
+    services::iot_config::{
+        self, GatewayRegionParamsReqV1, GatewayRegionParamsResV1, LoadRegionReqV1, LoadRegionResV1,
+    },
     Message,
 };
 use node_follower::{
     follower_service::FollowerService,
     gateway_resp::{GatewayInfo, GatewayInfoResolver},
 };
+use sqlx::{Pool, Postgres};
 use tonic::{Request, Response, Status};
 
 pub struct GatewayService {
+    admin_pubkey: PublicKey,
     follower_service: FollowerService,
+    pool: Pool<Postgres>,
     signing_key: Keypair,
 }
 
 impl GatewayService {
-    pub fn new(settings: &Settings) -> Result<Self> {
+    pub async fn new(settings: &Settings) -> Result<Self> {
         Ok(Self {
+            admin_pubkey: settings.admin_pubkey()?,
             follower_service: FollowerService::from_settings(&settings.follower),
+            pool: settings.database.connect(10).await?,
             signing_key: settings.signing_keypair()?,
         })
+    }
+
+    fn verify_admin_signature<R>(&self, request: R) -> Result<R, Status>
+    where
+        R: MsgVerify,
+    {
+        request
+            .verify(&self.admin_pubkey)
+            .map_err(|_| Status::permission_denied("invalid admin signature"))?;
+        Ok(request)
     }
 }
 
@@ -65,7 +82,27 @@ impl iot_config::Gateway for GatewayService {
         Ok(Response::new(resp))
     }
 
-    async fn load_region(&self, _request: Request<LoadRegionReqV1>) -> GrpcResult<LoadRegionResV1> {
-        unimplemented!()
+    async fn load_region(&self, request: Request<LoadRegionReqV1>) -> GrpcResult<LoadRegionResV1> {
+        let request = request.into_inner();
+        let req = self.verify_admin_signature(request)?;
+
+        let params = match req.params {
+            Some(params) => params,
+            None => return Err(Status::invalid_argument("missing region")),
+        };
+
+        region_map::persist_params(req.region, params, &self.pool)
+            .await
+            .map_err(|_| Status::internal("region params save failed"))?;
+
+        if let Some(indexes) = req.hex_indexes {
+            region_map::persist_indexes(req.region, &indexes, &self.pool)
+                .await
+                .map_err(|_| Status::internal("region indexes save failed"))?;
+        } else {
+            tracing::debug!("h3 region index update skipped")
+        };
+
+        Ok(Response::new(LoadRegionResV1 {}))
     }
 }
--- a/iot_config/src/gateway_service.rs
+++ b/iot_config/src/gateway_service.rs
@@ -72,18 +72,20 @@ impl iot_config::Gateway for GatewayService {
             .await
             .map_err(|_| Status::internal("gateway lookup error"))?;
 
-        let location = {
-            let location =
-                location.ok_or_else(|| Status::internal("gateway location undefined"))?;
-            Cell::from_raw(location)
-                .map_err(|_| Status::internal("gateway location is not a valid h3 index"))?
+        let region = match location {
+            Some(location) => {
+                let h3_location = Cell::from_raw(location).map_err(|_| {
+                    Status::internal("gateway asserted location is not a valid h3 index")
+                })?;
+                self.region_map
+                    .get_region(h3_location)
+                    .await
+                    .ok_or_else(|| Status::internal("region lookup failed for assert location"))?
+            }
+            None => Region::from_i32(request.region)
+                .ok_or_else(|| Status::invalid_argument("invalid default region"))?,
         };
 
-        let region = self
-            .region_map
-            .get_region(location)
-            .await
-            .ok_or_else(|| Status::internal("region lookup failed for asserted location"))?;
         let params = self.region_map.get_params(&region).await;
 
         let mut resp = GatewayRegionParamsResV1 {
--- a/iot_config/src/gateway_service.rs
+++ b/iot_config/src/gateway_service.rs
@@ -11,6 +11,7 @@ use helium_proto::{
     },
     Message, Region,
 };
+use hextree::Cell;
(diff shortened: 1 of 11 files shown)


iot_config/src/region_map.rs
```diff
@@ -0,0 +1,149 @@
...
+        let params = BlockchainRegionParamsV1::decode(region_row.params.as_slice())?;
+        params_map.insert(region, params);
+    }
+    Ok(params_map)
+}
+
+pub async fn update_region(
+    region: i32,
+    params: &BlockchainRegionParamsV1,
+    indexes: &[u8],
```
JayKickliter, line 111: Tracing a path to this function, i think it would better show intent to take `Option<&[u8]>` here, instead of doing an `indixes.is_empty()` below.

JayKickliter, line 111: pedantic: ~~indexes~~ indices

iot_config/src/region_map.rs
```diff
@@ -0,0 +1,149 @@
...
+
+    pub async fn get_params(&self, region: &Region) -> Option<BlockchainRegionParamsV1> {
+        self.params_hashmap.read().await.get(region).cloned()
+    }
+
+    pub async fn insert_params(&self, region: Region, params: BlockchainRegionParamsV1) {
+        _ = self.params_hashmap.write().await.insert(region, params)
+    }
+
+    pub async fn swap_tree(&self, new_map: HexTreeMap<Region, EqCompactor>) {
```
JayKickliter, line 43: feels more like `replace_tree`

iot_config/src/region_map.rs
```diff
@@ -0,0 +1,149 @@
...
+    while let Some(region_row) = regions.try_next().await? {
+        let region = Region::from_i32(region_row.region)
+            .ok_or(RegionMapError::UnsupportedRegion(region_row.region))?;
+        let mut idx_decoder = Decoder::new(&region_row.indexes[..])?;
+        let mut buf = Vec::new();
+        idx_decoder.read_to_end(&mut buf)?;
+
+        let mut idx_buf = [0_u8; 8];
+        for chunk in buf.chunks(8) {
+            idx_buf.as_mut_slice().copy_from_slice(chunk);
```
JayKickliter, line 83: Untested pseudoish code
```suggestion
            if chunk.len() < idx_buf.len() {
                log-warning!("malformed serialized region is not a multiple of 8 bytes in length");
                break
            }
            idx_buf.as_mut_slice().copy_from_slice(chunk);
```

JayKickliter: maybe even return an error as something likely went wrong when generating the serialized indices

iot_config/src/region_map.rs
```diff
@@ -0,0 +1,149 @@
...
+
+pub async fn build_region_tree(
+    db: impl sqlx::PgExecutor<'_>,
+) -> Result<HexTreeMap<Region, EqCompactor>, RegionMapError> {
+    let mut region_tree = HexTreeMap::with_compactor(EqCompactor);
+
+    let mut regions = sqlx::query_as::<_, HexRegion>("select * from regions").fetch(db);
+
+    while let Some(region_row) = regions.try_next().await? {
+        let region = Region::from_i32(region_row.region)
```
JayKickliter, line 75: does the `from_i32` imply that the `Region` type is `repor(i32)`? If so, is there any safe way to reduce it to a `u8` to safe memory and improve lookup performance?

iot_config/src/region_map.rs
jeffgrunewald: definitely; good callout. that seems much more semantically correct.

jeffgrunewald: technically it's either or https://www.merriam-webster.com/dictionary/index 🤷

iot_config/src/region_map.rs
jeffgrunewald: good catch; this should be a safe thing to (re)attempt as needed if there are issues with the region files

iot_config/src/region_map.rs
jeffgrunewald: fixed: [3f49af9](https://github.com/helium/oracles/pull/311/commits/3f49af9868222dfe2a512d391e972e1e61c236f8)

iot_config/src/region_map.rs
jeffgrunewald: added: [3f49af9](https://github.com/helium/oracles/pull/311/commits/3f49af9868222dfe2a512d391e972e1e61c236f8)

iot_config/src/region_map.rs
jeffgrunewald: near as i can tell from the `prost_types` that's correct (`repr(i32)`: https://docs.rs/prost-types/latest/prost_types/enum.Syntax.html).
so you're thinking something like pulling the i32 out of the database row, casting to u8 when inserting into the hextreemap and then casting _back_ to a `Region` enum variant when i retrieve it from the hextreemap in order to do the params lookup in the hashmap for that region?

iot_config/src/gateway_service.rs
```diff
@@ -1,17 +1,130 @@
...
-        unimplemented!()
+        let request = request.into_inner();
+
+        let pubkey = PublicKey::try_from(request.address.clone())
+            .map_err(|_| Status::invalid_argument("invalid gateway address"))?;
+        request
+            .verify(&pubkey)
+            .map_err(|_| Status::permission_denied("invalid request signature"))?;
+
+        let GatewayInfo { location, gain, .. } = self
```
andymck, line 68: I assume the use of the follower here in temporary, until a helius db instance is available ?

iot_config/src/gateway_service.rs
```diff
@@ -1,17 +1,130 @@
...
+            .map_err(|_| Status::internal("gateway lookup error"))?;
+
+        let location = {
+            let location =
+                location.ok_or_else(|| Status::internal("gateway location undefined"))?;
+            Cell::from_raw(location)
+                .map_err(|_| Status::internal("gateway location is not a valid h3 index"))?
+        };
+
+        let region = self
```
andymck, line 82: I note the request for this API requires a region to be specified.  In the old sibyl gateway APIs that specified region would be used if the gateway was not asserted.  I dont see that same functionality here, Do we need to add that ? If not then we should remove the region field from the request

iot_config/src/gateway_service.rs
```diff
@@ -1,17 +1,130 @@
...
+    }
+
+    async fn load_region(&self, request: Request<LoadRegionReqV1>) -> GrpcResult<LoadRegionResV1> {
+        let request = request.into_inner();
+        let req = self.verify_admin_signature(request)?;
+
+        let region = Region::from_i32(req.region)
+            .ok_or_else(|| Status::invalid_argument("invalid region"))?;
+
+        let params = match req.params {
```
andymck, line 109: re the match on Some/None, wont the proto itself prevent a none in this instance ?  the proto field is not optional

iot_config/src/gateway_service.rs
madninja: Yeah that needs to be added.

iot_config/src/gateway_service.rs
madninja: yeah this allows gateway-rs to use the config service and remove validators from the equation until switchover time.

iot_config/src/gateway_service.rs
jeffgrunewald: still don't understand exactly how this works but the compiler requires this to be an option even though it's not listed as optional in the proto

iot_config/src/gateway_service.rs
jeffgrunewald: so we're saying in the event the gateway doesn't have an asserted location on the chain based on the pubkey lookup then find and return the region params for the region specified in the request?

iot_config/src/gateway_service.rs
jeffgrunewald: as i recall, it has something to do with the combination of the fact that all values in proto3 are "technically" optional and how proto3 treats default values for non-scalar types, of which region would be one

iot_config/src/gateway_service.rs
andymck: that looks to be the expectation.  In the same  scenario, sibyl would return the specified region, the associated region params and defaulted gain to undefined/none

iot_config/src/region_map.rs
JayKickliter: that's what I was thinking, but just use i32 unless memory use or lookup perf becomes a bottleneck.

iot_config/src/gateway_service.rs
madninja: right.. good recall and a common cause of confusion

iot_config/src/gateway_service.rs
madninja: > so we're saying in the event the gateway doesn't have an asserted location on the chain based on the pubkey lookup then find and return the region params for the region specified in the request?

that's correct.. gateway-rs has a default region but no region params.. With this approach it can get region parameters before it has an asserted location on chain

jeffgrunewald: implemented in [1634b3c](https://github.com/helium/oracles/pull/311/commits/1634b3c7975c298d09acf50afbfcb1b0150052e9)

andymck (approved):

bbalser (approved):

PR #312, opened 2023-01-26
configure loader windows via settings

andymck:
Adds the ability to configure the poc loader window via settings, including the window width, max age and poll time.  The start point of the window can be configured by updating the meta db for keys `report` .  Through a combination of the newly added settings and the modification of the meta keys the loaders can be configured to load historic data, looking as far back in time into the S3 bucket as required.

The driver for this is the upcoming switch to solana, whereby the foundation has indicated it may be required to backfill 24hours worth of reports.  

Also includes a refactor of the entropy loader to use continious source. The poc loader cannot move to continuous source as its loading process is more involved.

Also removes the cloning of the full settings struct into various verifier processes.  Processes now only select the settings they require

### DIFF
--- a/iot_verifier/src/purger.rs
+++ b/iot_verifier/src/purger.rs
@@ -1,4 +1,5 @@
 use crate::{entropy::Entropy, metrics::Metrics, poc_report::Report, Settings};
+use chrono::Duration;
 use file_store::{
     file_sink::{self, FileSinkClient},
     file_upload,
@@ -9,13 +10,13 @@ use file_store::{
     traits::{IngestId, MsgDecode},
     FileType,
 };
+use futures::stream::{self, StreamExt};
 use helium_proto::services::poc_lora::{
     InvalidParticipantSide, InvalidReason, LoraInvalidBeaconReportV1, LoraInvalidWitnessReportV1,
 };
-use std::{ops::DerefMut, path::Path};
-
-use futures::stream::{self, StreamExt};
+use lazy_static::lazy_static;
 use sqlx::{PgPool, Postgres};
+use std::{ops::DerefMut, path::Path};
 use tokio::{
     sync::Mutex,
     time::{self, MissedTickBehavior},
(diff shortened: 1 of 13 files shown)


iot_verifier/src/purger.rs
```diff
@@ -57,12 +56,7 @@ impl Purger {
     pub async fn from_settings(settings: &Settings) -> Result<Self, NewPurgerError> {
         let pool = settings.database.connect(PURGER_DB_POOL_SIZE).await?;
         let settings = settings.clone();
```
madninja, line 58: rather than keeping all of settings I prefer to pull the values you need out of settings. It makes it clear what things you need and avoids cloning unneeded things

iot_verifier/src/settings.rs
```diff
@@ -40,6 +40,41 @@ pub struct Settings {
     /// Tolerance applied to beacon intervals within which beacons will be accepted (in seconds)
     #[serde(default = "default_beacon_interval_tolerance")]
     pub beacon_interval_tolerance: i64,
+    /// window width for the entropy and poc report loader ( in seconds )
```
madninja, line 43: Don't forget to update the settings template toml file with these, including their comments and requirements like for `default_poc_loader_window_max_lookback_age`

madninja (commented):
What makes it "24 hours" as the PR description says? Or is that just outdated and this allows it to be configured  to any value?

andymck:
> What makes it "24 hours" as the PR description says? Or is that just outdated and this allows it to be configured to any value?

It can be configured to any value.  The driver for this was that foundation wants to look back 24 hours.  I can remove that from the description if it makes it cleare

iot_verifier/src/purger.rs
andymck: bah, that was how i had been doing it and then noticed this pattern being used elsewhere.

madninja:
> > What makes it "24 hours" as the PR description says? Or is that just outdated and this allows it to be configured to any value?
> 
> It can be configured to any value. The driver for this was that foundation wants to look back 24 hours. I can remove that from the description if it makes it cleare

Yeah that'd help.. I was looking for "24 hours" somewhere

iot_verifier/src/purger.rs
madninja: yeah, no need to follow bad patterns :-)

andymck: touche

jeffgrunewald (commented):
i think any time we have configurable settings related to timing where we take a settings value and then multiply it by something else in the application code we're setting ourselves up for confusion by distributing out where all of the actual runtime timings are set. if we want to take round numbers like 60 and multiply them by some factor like 5 i think that's fine but it should be done all in one place in the settings, either explicitly in the accessor method we write to use that setting value or make it all one value and document the `60 * 5` in a comment

andymck:
@jeffgrunewald I am not entirely sure I am getting your meaning here.   what exactly are you proposing ?  Is it the specific use of those `60 * 5` instances or were you using those are as example ?

The 60 * 5 instances in this PR are not used in conjunction with any settings specifically/. Instead they are taking an existing calculate window width and widening it by a factor.

jeffgrunewald (approved):

bbalser (approved):

PR #313, opened 2023-01-26
handle enable and disable org rpcs

jeffgrunewald:
Implement the config service's `enable` and `disable` org RPCs to allow the packet verifier to disable orgs and have their routes cascade-removed from the downstream helium packet routers when they run out of DC to pay for routing and allow re-enabling the orgs to cascade-recreate their routes when funds are replenished.

This change necessitates cloning the router service's ~~Arc-wrapped~~ broadcast channel transmitter to allow it to broadcast to subscribed packet routers when an org's status changes, thereby deleting or creating the routes registered to those orgs from the routing infrastructure

### DIFF
--- a/iot_config/src/org.rs
+++ b/iot_config/src/org.rs
@@ -12,7 +12,7 @@ pub mod proto {
     pub use helium_proto::services::iot_config::{OrgResV1, OrgV1};
 }
 
-#[derive(Clone, Debug, Serialize, sqlx::Type)]
+#[derive(Clone, Debug, PartialEq, Serialize, sqlx::Type)]
 #[sqlx(type_name = "org_status", rename_all = "snake_case")]
 pub enum OrgStatus {
     Enabled,
@@ -140,6 +140,17 @@ pub async fn get_with_constraints(
             end_addr: end_addr.into(),
         },
     })
+}
+
+pub async fn get_status(oui: u64, db: impl sqlx::PgExecutor<'_>) -> Result<OrgStatus, sqlx::Error> {
+    sqlx::query_scalar::<_, OrgStatus>(
+        r#"
+        select status from organizations where oui = $1
+        "#,
+    )
+    .bind(oui as i64)
+    .fetch_one(db)
+    .await
 }
 
 pub async fn toggle_status(
--- a/iot_config/src/org.rs
+++ b/iot_config/src/org.rs
@@ -142,6 +142,26 @@ pub async fn get_with_constraints(
     })
 }
 
+pub async fn toggle_status(
+    oui: u64,
+    status: OrgStatus,
+    db: impl sqlx::PgExecutor<'_>,
+) -> Result<(), sqlx::Error> {
+    sqlx::query(
+        r#"
+        update organizations
+        set status = $1
+        where oui = $2
+        "#,
+    )
+    .bind(status)
+    .bind(oui as i64)
+    .execute(db)
+    .await?;
+
+    Ok(())
+}
+
 #[derive(thiserror::Error, Debug)]
 pub enum OrgPubkeysError {
     #[error("error retrieving saved org keys: {0}")]
--- a/iot_config/src/org_service.rs
+++ b/iot_config/src/org_service.rs
@@ -181,30 +181,36 @@ impl iot_config::Org for OrgService {
 
         let req = self.verify_admin_signature(request)?;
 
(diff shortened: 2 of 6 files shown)


iot_config/src/org_service.rs
```diff
@@ -1,26 +1,39 @@
...
 use sqlx::{Pool, Postgres};
+use std::sync::Arc;
+use tokio::sync::broadcast::Sender;
 use tonic::{Request, Response, Status};
 
 pub struct OrgService {
     admin_pubkey: PublicKey,
     pool: Pool<Postgres>,
     required_network: Network,
+    route_update_tx: Arc<Sender<RouteStreamResV1>>,
```
madninja, line 24: Why is the arc needed at all? a broadcast:Sender is Clone right?

jeffgrunewald: It is Clone, but I took this paragraph to mean it wasn’t safe to clone across threads: https://docs.rs/tokio/latest/tokio/sync/broadcast/fn.channel.html

jeffgrunewald: Since I need to share the sender with _from_ the router service where it’s created and the route stream lives

madninja: it's safe to be clone'd within the same process and even across processes but then you need an Arc to pass it across a process boundary for lifetime reasons..  which means you really only need to have an Arc if the compiler complains that it can't figure out the lifetime. 

Did you try it without?

jeffgrunewald: nope, but it appears to work just fine; updated here [65c1533](https://github.com/helium/oracles/pull/313/commits/65c15331459a8788d93e5668ed359da8f74fa611)

iot_config/src/org.rs
```diff
@@ -142,6 +142,26 @@ pub async fn get_with_constraints(
     })
 }
 
+pub async fn toggle_status(
```
maplant, line 145: Nit: Maybe call this set status? Toggle implies turning on or off based on the previous state

iot_config/src/org_service.rs
```diff
@@ -158,11 +170,83 @@ impl iot_config::Org for OrgService {
...
+                .map_err(|_| Status::internal(format!("org disable failed for: {}", req.oui)))?;
+
+            let org_routes = list_routes(req.oui, &self.pool).await.map_err(|_| {
+                Status::internal(format!(
+                    "error retrieving routes for disabled org: {}",
+                    req.oui
+                ))
+            })?;
+
+            for route in org_routes {
```
andymck, line 194: Does it matter if the org is successfully disabled but one or more deletes of the routes fail ?  what are the implications of that ?
Same question applies to the org enable flow

iot_config/src/route.rs
```diff
@@ -7,7 +7,7 @@ use helium_proto::Region;
 use serde::{Deserialize, Serialize};
 use serde_json::json;
 use sqlx::{types::Uuid, Row};
-use std::{collections::BTreeMap, sync::Arc};
```
andymck, line 10: should update the PR description now that the arc is dropped

andymck (approved):
a couple of comments but lgtm

iot_config/src/route.rs
jeffgrunewald: Good catch

iot_config/src/org_service.rs
michaeldjeffrey: My understanding is that `send` fails when there are no receivers. 
If there are no receivers, we probably don't need to continue trying to send updates. 
But we should not error out to the caller.

iot_config/src/org_service.rs
```diff
@@ -158,11 +170,83 @@ impl iot_config::Org for OrgService {
...
+            let org_routes = list_routes(req.oui, &self.pool).await.map_err(|_| {
+                Status::internal(format!(
+                    "error retrieving routes for enabled org: {}",
+                    req.oui
+                ))
+            })?;
+
+            for route in org_routes {
+                let route_id = route.id.clone();
+                self.route_update_tx
```
michaeldjeffrey, line 236: Enabling an Orgs Routes also needs to send the Devaddrs and Euis updates.

iot_config/src/org_service.rs
andymck: sgtm,  I read the docs but wasnt entirely sure offline queuing or delivery guarantees were supported

bbalser (approved):

PR #317, opened 2023-01-30
verify min cell distance between beaconer and witness

andymck:
depends on https://github.com/helium/proto/pull/274

Adds a missed verification to determine the witness is located a minimum number of cells from the beaconer.  If not the witness will be declared invalid

The test data was run via the blockchain core code using its h3 implementation in order to sanitise the input and results.

### DIFF
--- a/iot_verifier/src/poc.rs
+++ b/iot_verifier/src/poc.rs
@@ -30,6 +30,10 @@ pub const R: f64 = 6.371e6;
 
 /// max permitted distance of a witness from a beaconer measured in KM
 const POC_DISTANCE_LIMIT: i32 = 100;
+/// the minimum distance in cells between a beaconer and witness
+const POC_CELL_DISTANCE_MINIMUM: u32 = 8;
+/// the resolution at which parent cell distance is derived
+const POC_CELL_PARENT_RES: u8 = 11;
 
 pub struct Poc {
     beacon_report: IotBeaconIngestReport,
@@ -286,6 +290,7 @@ impl Poc {
             witness_report.report.frequency,
         )?;
         verify_witness_region(beaconer_info.region, witness_info.region)?;
+        verify_witness_cell_distance(beaconer_info.location, witness_info.location)?;
         verify_witness_distance(beaconer_info.location, witness_info.location)?;
         verify_witness_rssi(
             witness_report.report.signal,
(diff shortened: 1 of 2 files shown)


vihu:
Just a general thing I'd suggest is that the first check should be something like: `verify_witness_and_beacon_loc`; if that fails, you exit (fail the verification step) and don't do anything. That can return the location of beacon and witness,  the subsequent remaining checks will then no longer need to have `Option<u64>` as the incoming location(s).

iot_verifier/src/poc.rs
```diff
@@ -743,6 +782,20 @@ mod tests {
...
+    #[test]
+    fn test_calc_cell_distance() {
+        // location 1 is 51.51231394840223, -0.2919014665284206 ( ealing, london)
+        // location 2 is 51.5649315958581, -0.10295780727096393 ( finsbury pk, london)
+        // google maps distance is ~14.32km
+        // converted co-ords to h3 index at resolution 15
+        let loc1 = 644459695463521437;
+        let loc2 = 644460986971331488;
+        let dist: i32 = calc_cell_distance(loc1, loc2).unwrap() as i32;
+        // verify the calculated cell distance is more than min cell distance
```
JayKickliter, line 794: we're missing a verification step. Is this what you meant?

```suggestion
        verify_witness_cell_distance(Some(loc1), Some(loc2)).unwrap();
```

JayKickliter, line 794: Is this what you meant by the comment?
```suggestion
        // verify the calculated cell distance is more than min cell distance
        verify_witness_cell_distance(Some(loc1), Some(loc2)).unwrap();
```

iot_verifier/src/poc.rs
```diff
@@ -743,6 +782,20 @@ mod tests {
         assert!(100 > (gmaps_dist.abs_diff(dist) as i32));
     }
 
+    #[test]
+    fn test_calc_cell_distance() {
```
JayKickliter, line 786: We probably want at least one negative test (two cells that are less than the min).

iot_verifier/src/poc.rs
```diff
@@ -508,6 +513,31 @@ fn verify_witness_distance(
...
+        Ok(d) => d,
+        Err(_) => return Err(InvalidReason::BelowMinDistance),
+    };
+    if cell_distance < POC_CELL_DISTANCE_MINIMUM {
+        tracing::debug!(
+            "witness verification failed, reason: {:?}. cell distance {cell_distance}",
+            InvalidReason::BelowMinDistance
+        );
+        return Err(InvalidReason::BelowMinDistance);
+    }
```
jeffgrunewald, line 537:
```suggestion
    if let (Some(beacon_loc), Some(witness_loc)) = (beacon_loc, witness_loc) {
        calc_cell_distance(beacon_loc, witness_loc).map_err(|_| InvalidReason::BelowMinDistance)?;
        return Ok(())
    }
    tracing::debug!("witness verification failed, reason: {:?}. cell distance {cell_distance}", InvalidReason::BelowMinDistance);
    Err(InvalidReason::BelowMinDistance)
```

jeffgrunewald: untested but could simplify the double-option situation you're looking to validate where both locations are "Some"

iot_verifier/src/poc.rs
andymck: No, this test asserts the helper function calc_cell_distance only.
i actually need to add a sep test for the verification itself, which i totally forgot

iot_verifier/src/poc.rs
andymck: The negative test will be part of the test covering the verification itself, as per the other verifications.  Will add shortly

iot_verifier/src/poc.rs
andymck: i think i prefer the existing patten, more readable but will give it a go locally

andymck:
> Just a general thing I'd suggest is that the first check should be something like: `verify_witness_and_beacon_loc`; if that fails, you exit (fail the verification step) and don't do anything. That can return the location of beacon and witness, the subsequent remaining checks will then no longer need to have `Option<u64>` as the incoming location(s).

location is currently verified prior to the cell distance and indeed before most other checks.  However the verifications are purely functional atm and do not return any data...with each test passed in the raw data sets.   I do take your point however and can likely prep the location data at the outset of the tests.  But i will address that in a sep PR

iot_verifier/src/poc.rs
andymck: verification test added here: 0c266f8131af8ba7ca0a1b955603a166f8215314

iot_verifier/src/poc.rs
andymck: thanks for suggesting but i will keep as is for time being.

JayKickliter (approved):

PR #319, opened 2023-01-30
[iot-config] upgrade hextree

JayKickliter:
This PR:

- upgrades the hextree to 0.2.0
- moves a validity check out of a hot loop
- replaces using of `h3ron::H3Cell` with `hextree::Cell`

Not that this PR targets @jeffgrunewald's branch, not main.

### DIFF
--- a/iot_config/src/gateway_service.rs
+++ b/iot_config/src/gateway_service.rs
@@ -11,6 +11,7 @@ use helium_proto::{
     },
     Message, Region,
 };
+use hextree::Cell;
 use node_follower::{
     follower_service::FollowerService,
     gateway_resp::{GatewayInfo, GatewayInfoResolver},
@@ -71,7 +72,12 @@ impl iot_config::Gateway for GatewayService {
             .await
             .map_err(|_| Status::internal("gateway lookup error"))?;
 
-        let location = location.ok_or_else(|| Status::internal("gateway location undefined"))?;
+        let location = {
+            let location =
+                location.ok_or_else(|| Status::internal("gateway location undefined"))?;
+            Cell::from_raw(location)
+                .map_err(|_| Status::permission_denied("invalid h3 index"))?
+        };
 
         let region = self
             .region_map
--- a/iot_config/src/gateway_service.rs
+++ b/iot_config/src/gateway_service.rs
@@ -76,7 +76,7 @@ impl iot_config::Gateway for GatewayService {
             let location =
                 location.ok_or_else(|| Status::internal("gateway location undefined"))?;
             Cell::from_raw(location)
-                .map_err(|_| Status::permission_denied("invalid h3 index"))?
+                .map_err(|_| Status::internal("gateway location is not a valid h3 index"))?
         };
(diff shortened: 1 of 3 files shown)


iot_config/src/gateway_service.rs
```diff
@@ -71,7 +72,12 @@ impl iot_config::Gateway for GatewayService {
             .await
             .map_err(|_| Status::internal("gateway lookup error"))?;
 
-        let location = location.ok_or_else(|| Status::internal("gateway location undefined"))?;
+        let location = {
+            let location =
+                location.ok_or_else(|| Status::internal("gateway location undefined"))?;
+            Cell::from_raw(location)
```
JayKickliter, line 78: I felt like adding the is-valid-cell check to this module made sense, thus keeping invalid indices out of the `get_region` call.

iot_config/src/region_map.rs
```diff
@@ -76,19 +68,28 @@ pub async fn build_region_tree(
     while let Some(region_row) = regions.try_next().await? {
         let region = Region::from_i32(region_row.region)
             .ok_or(RegionMapError::UnsupportedRegion(region_row.region))?;
-        let mut idx_decoder = Decoder::new(&region_row.indexes[..])?;
-        let mut buf = Vec::new();
-        idx_decoder.read_to_end(&mut buf)?;
-
-        let mut idx_buf = [0_u8; 8];
-        for chunk in buf.chunks(8) {
-            if chunk.len() < idx_buf.len() {
```
JayKickliter, line 85: note that I moved this check out of the hot-loop by doing a single is-multiple-of-8 check before

iot_config/src/gateway_service.rs
```diff
@@ -71,7 +72,12 @@ impl iot_config::Gateway for GatewayService {
...
+                .map_err(|_| Status::permission_denied("invalid h3 index"))?
```
jeffgrunewald, line 79: not a fan of the fairly limited offering tonic `Status` options we have to provide feedback to the user; i have been leaning on `internal` errors whenever it's caused by a problem on "our" end which i would argue we have here since the internal mechanics of the oracles/chain are trying to retrieve the asserted location. in as far as anyone is looking at their gateway logs this feels like it's not actionable by the operator since they have no direct control over what h3 index the gateway is asserted against to correct this but if anything i'd argue it's at least an `invalid_argument` error. thought?

jeffgrunewald (approved):

iot_config/src/gateway_service.rs
JayKickliter: I hadn't formed any mental image for this and your rationale seems sound. I didn't think about the fact that the user can't really do anything about the extremely unlikely chance their hotspot has an invalid index.

JayKickliter:
@jeffgrunewald I am done pushing to this unless you have further feedback. Since this targets your PR, you may want to choose when to merge and what type of merge to do.

jeffgrunewald (approved):

PR #322, opened 2023-01-31
tidy up use of location in verifications

andymck:
Tidies up use of location values in verifications.  The only verification which handles an Option is that which verifies the gateway is asserted.  Thereafter the remaining verification assume we are able to derive an asserted location value and work of the u64

### DIFF
--- a/iot_verifier/src/poc.rs
+++ b/iot_verifier/src/poc.rs
@@ -242,6 +242,7 @@ impl Poc {
             "verifying beacon from beaconer: {:?}",
             PublicKeyBinary::from(beaconer_info.address.clone())
         );
+        verify_gw_location(beaconer_info.location)?;
         let beacon_received_ts = self.beacon_report.received_timestamp;
         verify_entropy(self.entropy_start, self.entropy_end, beacon_received_ts)?;
         verify_beacon_schedule(
@@ -250,7 +251,6 @@ impl Poc {
             beacon_interval,
             beacon_interval_tolerance,
         )?;
-        verify_gw_location(beaconer_info.location)?;
         verify_gw_capability(beaconer_info.staking_mode)?;
         tracing::debug!(
             "valid beacon from beaconer: {:?}",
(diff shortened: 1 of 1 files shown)


iot_verifier/src/poc.rs
```diff
@@ -269,7 +269,13 @@ impl Poc {
             "verifying witness from gateway: {:?}",
             PublicKeyBinary::from(witness_info.address.clone())
         );
+        verify_gw_location(witness_info.location)?;
         let beacon_report = &self.beacon_report;
+        // at this point the location has been verified for both beaconer and witness
+        // we need to use the location values for both in subsequent verifications
+        // derive the asserted value whilst avoiding unwraps
+        let beaconer_location = beaconer_info.location.ok_or(InvalidReason::NotAsserted)?;
+        let witness_location = witness_info.location.ok_or(InvalidReason::NotAsserted)?;
```
jeffgrunewald, line 278: this just feels like doing the same validation a second time when you could modify the `verify_gw_location()` function to return the location in the success case and just have that be one instance where the verifying function doesn't return the `GenericVerifyResult` or else make `GenericVerifyResult` an enum like a regular `Result` type where it can return multiple values in the success case but the default is `GenericVerifyResult<()>`

maplant: Here's what you should do. You should replace `verify_gw_location` with the following function:
```rust
/// verify gateway has an asserted location
fn verify_gw_location(gateway_loc: Option<u64>) -> GenericVerifyResult<u64> {
    match gateway_loc {
        Some(location) => Ok(location),
        None => {
            tracing::debug!(
                "beacon verification failed, reason: {:?}",
                InvalidReason::NotAsserted
            );
            Err(InvalidReason::NotAsserted)
        }
    }
}
```

Then, replace your unwraps with the following: 
```rust
let beaconer_location = verify_gw_location(beaconer_info.location)?;
let witness_location = verify_gw_location(witness_info.location)?;
```

PR #326, opened 2023-02-02
Allow db url to be optional

madninja:
When not specified the database connection settings are pulled from the following environment variables. Note that using this approach removes the ability to have multiple db connections sections in the same settings file.

* `PGHOST`
* `PGPORT`
* `PGUSER`
* `PGPASSWORD`
* `PGDATABASE`
* `PGSSLROOTCERT`
* `PGSSLMODE`
* `PGAPPNAME`

### DIFF
--- a/db_store/src/settings.rs
+++ b/db_store/src/settings.rs
@@ -1,7 +1,9 @@
 use crate::Result;
-use http::Uri;
 use serde::Deserialize;
-use sqlx::{postgres::PgPoolOptions, Pool, Postgres};
+use sqlx::{
+    postgres::{PgConnectOptions, PgPoolOptions},
+    Pool, Postgres,
+};
 
 #[derive(Debug, Deserialize, Clone)]
 pub struct Settings {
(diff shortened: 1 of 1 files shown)


andymck (approved):

jeffgrunewald (commented):
if this breaks the ability to have multiple db connection settings won't this make it impossible to connect an app to its own db instance _and_ the helius db instance once that's swapped in in place of the blockchain-node connection?

bbalser (approved):

bbalser:
> if this breaks the ability to have multiple db connection settings won't this make it impossible to connect an app to its own db instance _and_ the helius db instance once that's swapped in in place of the blockchain-node connection?

I believe the restriction is we can only have 1 db connection controlled via the environment variables, you can still set a url in the settings file and environment variables will not be read. 

So we can have one db connection to our db configured via the settings file and one to helius configured via environment variables.

jeffgrunewald:
> So we can have one db connection to our db configured via the settings file and one to helius configured via environment variables.

if i understand the reason for this change in the first place it's that the environment variables were needed to connect to the oracle's db instance through added security layers and we're not currently even trying to connect to helius, so once that's in play we'd need to be able to make two separate connections via environment variable configuration. is that not correct?

madninja:
None of this is related to the "helium" database. Some db connections require data that can't be represented in a URL (looking at you, AWS generated password tokens). 

@JayKickliter you are correct.. env variable databases are restricted to one per settings file. You can have as many url based ones as needed.

PR #327, opened 2023-02-02
post self beaconing live deployment additions

andymck:
1.  dont include any self witness in the final 14 witness list, drop em to the the floor
2. refactor calc_fspl, account for beacon gain

### DIFF
--- a/iot_verifier/src/poc.rs
+++ b/iot_verifier/src/poc.rs
@@ -29,7 +29,7 @@ pub const C: f64 = 2.998e8;
 pub const R: f64 = 6.371e6;
 
 /// max permitted distance of a witness from a beaconer measured in KM
-const POC_DISTANCE_LIMIT: i32 = 100;
+const POC_DISTANCE_LIMIT: u32 = 100;
 /// the minimum distance in cells between a beaconer and witness
 const POC_CELL_DISTANCE_MINIMUM: u32 = 8;
 /// the resolution at which parent cell distance is derived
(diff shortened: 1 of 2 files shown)


andymck:
this version will drop the self witnessing in such a way they will not show up in the selected or unselected lists.  It is likely preferable to get the self witnesses into the unselected lists.  I am still trying to work out an efficient way to do that ( ie no quadratic within a for loop ).  

There is the drain_filter method but that is experimental ?  @maplant @madninja any ideas here and thoughts on using experimental   APIs ?

madninja:
> this version will drop the self witnessing in such a way they will not show up in the selected or unselected lists. It is likely preferable to get the self witnesses into the unselected lists. I am still trying to work out an efficient way to do that ( ie no quadratic within a for loop ).
> 
> There is the drain_filter method but that is experimental ? @maplant @madninja any ideas here and thoughts on using experimental APIs ?

Not a fan of experimental apis for production code

andymck:
I think i am overthinking this.  Unless someone objects, I am going to stick with filtering the self witnesses out.  They wont make it to the selected or unselected lists.  This replicates the erlang behaviour where they will simply dropped to the floor

madninja:
> I think i am overthinking this. Unless someone objects, I am going to stick with filtering the self witnesses out. They wont make it to the selected or unselected lists. This replicates the erlang behaviour where they are simply dropped to the floor

Fine.. I was hoping to be able to track them for any gaming angles and improving the client side.

PR #331, opened 2023-02-03
Drop self witnesses, monitor cache for evictions

andymck:


bbalser (approved):

iot_verifier/src/loader.rs
```diff
@@ -105,6 +105,7 @@ impl Loader {
             }
         }
         tracing::info!("stopping verifier loader");
+        gateway_cache.cache_monitor.abort();
```
jeffgrunewald, line 108: this is probably a question for a tokio-expert like @madninja or @maplant but is there any disadvantage to cleaning up this spawned thread here when the cache/verifier is going through a shutdown sequence vs. having the GatewayCache `from_settings` method instead return a tuple (gateway_cache_instance, monitor_join_handle) and then passing the cache instance around like we're already doing but have the cache monitor thread managed by the main `try_join!`?

andymck: I couldnt find a clear home for this, settled for here and prob not ideal. open to suggestions

jeffgrunewald: my only thought would be to kick it up to the same "supervising" thread that instantiated the cache but again, i can't say if that's better or not. would be interested to hear thoughts after-the-fact even if this is functionally good to go and solves the problem

jeffgrunewald (approved):

PR #336, opened 2023-02-04
base reward units on Beaconer TS, and only on valid selected witnesses

andymck:


iot_verifier/src/runner.rs
```diff
@@ -269,12 +264,37 @@ impl Runner {
...
+                        &beacon_id,
+                        &mut selected_witnesses,
+                        max_witnesses_per_poc,
                     )?;
 
-                    verified_witnesses_result.update_reward_units(witness_reward_units);
+                    // get the number of valid witnesses in our selected list
+                    let num_valid_selected_witnesses = selected_witnesses
+                        .clone()
+                        .into_iter()
```
jeffgrunewald, line 284: should be able to do this without clone/into_iter; i would think counting the results wouldn't need to consume the iterator and counting on the references in the `iter()` version would work

jeffgrunewald (approved):
for @andymck 's part fixing the reward unit calculations based on valid _and_ selected witness counts only, looks good to me; would probably be reasonable to get three approvals to cover the spread since i contributed the hex_density_scaling lookup for witness results fix

madninja (approved):

bbalser (approved):

PR #340, opened 2023-02-06
Fix HIP15 TX reward unit formula

mawdegroot:


jeffgrunewald (approved):

bbalser (approved):

andymck (approved):

vihu (approved):

iot_verifier/src/runner.rs
```diff
@@ -444,7 +444,7 @@ fn poc_challengee_reward_unit(num_witnesses: u32) -> anyhow::Result<Decimal> {
     let reward_units = if num_witnesses == 0 {
         Decimal::ZERO
     } else if num_witnesses < WITNESS_REDUNDANCY {
-        Decimal::from(WITNESS_REDUNDANCY / num_witnesses)
+        Decimal::from(num_witnesses / WITNESS_REDUNDANCY)
```
jeffgrunewald, line 447: turns out this will not inherently work as-is b/c the integer division is going to result in this being `Decimal::from(0)`

jeffgrunewald (changes requested):

iot_verifier/src/runner.rs
mawdegroot: Close this, roll it into the big patch and I'll review that one?

PR #342, opened 2023-02-07
log when we see valid witness with 0 reward shares

andymck:


poc_iot_injector/src/receipt_txn.rs
```diff
@@ -135,11 +135,23 @@ fn construct_poc_witnesses(
...
+        let witness_invalid_reason = witness_report.invalid_reason as i32;
         let reward_shares = ((witness_report.hex_scale * witness_report.reward_unit)
             * REWARD_SHARE_MULTIPLIER)
             .to_u32()
             .unwrap_or_default();
 
+        if reward_shares == 0 && witness_invalid_reason == 0 {
+            tracing::warn!(
+                "valid witness with zero reward shares.
+                pubkey: {:?},
```
madninja, line 147: none of these need the debug version of the format specifier right? Just `{}` would work ?

andymck: fixed

madninja (approved):

bbalser (approved):

PR #344, opened 2023-02-07
Update Readme of each service to include S3 Inputs and Outputs

bbalser:


ingest/README.md
```diff
@@ -0,0 +1,20 @@
+# Ingest
+
+## IOT
+
```
madninja, line 4: document the ingest grpc input  protos if you could

bbalser: Added c9e101c

madninja (approved):

PR #345, opened 2023-02-07
iot config session key service

michaeldjeffrey:
- removed unique constraint on owner_pubkey in the organizations table.
- If an org is deleted, session_key remove updates will not be sent out. I'm not sure if this is a terribly huge problem, as they're removed from the DB, and will not be resent on a fresh connection.


iot_config/src/session_key.rs
```diff
@@ -0,0 +1,184 @@
...
+pub async fn list_stream<'a>(
+    db: impl sqlx::PgExecutor<'a> + 'a,
+) -> impl Stream<Item = SessionKeyFilter> + 'a {
+    sqlx::query_as::<_, SessionKeyFilter>("select * from session_key_filters")
+        .fetch(db)
+        .filter_map(|filter| async move { filter.ok() })
+}
+
+pub async fn list_for_oui(
+    oui: u64,
```
andymck, line 35: as you need the oui as an i64 maybe enforce that here in the function sig and have the callers perform any cast.  Likewise for the other instance of same

iot_config/src/session_key_service.rs
```diff
@@ -1,43 +1,175 @@
...
+    pool: Pool<Postgres>,
+    update_channel: Sender<SessionKeyFilterStreamResV1>,
+}
+
+impl SessionKeyFilterService {
+    pub async fn new(settings: &Settings) -> Result<Self> {
+        let (update_tx, _) = tokio::sync::broadcast::channel(128);
+        Ok(Self {
+            admin_pubkey: settings.admin_pubkey()?,
+            pool: settings.database.connect(10).await?,
```
andymck, line 32: nit: add a const for the connection pool size

iot_config/src/session_key_service.rs
```diff
@@ -1,43 +1,175 @@
...
+        request: Request<SessionKeyFilterListReqV1>,
     ) -> GrpcResult<Self::listStream> {
-        unimplemented!();
+        let request = request.into_inner();
+
+        let filters = session_key::list_for_oui(request.oui, &self.pool)
+            .await
+            .map_err(|_| Status::internal("list session key filters failed"))?;
+
+        let (tx, rx) = tokio::sync::mpsc::channel(20);
```
andymck, line 69: nit: add a const for the buffer size, used in multiple places so be simpler to bump size

andymck (approved):

iot_config/src/session_key.rs
jeffgrunewald: my goal is to wrap all of these "db wrapper" functions that wrap sqlx queries in a trait that can be tested with non-postgres impls for testing, in which case i'd keep this as-is since the i64 is a constraint of postgres only and the rust and proto definitions of these types all expect an unsigned integer

iot_config/src/session_key_service.rs
jeffgrunewald: @bbalser is making changes to support postgres password vs. iam role-based authentication and i believe his change is going to include this type of centralized pool configuration

bbalser (approved):

PR #348, opened 2023-02-08
Fix queries for setting and getting region index files

jeffgrunewald:
Also update helium-crypto to update solana-sdk and address a Verify trait bug when the message in question isn't the expected length

### DIFF
--- a/iot_config/migrations/4_regions.sql
+++ b/iot_config/migrations/4_regions.sql
@@ -1,5 +1,5 @@
 create table regions (
-    region int primary key not null,
+    region text primary key not null,
     params bytea not null,
     indexes bytea,
 
--- a/iot_config/src/region_map.rs
+++ b/iot_config/src/region_map.rs
@@ -122,29 +122,23 @@ pub async fn update_region(
 
     sqlx::query(
         r#"
-        insert into regions (region, params)
-        values ($1, $2)
-        on conflict (region) do update set params = excluded.params
+        insert into regions (region, params, indexes)
+        values ($1, $2, $3)
+        on conflict (region) do update set
+            params = excluded.params,
+            indexes = case when excluded.indexes is not null
+                          then excluded.indexes
+                          else regions.indexes
+                      end
         "#,
     )
     .bind(region.to_string())
     .bind(params.encode_to_vec())
+    .bind(indexes)
     .execute(&mut transaction)
     .await?;
 
-    let updated_region = if let Some(indexes) = indexes {
-        sqlx::query(
-            r#"
-            update regions
-            set indexes = $1
-            where region = $2
-            "#,
-        )
-        .bind(indexes)
-        .bind(region.to_string())
-        .execute(&mut transaction)
-        .await?;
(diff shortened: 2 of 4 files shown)


iot_config/src/region_map.rs
```diff
@@ -65,10 +65,14 @@ pub async fn build_region_tree(
...
 
-    while let Some(region_row) = regions.try_next().await? {
-        let region = Region::from_i32(region_row.region)
-            .ok_or(RegionMapError::UnsupportedRegion(region_row.region))?;
-        let mut h3_idx_decoder = Decoder::new(&region_row.indexes[..])?;
+    while let Some(HexRegion {
+        region,
+        indexes: Some(indexes),
+        ..
+    }) = regions.try_next().await?
```
michaeldjeffrey, line 72: This will stop processing the first row it encounters that doesn't have indexes.
Is there any chance a row without indexes will be sandwiched between others that do?

jeffgrunewald: Darn, you’re right. Please hold

jeffgrunewald: fixed [df46ecb](https://github.com/helium/oracles/pull/348/commits/df46ecbd7189f68169b6e84f79ea577815f397de)

iot_config/src/region_map.rs
```diff
@@ -127,21 +127,21 @@ pub async fn update_region(
...
     .execute(&mut transaction)
     .await?;
 
     let updated_region = if let Some(indexes) = indexes {
         sqlx::query(
             r#"
-            insert into regions (region, indexes)
-            values ($1, $2)
-            on conflict (region) do update set indexes = excluded.indexes
+            update regions
```
andymck, line 138: is it not possible to include the indexes in the prior upsert ?

andymck (approved):

iot_config/src/region_map.rs
jeffgrunewald: i wasn't optimistic this was possible since the indexes are a nullable field but as it turns out this _is_ possible thanks to some conditional logic in the query and the fact that sqlx will automatically accept type `T` or `Option<T>` interchangeably when dealing with a nullable field (TIL!); thanks for the push

jeffgrunewald: [8a48ab7](https://github.com/helium/oracles/pull/348/commits/8a48ab7d9556143bf60d1255a544ff0e754185bb)

iot_config/migrations/4_regions.sql
```diff
@@ -1,5 +1,5 @@
 create table regions (
-    region int primary key not null,
+    region text primary key not null,
```
vihu, line 2: Does this require a new migration with an update table?

jeffgrunewald: nope; already reset the db. @madninja wants to continue to collapse these migrations until we're out of the "prototyping" phase

michaeldjeffrey (approved):

PR #349, opened 2023-02-09
write out metrics for invalid witnesses directly from runner

andymck:
When we switched to the iot verifier performing the final 14 witness selection we broke the metrics for invalid witnesses.

Prior to this change, invalid witnesses were all written directly to S3 via their own invalid_witness proto.  The write function of the filestore took care of firing a metric for each file written to the sink.  

Post the above change however invalid witnesses form part of the `poc` proto, where invalid and valids make up the `poc` proto's `selected` and `unselected` lists.  This means the invalids now bypass the existing filestore metric generation.

This PR fixes the issue by firing the same metric as previously but from outside of the filestore and from the runner directly.  The filestore metric remains in place as there are other uses cases for it including the purger which writes out stales witnesses as invalid_witnesses.

### DIFF
--- a/iot_verifier/src/runner.rs
+++ b/iot_verifier/src/runner.rs
@@ -429,8 +429,8 @@ impl Runner {
         let iot_poc: IotPoc = IotPoc {
             poc_id: beacon_id,
             beacon_report: valid_beacon_report,
-            selected_witnesses,
-            unselected_witnesses,
+            selected_witnesses: selected_witnesses.clone(),
+            unselected_witnesses: unselected_witnesses.clone(),
         };
 
         let mut transaction = self.pool.begin().await?;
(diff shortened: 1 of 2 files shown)


bbalser (approved):

iot_verifier/src/runner.rs
```diff
@@ -451,6 +451,13 @@ impl Runner {
                 return Ok(());
             }
         }
+        // write out metrics for any witness which failed verification
+        // TODO: work our approach that doesnt require the prior cloning of
+        // the selected and unselected witnesses vecs
+        // tried to do this directly from the now discarded poc_proto
+        // but could nae get it to get a way past the lack of COPY
+        fire_invalid_witness_metric(&selected_witnesses);
```
jeffgrunewald, line 459: Looks like these could be combined with Vec::append() and make a single call to fire the metric?

jeffgrunewald (approved):

iot_verifier/src/runner.rs
madninja: I don't get why `fire_invalid_witness_metric` is called for selected_witnesses. they're a mix of valids and invalids right?

andymck: bot selected and unselected lists can contain invalids.  the function `fire_invalid_witness_metric` filters out the invalids and only fires the metric for those

PR #350, opened 2023-02-09
Attach hex_scale and reward_unit to poc receipt txns

vihu:
Summary
----
This PR is to attach `hex_scale` and `reward_unit` to poc receipts and poc witnesses present in the poc receipt txns.
Depends on [helium-proto#280](https://github.com/helium/proto/pull/280).

Notes
----
- Attach the `hex_scale` and `reward_unit` on txn construction
- Before signing the txn, zero out both of those fields
- Re-attach the originally constructed path back to the signed txn

TODO
----
- [x] Merge [helium-proto#280](https://github.com/helium/proto/pull/280)
- [x] Update proto back to master in dependencies
- [ ] Test on devnet
- [ ] On erlang side: Check that an unmodified proto can still handle these new, unsigned, fields AND then bump the proto and fix signature checking to also zero out these fields before checking the signature.

### DIFF
--- a/iot_config/src/gateway_service.rs
+++ b/iot_config/src/gateway_service.rs
@@ -7,8 +7,8 @@ use file_store::traits::MsgVerify;
 use helium_crypto::{Keypair, PublicKey, Sign};
 use helium_proto::{
     services::iot_config::{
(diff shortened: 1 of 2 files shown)


andymck (approved):
lgtm

bbalser (approved):

PR #351, opened 2023-02-10
widen witness loaded window to be > ingestor roll up time

andymck:
When loading witnesses the loader will take the beaon's window start/end times and widen it so that we load the roll up files not just for the same period as the beacons but those rolls up to the left and right of the beacons.
This is to ensure we do not miss witnesses which were included in a roll up just prior or after that of the beaconers.

The period of time the witness window should be widened by should be at least equal to that of the ingestor roll up time.  This was not the case with the current settings.  The ingestor roll up time is 5 mins, whilst the verifier was only widening the witness window by 2 mins on each side.

This fixes that by adding widening the witness window on both ends by the ingestor roll up time ( currently 5 mins )

A side effect of this change is that it would push the witness's window into the danger zone of being less that 2x that of the ingestor roll up time and thus susceptible to missing a file from one of the ingestors which has not yet made it to S3.
To accomodate this I have had to push the beaconers window start time back by another 1x the ingestor roll up time.  It now moves from 3x to 4x.

**NOTE: when deploying this on mainnet we will need to wipe the meta DB timestamp for `report` in order to permit the lookback to stretch to 4x.  Without changing that the loader's start point will be incorrect and thus push the witness endpoint into the danger zone again**

### DIFF
--- a/iot_verifier/src/loader.rs
+++ b/iot_verifier/src/loader.rs
@@ -209,8 +209,8 @@ impl Loader {
                 FileType::IotWitnessIngestReport,
                 &self.ingest_store,
                 gateway_cache,
-                after - ChronoDuration::seconds(60 * 6),
-                before + ChronoDuration::seconds(60 * 6),
+                after - ChronoDuration::minutes(6),
+                before + ChronoDuration::minutes(6),
                 None,
                 Some(&filter),
             )
--- a/iot_verifier/src/loader.rs
+++ b/iot_verifier/src/loader.rs
@@ -26,8 +26,8 @@ use twox_hash::XxHash64;
 use xorf::{Filter as XorFilter, Xor16};
 
 const REPORTS_META_NAME: &str = "report";
-/// cadence for how often to look for  reports from s3 buckets
-const REPORTS_POLL_TIME: u64 = 60 * 5;
+/// cadence for how often to look for  reports from s3 buckets in minutes
+const REPORTS_POLL_TIME: i64 = 5;
 
 const STORE_WORKERS: usize = 100;
 // DB pool size if the store worker count multiplied by the number of file types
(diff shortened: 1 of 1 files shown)


iot_verifier/src/loader.rs
```diff
@@ -201,14 +201,16 @@ impl Loader {
...
+        // of the ingestor roll up time
         // for witnesses we do need the filter but not the arc
         match self
             .process_events(
                 FileType::IotWitnessIngestReport,
                 &self.ingest_store,
                 gateway_cache,
-                after - ChronoDuration::seconds(60 * 2),
-                before + ChronoDuration::seconds(60 * 2),
+                after - ChronoDuration::seconds(60 * 6),
```
madninja, line 212: Should this be 10 minutes to follow the 2x roll up guideline?

also why not use the minute constructor on ChronoDuration?

andymck: i suspect one 1x is sufficient.  Each multiple involves doubling the amount of files we have to process, so we should only extend if necessary

re chrono seconds vs minutes...no reason...just what has been there.  I can change to minutes to give you a good friday

iot_verifier/src/loader.rs
```diff
@@ -26,8 +26,8 @@ use twox_hash::XxHash64;
 use xorf::{Filter as XorFilter, Xor16};
 
 const REPORTS_META_NAME: &str = "report";
-/// cadence for how often to look for  reports from s3 buckets
-const REPORTS_POLL_TIME: u64 = 60 * 5;
+/// cadence for how often to look for  reports from s3 buckets in minutes
+const REPORTS_POLL_TIME: i64 = 5;
```
madninja, line 30: ok.. so time for a lazy_static ChronoDuration::minutes here so you're not doing it multiple times in the code?

andymck: all of this is changed already in another PR: https://github.com/helium/oracles/pull/312

So i wouldnt get overly hung on on those details for this change ?

madninja: ah ok

madninja (approved):

bbalser (approved):

PR #352, opened 2023-02-10
beacon verification

andymck:


madninja (approved):

iot_verifier/src/poc.rs
```diff
@@ -406,6 +420,60 @@ fn verify_entropy(
...
+        entropy_version,
+        &beacon_report.local_entropy,
+        &beacon_report.remote_entropy,
+    ) {
+        Ok(res) => res,
+        Err(e) => {
+            tracing::warn!("failed to generate beacon, reason: {:?}", e);
+            return Err(InvalidReason::InvalidPacket);
+        }
+    };
```
maplant, line 454: This can be written more concisely with a let else binding https://doc.rust-lang.org/nightly/rust-by-example/flow_control/let_else.html

maplant (approved):
Besides my one comment, nothing sticks out to me so I _think_ this is good to go

iot_verifier/src/poc.rs
bbalser: I was thinking `map_err` would make it easier to read
```
    let generated_beacon = generate_beacon(
        &beacon_region_params,
        entropy_start.timestamp_millis(),
        entropy_version,
        &beacon_report.local_entropy,
        &beacon_report.remote_entropy,
    )
    .map_err(|e| {
        tracing::warn!("failed to cast report to beacon, reason: {:?}", e);
        InvalidReason::InvalidPacket
    })?;
```

bbalser (approved):

iot_verifier/src/poc.rs
andymck: used the map suggestion...

file_store/src/iot_beacon_report.rs
```diff
@@ -144,3 +145,30 @@ impl From<IotBeaconReport> for LoraBeaconReportReqV1 {
         }
     }
 }
+
+impl IotBeaconReport {
```
jeffgrunewald, line 149: why not `From<IotBeaconReport> for beacon::Beacon {}`?

iot_verifier/src/poc_report.rs
```diff
@@ -58,6 +58,7 @@ pub struct Report {
     pub created_at: Option<DateTime<Utc>>,
     #[sqlx(default)]
     pub timestamp: Option<DateTime<Utc>>,
+    pub version: Option<i32>,
```
jeffgrunewald, line 61: should this also be `entropy_version` so it's clear it refers to the version of the entropy struct and not the version of the Report struct?

file_store/src/iot_beacon_report.rs
andymck: because  the report doesnt have all the data required to generate the beacon.  Requires additional entropy data

iot_verifier/src/poc_report.rs
andymck: agree it should and same for timestamp.  but i would have to use serde rename and then backfill the changes.  Can follow that up with a diff PR.

file_store/src/iot_beacon_report.rs
jeffgrunewald: good point; wasn't paying attention to the other arguments

jeffgrunewald (approved):

PR #360, opened 2023-02-13
Use seconds precision for entropy timestamp.  Use hotspot generated entropy inputs and outputs for beacon verification test

andymck:
This addresses a mismatch on the beacon generation on the verifier with that of the beacon generation on hotspots where the resultant generated data payload values did not match.

The root cause of the issue was that the verifier was passing in millisecs precision for its remote entropy timestamp value when generating the beacon to verify whereas the hotspots uses seconds precision.

### DIFF
--- a/iot_verifier/src/poc.rs
+++ b/iot_verifier/src/poc.rs
@@ -404,7 +404,9 @@ fn verify_entropy(
     entropy_end: DateTime<Utc>,
     received_ts: DateTime<Utc>,
 ) -> GenericVerifyResult {
-    if received_ts < entropy_start || received_ts > entropy_end {
+    if received_ts.timestamp() < entropy_start.timestamp()
+        || received_ts.timestamp() > entropy_end.timestamp()
+    {
         tracing::debug!(
             "report verification failed, reason: {:?}.
                 received_ts: {:?},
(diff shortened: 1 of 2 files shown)


iot_verifier/src/poc.rs
```diff
@@ -404,7 +404,9 @@ fn verify_entropy(
     entropy_end: DateTime<Utc>,
     received_ts: DateTime<Utc>,
 ) -> GenericVerifyResult {
-    if received_ts < entropy_start || received_ts > entropy_end {
+    if received_ts.timestamp() < entropy_start.timestamp()
```
bbalser, line 407: I don't understand the need for this change

iot_verifier/src/poc.rs
```diff
@@ -451,6 +453,7 @@ fn verify_beacon_payload(
         InvalidReason::InvalidPacket
     })?;
     tracing::debug!("generated beacon {:?}", generated_beacon);
+    println!("generated beacon {:?}", generated_beacon);
```
bbalser, line 456: Did you mean to leave these `println!` in?

iot_verifier/src/poc.rs
andymck: This is not strictly necessary but as the source timestamps of entropy are in seconds precision, its better fitting to compare the receive timestamp at the same precision

iot_verifier/src/poc.rs
andymck: removed

madninja (approved):

bbalser (approved):

PR #361, opened 2023-02-13
Refactor poc_iot_injector to use continuous_source

bbalser:


poc_iot_injector/src/server.rs
```diff
@@ -48,171 +40,96 @@ impl Server {
...
-            }
             tokio::select! {
                 _ = shutdown.clone() => break,
-                _ = poc_iot_timer.tick() => match self.handle_poc_tick().await {
-                    Ok(()) => (),
-                    Err(err) => {
-                        tracing::error!("fatal poc_iot_injector error: {err:?}");
-                        return Err(err)
-                    }
+                msg = receiver.recv() => if let Some(stream) =  msg {
```
andymck, line 65: this seems weird in that why would the received msg potentially be a None ? why would the filestore push a None stream

bbalser: https://docs.rs/tokio/1.25.0/tokio/sync/mpsc/struct.Receiver.html#method.recv

The `recv` function always returns an `Option`. The file_store is not sending an Option on the channel.

poc_iot_injector/migrations/3_files_processed.sql
```diff
@@ -0,0 +1,11 @@
+create table files_processed (
+	file_name varchar primary key,
+	file_type varchar,
```
vihu, line 3: Do we want to make `file_type`, `file_timestamp` and `processed_at` not null as well?

vihu (approved):

poc_iot_injector/src/server.rs
maplant: To further elaborate, recv will return a None if the Sender has been dropped

maplant (approved):
-98, you love to see it

poc_iot_injector/src/server.rs
```diff
@@ -48,171 +40,96 @@ impl Server {
...
         tracing::info!("starting poc-iot-injector server");
-        let mut poc_iot_timer = time::interval(self.tick_time);
-        poc_iot_timer.set_missed_tick_behavior(time::MissedTickBehavior::Delay);
-
         loop {
-            if shutdown.is_triggered() {
-                break;
-            }
             tokio::select! {
                 _ = shutdown.clone() => break,
```
jeffgrunewald, line 64: does shutdown still need to be cloned if you're not checking it any longer in the top of the loop?

PR #362, opened 2023-02-14
Add denylist check cli

vihu:
Summary
----
This PR adds a check cli to the denylist crate. This is useful to check whether a given hotspot b58 address is on the denylist or not.

Usage:
```
./target/release/denylist check --key <b58_key> --filter-path <local_filter_path.bin>
```

### DIFF
--- a/denylist/src/denylist.rs
+++ b/denylist/src/denylist.rs
@@ -2,7 +2,12 @@ use crate::{client::DenyListClient, models::metadata::Asset, Error, Result};
 use bytes::Buf;
 use helium_crypto::{PublicKey, Verify};
 use serde::Serialize;
-use std::{fs, hash::Hasher, path, str::FromStr};
+use std::{
+    fs,
+    hash::Hasher,
+    path::{self, Path},
+    str::FromStr,
+};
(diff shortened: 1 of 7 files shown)


denylist/src/denylist.rs
```diff
@@ -51,6 +56,24 @@ impl DenyList {
         })
     }
 
+    pub fn from_filter<P: AsRef<Path>>(filter_path: P) -> Result<Self> {
```
andymck, line 59: given this code is shared pretty much as is with the `new()` function, be good to spin it out into a shared function and used by both new() and from_filter()

andymck (approved):

denylist/src/denylist.rs
vihu: Fair enough, I'll update shortly

vihu: Updated [a1ba4fd](https://github.com/helium/oracles/pull/362/commits/a1ba4fdba49ee939d75ac10fae9ee147bf8eb6ba)

andymck (approved):

PR #369, opened 2023-02-17
Transmit scaling with h3dex garbage collection

jeffgrunewald:
Support garbage collection of the h3dex again in the IoT Verifier:
- migrate the hex density scaler into the iot verifier crate to allow it access to the verifier's store of recent beacon activity data
- filter hotspots included in transmit scaling calculations to "recently active" hotspots
  - recent activity is here defined as "last beacon within 3600 minutes" which is intended to mirror the previous "last beacon within 3600 blocks" given an optimal 1 block/minute chain operating pace
- provide a default tx scaling factor for otherwise valid hotspots that do not have an entry found in the tx scaling density map
  - the default tx scale is intentionally set on the low end of the 0.1 -> 1.0 scale to discourage gaming by users continuing to re-assert hotspots or turn hotspots on and off for more than the 3600 minute timeframe without penalizing valid beacons/witnesses completely. the default only applies to beacons and witnesses measured until the next calculation of the tx scaling density map, currently every 30 minutes


iot_verifier/src/poc.rs
```diff
@@ -127,11 +137,13 @@ impl Poc {
...
-                    Ok(VerifyBeaconResult::valid(beaconer_info, scaling_factor))
-                } else {
-                    Ok(VerifyBeaconResult::scaling_factor_not_found(beaconer_info))
-                }
+                let tx_scale =
+                    if let Some(scaling_factor) = hex_density_map.get(beaconer_location).await {
+                        scaling_factor
+                    } else {
+                        *DEFAULT_TX_SCALE
+                    };
```
andymck, line 145:
```suggestion
let tx_scale = hex_density_map.get(beaconer_location).await.unwrap_or(*DEFAULT_TX_SCALE);
```

iot_verifier/src/poc.rs
```diff
@@ -220,11 +232,13 @@ impl Poc {
                     return Ok(VerifyWitnessResult::not_asserted(InvalidParticipantSide::Beaconer))
                 };
 
-                if let Some(hex_scale) = hex_density_map.get(beaconer_location).await {
-                    Ok(VerifyWitnessResult::valid(witness_info, hex_scale))
+                let tx_scale = if let Some(hex_scale) = hex_density_map.get(beaconer_location).await
```
andymck, line 235: similar to previous instance of setting tx_scale, could use a unwrap_or in place of the if else

iot_verifier/src/tx_scaler.rs
```diff
@@ -0,0 +1,114 @@
...
+impl Server {
+    pub async fn from_settings(settings: &Settings) -> Result<Self, TxScalerError> {
+        let mut server = Self {
+            follower: FollowerService::from_settings(&settings.follower),
+            hex_density_map: SharedHexDensityMap::new(),
+            pool: settings.database.connect(2).await?,
+            trigger_interval: Duration::seconds(settings.transmit_scale_interval),
+        };
+
+        server.refresh_scaling_map().await?;
```
andymck, line 42: why refresh here ?  The tick will fire straight away and thus wont you be refreshing twice upon creation ?

Hmm actually i see we already do the same thing on main branch and thus it looks like we are generating the scaler twice currently on mainnet during startup.

jeffgrunewald: this was the best way we could figure out how to init the scaling map given that the rust timer ticks immediately but _also_ that we need the scaling map to be able to populate before we start processing poc reports to avoid the first 5-10 minutes of reports all failing to lookup a tx scale

iot_verifier/src/poc.rs
```diff
@@ -38,6 +39,15 @@ const POC_CELL_DISTANCE_MINIMUM: u32 = 8;
 /// the resolution at which parent cell distance is derived
 const POC_CELL_PARENT_RES: u8 = 11;
 
+lazy_static! {
+    /// Scaling factor when inactive gateway is not found in the tx scaling map (20%).
+    /// A default tx scale is required to allow for inactive hotspots to become active
+    /// again when an inactive hotspot's h3 index would otherwise be garbage-collected
+    /// from density scaling calculations and not finding a value on subsequent lookups
+    /// would disqualify the hotspot from validating further beacons
+    static ref DEFAULT_TX_SCALE: Decimal = Decimal::new(2000, 4);
```
vihu, line 48: Just confirming that this value is `0.2000`

iot_verifier/src/poc.rs
jeffgrunewald: excellent point!

iot_verifier/src/poc.rs
jeffgrunewald: according to the `rust_decimal` docs https://docs.rs/rust_decimal/latest/rust_decimal/struct.Decimal.html#method.new this should equate to
```
assert_eq!("0.200", Decimal::new(2000, 4).to_string())
```

iot_verifier/src/tx_scaler.rs
jeffgrunewald: fixed [97e9ef5](https://github.com/helium/oracles/pull/369/commits/97e9ef557bd9b8daccbc4b32915de4d51715802a)

iot_verifier/src/tx_scaler.rs
```diff
@@ -0,0 +1,114 @@
...
+        &self,
+        now: DateTime<Utc>,
+    ) -> Result<HashMap<Vec<u8>, DateTime<Utc>>, sqlx::Error> {
+        let interactivity_deadline = now - Duration::minutes(HIP_17_INTERACTIVITY_LIMIT);
+        Ok(
+            LastBeacon::get_all_since(interactivity_deadline, &self.pool)
+                .await?
+                .into_iter()
+                .map(|beacon| (beacon.id, beacon.timestamp))
+                .collect::<HashMap<Vec<u8>, DateTime<Utc>>>(),
```
jeffgrunewald, line 111: per a suggestion by @bbalser i switched from collecting the active gateways into a Vec and doing a `contains` check to collecting them into a HashMap with a `contains_key` check as this showed significant constant time lookup improvements getting until into the 100s of thousands of map entries (constant microseconds in the teens to 20s) as opposed to O(n) time increasing into the 60s-70s of milliseconds towards the end of the Vec)

andymck (approved):

bbalser (approved):

vihu (approved):

PR #372, opened 2023-02-20
invalidate dup witness reports

andymck:
Address duplicate witness reports received for any one individual beacon and includes a slight refactor of witness verification flow.

The first witness report from any one gateway is accepted.  Any subsequent report from the same gateway for the same beacon will be rendered invalid with reason 'duplicate'.  Verifications will not be run on duplicates.

Reports marked as duplicate will be excluded from the last 14 and will make it to S3 as part of the `unselected` list.

There are multiple ways to handle duplications but the key considerations are:

1.  Do we require a report to make it to S3 for any witness rendered a dup ?
2.  In the scenario whereby we have both a valid and N invalids report from the same gateway....which one is deduped ?  How do we decide on that ?  
3. In the scenario whereby we have dups that include 1 valid and one or more invalids ( invalid in the sense they would fail verifications ) do we allow for the possibility that the last 14 could consist of both a valid and an invalid report from the same gateway ?
4.  Can a report marked as a dup be selected for the last 14 ?

This PR deals with each of the above via:

1. Dup reports make it to S3, they are not dropped to the floor
2. Witnesses are sorted based on received timestamp ( time they hit the ingestor ).  The first received report is accepted, any subsequent report is rejected and marked as a dup
3.  Only the first witness report has the potential to make it to the last 14.  A gateway cannot have more that 1 report in the last 14 
4.  Reports marked as dups cannot make it to the last 14 list

### DIFF
--- a/iot_verifier/src/poc.rs
+++ b/iot_verifier/src/poc.rs
@@ -8,7 +8,6 @@ use chrono::{DateTime, Duration, Utc};
 use density_scaler::HexDensityMap;
 use file_store::{
     iot_beacon_report::{IotBeaconIngestReport, IotBeaconReport},
-    iot_invalid_poc::IotInvalidWitnessReport,
     iot_valid_poc::IotVerifiedWitnessReport,
     iot_witness_report::IotWitnessIngestReport,
 };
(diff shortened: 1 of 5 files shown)


iot_verifier/src/poc.rs
```diff
@@ -357,6 +376,24 @@ impl Poc {
...
+        Ok(IotVerifiedWitnessReport {
+            received_timestamp: witness_report.received_timestamp,
+            status: VerificationStatus::Invalid,
+            report: witness_report.report,
+            location: None,
+            hex_scale: Decimal::ZERO,
+            // default reward units to zero until we've got the full count of
+            // valid, non-failed witnesses for the final validated poc report
+            reward_unit: Decimal::ZERO,
+            invalid_reason: InvalidReason::InvalidCapability,
```
madninja, line 392: Missing the actual "duplicate" error?

andymck: yeah proto is not merged, so using this to see if the PR goes green.

bbalser (approved):

bbalser (approved):

jeffgrunewald (approved):

PR #374, opened 2023-02-22
Update FileInfoPoller return type to implement Future and unwrap oute…

bbalser:
…r Result


maplant (approved):

andymck (approved):
looks a lil weird but ok...

PR #379, opened 2023-02-23
Mobile Packet Verifier 

maplant:


mobile_packet_verifier/src/verifier.rs
```diff
@@ -0,0 +1,89 @@
...
+        }
+    }
+
+    pub async fn run(mut self, shutdown: &triggered::Listener) -> Result<(), VerificationError> {
+        loop {
+            tokio::select! {
+                _ = shutdown.clone() => return Ok(()),
+                file = self.reports.recv() => {
+                    if let Some(file) = file {
+                        let _db_lock = self.db_lock.lock().await;
```
bbalser, line 75: What is the reason for this db_lock?

mobile_packet_verifier/src/burner.rs
```diff
@@ -0,0 +1,287 @@
...
+
+            // Delete from the data transfer session and write out to S3
+
+            sqlx::query("DELETE FROM data_tranfer_sessions WHERE payer = $1")
+                .bind(payer)
+                .execute(&self.pool)
+                .await?;
+
+            for session in sessions {
+                self.valid_sessions
```
bbalser, line 226: auto_commit is set to false for this file_sink, but I don't see the call to commit anywhere.

bbalser:
Why do we wait until the end of the burn period to write the valid sessions to s3?  That means we will output the valid session once a day?

maplant:
> Why do we wait until the end of the burn period to write the valid sessions to s3?  That means we will output the valid session once a day?

Because we can only write out rewardable sessions after we've burned the data credits for them. Otherwise we could erroneously reward. to compensate, the sessions are accumulated

bbalser (approved):

mobile_packet_verifier/src/main.rs
```diff
@@ -0,0 +1,101 @@
...
+
+        let burner = Burner::new(&settings, valid_sessions, rpc_client, burn_keypair).await?;
+
+        let file_store = FileStore::from_settings(&settings.ingest).await?;
+
+        let (shutdown_trigger, shutdown_listener) = triggered::trigger();
+        tokio::spawn(async move {
+            let _ = tokio::signal::ctrl_c().await;
+            shutdown_trigger.trigger()
+        });
```
jeffgrunewald, line 71: move this above the rest of the setup in case any of that encounters a hang and the operator needs to kill it?

mobile_packet_verifier/src/verifier.rs
```diff
@@ -0,0 +1,44 @@
...
+    curr_file_ts: DateTime<Utc>,
+    reports: impl Stream<Item = DataTransferSessionIngestReport>,
+) -> Result<(), VerificationError> {
+    tokio::pin!(reports);
+
+    while let Some(DataTransferSessionIngestReport { report, .. }) = reports.next().await {
+        sqlx::query(
+            r#"
+            INSERT INTO data_transfer_sessions (pub_key, payer, uploaded_bytes, downloaded_bytes, first_timestamp, last_timestamp)
+            VALUES ($1, $2, $3, $4, $5, $5)
```
jeffgrunewald, line 27:
```suggestion
            VALUES ($1, $2, $3, $4, $5, $6)
```

mobile_packet_verifier/src/verifier.rs
```diff
@@ -0,0 +1,44 @@
...
+pub enum VerificationError {
+    #[error("file store error: {0}")]
+    FileStoreError(#[from] file_store::Error),
+    #[error("sqlx error: {0}")]
+    SqlxError(#[from] sqlx::Error),
+    #[error("reports stream dropped")]
+    ReportsStreamDropped,
+}
+
+pub async fn verify(
```
jeffgrunewald, line 16: is this module going to be doing additional things over time? this doesn't seem to be "verifying" anything

mobile_packet_verifier/src/verifier.rs
```diff
@@ -0,0 +1,44 @@
...
+    reports: impl Stream<Item = DataTransferSessionIngestReport>,
+) -> Result<(), VerificationError> {
+    tokio::pin!(reports);
+
+    while let Some(DataTransferSessionIngestReport { report, .. }) = reports.next().await {
+        sqlx::query(
+            r#"
+            INSERT INTO data_transfer_sessions (pub_key, payer, uploaded_bytes, downloaded_bytes, first_timestamp, last_timestamp)
+            VALUES ($1, $2, $3, $4, $5, $5)
+            ON CONFLICT (pub_key, payer) DO UPDATE SET
```
jeffgrunewald, line 28: is the intention that within a given burn period any new data transfer for a given `(pub_key, payer)` will get appended to the existing up/down totals and the `last_timestamp` updated to the latest and that _after_ a given burn period the records in the db table are wiped? what happens if there are more than one "session" of data transfer within a given period for a single `(pub_key, payer)` combo?

mobile_packet_verifier/src/daemon.rs
```diff
@@ -0,0 +1,57 @@
...
+        Self {
+            pool,
+            burner,
+            reports,
+            burn_period: Duration::from_secs(60 * 60 * settings.burn_period as u64),
+        }
+    }
+
+    pub async fn run(mut self, shutdown: &triggered::Listener) -> anyhow::Result<()> {
+        let mut burn_time = Instant::now() + self.burn_period;
```
jeffgrunewald, line 34: maybe a `Settings` method that computes `self.burn_time()` since this seems to always be a static duration determined by the `burn_period` settings value?

file_store/src/mobile_session.rs
```diff
@@ -0,0 +1,88 @@
...
+}
+
+impl TryFrom<DataTransferSessionReqV1> for DataTransferSessionReq {
+    type Error = Error;
+
+    fn try_from(v: DataTransferSessionReqV1) -> Result<Self> {
+        let timestamp = v.timestamp()?;
+        let radio_access_technology = DataTransferRadioAccessTechnology::from_i32(
+            v.radio_access_technology,
+        )
```
jeffgrunewald, line 69:
```suggestion
        let radio_access_technology = v.radio_access_technology(),
```
would this not work in this scenario?

mobile_packet_verifier/src/burner.rs
```diff
@@ -0,0 +1,266 @@
...
+            // Fetch the sub dao epoch info:
+            let epoch = self.provider.get_epoch_info().await?.epoch;
+            let (sub_dao_epoch_info, _) = Pubkey::find_program_address(
+                &[
+                    "sub_dao_epoch_info".as_bytes(),
+                    self.program_cache.sub_dao.as_ref(),
+                    &epoch.to_le_bytes(),
+                ],
+                &helium_sub_daos::ID,
+            );
```
jeffgrunewald, line 128: do we want to do these steps for each payer in the collection of payer totals or just once for each burn session before going into the iterator?

mobile_packet_verifier/src/burner.rs
```diff
@@ -0,0 +1,266 @@
...
+            let escrow_account =
+                DelegatedDataCreditsV0::try_deserialize(&mut account_data)?.escrow_account;
+
+            tracing::info!("Burning {} DC from {}", amount, payer);
+
+            let instructions = {
+                let request = RequestBuilder::from(
+                    data_credits::id(),
+                    &self.cluster,
+                    std::rc::Rc::new(Keypair::from_bytes(&self.keypair).unwrap()),
```
jeffgrunewald, line 145: is this an infallible call and doesn't need to be threadsafe?

mobile_packet_verifier/src/burner.rs
```diff
@@ -0,0 +1,266 @@
...
+                let accounts = accounts::BurnDelegatedDataCreditsV0 {
+                    sub_dao_epoch_info,
+                    dao: self.program_cache.dao,
+                    sub_dao: self.program_cache.sub_dao,
+                    account_payer: self.program_cache.account_payer,
+                    data_credits: self.program_cache.data_credits,
+                    delegated_data_credits: crate::pdas::delegated_data_credits(
+                        &self.program_cache.sub_dao,
+                        &payer,
+                    ),
```
jeffgrunewald, line 159: can you reuse the `ddc_key` you already computed on line 133?

mobile_packet_verifier/src/burner.rs
```diff
@@ -0,0 +1,266 @@
...
+                // error paths.
+                request
+                    .accounts(accounts)
+                    .args(args)
+                    .instructions()
+                    .unwrap()
+            };
+
+            let blockhash = self.provider.get_latest_blockhash().await?;
+            let signer = Keypair::from_bytes(&self.keypair).unwrap();
```
jeffgrunewald, line 182: why do we convert this keypair to bytes when we get a new `Burner` if we're converting it back to bytes in the two places where we actually use it? couldn't we just stash the `Keypair` and avoid the potentially fallible `unwrap`s?

mobile_packet_verifier/src/pdas.rs
```diff
@@ -0,0 +1,22 @@
...
+//! actions on the Solana chain.
+
+use helium_crypto::PublicKeyBinary;
+use sha2::{Digest, Sha256};
+use solana_sdk::pubkey::Pubkey;
+
+/// Returns the PDA for the Delegated Data Credits of the given `payer`.
+pub fn delegated_data_credits(sub_dao: &Pubkey, payer: &PublicKeyBinary) -> Pubkey {
+    let mut hasher = Sha256::new();
+    hasher.update(&payer.as_ref()[1..]);
```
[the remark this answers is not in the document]
maplant: This is actually incorrect. It needs to be the B58 representation

mobile_packet_verifier/src/burner.rs
maplant: Keypair does not implement Copy or Clone unfortunately.

mobile_packet_verifier/src/burner.rs
maplant: This is infallible, as these bytes are guaranteed to be from a Keypair. This call actually requires a Rc, but it's not used across an await boundary so it is fine.

file_store/src/file_info.rs
```diff
@@ -37,6 +37,12 @@ impl FromStr for FileInfo {
     }
 }
 
+impl fmt::Display for FileInfo {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "{}", self.key)
```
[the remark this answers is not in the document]
maplant: That's already conveyed, so it would be redundant. Plus that's already conveyed in the debug info

mobile_packet_verifier/src/burner.rs
maplant: I thought we would have to do it individually, but we don't really. The epoch is per day, so we'll have an issue if this straddles a day

Cargo.toml
```diff
@@ -50,9 +52,11 @@ sqlx = {version = "0", features = [
   "runtime-tokio-rustls"
 ]}
 
-helium-crypto = {version = "0.6.3", features=["sqlx-postgres", "multisig"]}
+helium-crypto = {version = "0.6.8", features=["sqlx-postgres", "multisig"]}
```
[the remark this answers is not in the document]
maplant: Unfortunately that causes a dependency issue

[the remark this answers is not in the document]
jeffgrunewald: 🤦 pretty sure that was supposed to _resolve_ a dependency issue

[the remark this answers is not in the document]
maplant: lol, probably

mobile_packet_verifier/src/verifier.rs
maplant: It isn't. Not sure what a better module name would be. Accumulate maybe?

mobile_packet_verifier/src/burner.rs
maplant: ah yes, I definitely can

mobile_packet_verifier/src/verifier.rs
maplant: I'm not sure I understand the question, for the given period all of the data_transfer sessions are accumulated. Perhaps I need to change the `last_timestamp` to the max of the current last timestamp and the new one?

mobile_packet_verifier/src/settings.rs
```diff
@@ -0,0 +1,77 @@
...
+            .build()
+            .and_then(|config| config.try_deserialize())
+    }
+
+    pub fn dc_mint(&self) -> Result<Pubkey, ParsePubkeyError> {
+        self.dc_mint.parse()
+    }
+
+    pub fn dnt_mint(&self) -> Result<Pubkey, ParsePubkeyError> {
+        self.dnt_mint.parse()
```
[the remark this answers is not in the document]
maplant: The "subdao mint" pubkey

mobile_packet_verifier/src/verifier.rs
jeffgrunewald: you answered my question (there's only one "data transfer session" per burn period and the table is wiped after each period. i think catching the `MAX` of the timestamps is a good catch though in case the files maybe got received/processed out of order?

mobile_packet_verifier/src/settings.rs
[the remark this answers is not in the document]
jeffgrunewald: so probably "dao native token"?

mobile_packet_verifier/src/verifier.rs
maplant: It shouldn't really matter because it's the file timestamp and no the session timestamp

jeffgrunewald (approved):

PR #381, opened 2023-02-24
update beacon payload comparison, extend tests

andymck:


iot_verifier/src/poc.rs
```diff
@@ -440,12 +440,12 @@ fn verify_beacon_payload(
...
 
     // compare reports
-    if reported_beacon != generated_beacon {
+    if !generated_beacon.verify(&reported_beacon) {
         tracing::info!(
             "beacon construction verification failed, pubkey {:?}",
             beacon_report.pub_key,
         );
-        return Ok(());
+        return Err(InvalidReason::InvalidPacket);
```
jeffgrunewald, line 448: should this tracing statement be dropped down to debug again and should this test be made configurable given the churn/impact ratio this particular validation has? maybe we write a quick macro/function that we can use as a feature flag to quickly turn on and off validations that unexpectedly prove to be problematic and then remove them after a full release or two?

andymck: downgraded logging here: d928ede004388e62b83b13727b0072c1911724ab

macro/setting to disable the test seems overkill and its late on friday so a no go from me :)

madninja (approved):

PR #383, opened 2023-02-24
Change db_store to be able to create a connection to db with postgres…

bbalser:
… passwor or iam authentication

### DIFF
--- a/db_store/src/iam_auth_pool.rs
+++ b/db_store/src/iam_auth_pool.rs
@@ -7,8 +7,10 @@ use sqlx::{
 use aws_credential_types::Credentials;
 use aws_sig_auth::signer::{self, HttpSignatureType, OperationSigningConfig, RequestConfig};
 use aws_smithy_http::body::SdkBody;
-use aws_types::region::{Region, SigningRegion};
-use aws_types::SigningService;
+use aws_types::{
+    region::{Region, SigningRegion},
+    SigningService,
+};
 use std::time::{Duration, SystemTime, UNIX_EPOCH};
 
 pub async fn connect(
--- a/db_store/src/iam_auth_pool.rs
+++ b/db_store/src/iam_auth_pool.rs
@@ -1,0 +1,192 @@
+use crate::{Error, Result, Settings};
+use sqlx::{
+    postgres::{PgConnectOptions, PgPoolOptions, Postgres},
+    Pool,
+};
+
+use aws_credential_types::Credentials;
+use aws_sig_auth::signer::{
+    self, HttpSignatureType, OperationSigningConfig, RequestConfig, SigningError,
+};
+use aws_smithy_http::body::SdkBody;
+use aws_types::region::{Region, SigningRegion};
+use aws_types::SigningService;
+use std::time::{Duration, SystemTime, UNIX_EPOCH};
+
+pub async fn connect(
+    settings: &Settings,
+    default_max_connections: usize,
+    shutdown: triggered::Listener,
+) -> Result<(Pool<Postgres>, futures::future::BoxFuture<'static, Result>)> {
+    let aws_config = aws_config::load_from_env().await;
+    let client = aws_sdk_sts::Client::new(&aws_config);
+    let connect_options = connect_options(&client, settings).await?;
+
+    let pool = PgPoolOptions::new()
+        .max_connections(
+            settings
+                .max_connections
+                .unwrap_or(default_max_connections as u32),
+        )
+        .connect_with(connect_options)
+        .await?;
+
+    let cloned_settings = settings.clone();
+    let cloned_pool = pool.clone();
+    let join_handle =
+        tokio::spawn(async move { run(client, cloned_settings, cloned_pool, shutdown).await });
+
+    Ok((
+        pool,
+        Box::pin(async move {
+            match join_handle.await {
+                Ok(Ok(())) => Ok(()),
+                Ok(Err(err)) => Err(err),
+                Err(err) => Err(Error::from(err)),
+            }
+        }),
+    ))
+}
+
+async fn run(
+    client: aws_sdk_sts::Client,
+    settings: Settings,
+    pool: Pool<Postgres>,
+    shutdown: triggered::Listener,
+) -> Result {
+    let duration = std::time::Duration::from_secs(settings.iam_duration_seconds.unwrap() as u64)
+        - Duration::from_secs(120);
+
+    loop {
+        let shutdown = shutdown.clone();
+
+        tokio::select! {
+            _ = shutdown => break,
+            _ = tokio::time::sleep(duration) => {
+                let connect_options = connect_options(&client, &settings).await?;
+                pool.set_connect_options(connect_options);
+            }
+        }
+    }
+
+    Ok(())
+}
+
+async fn connect_options(
+    client: &aws_sdk_sts::Client,
+    settings: &Settings,
+) -> Result<PgConnectOptions> {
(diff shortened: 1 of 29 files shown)


iot_config/src/main.rs
```diff
@@ -60,34 +60,39 @@ impl Daemon {
...
         let (shutdown_trigger, shutdown_listener) = triggered::trigger();
         tokio::spawn(async move {
             let _ = signal::ctrl_c().await;
             shutdown_trigger.trigger()
         });
 
+        // Create database pool
+        let (pool, db_join_handle) = settings
+            .database
+            .connect(10, shutdown_listener.clone())
```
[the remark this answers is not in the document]
madninja: why would it not be able to use the max_connections  in the database settings stanza? or some other setting with a default value so it's not hardcoded in every crate?

db_store/src/iam_auth_pool.rs
```diff
@@ -0,0 +1,192 @@
...
+use aws_types::region::{Region, SigningRegion};
+use aws_types::SigningService;
+use std::time::{Duration, SystemTime, UNIX_EPOCH};
+
+pub async fn connect(
+    settings: &Settings,
+    default_max_connections: usize,
+    shutdown: triggered::Listener,
+) -> Result<(Pool<Postgres>, futures::future::BoxFuture<'static, Result>)> {
+    let aws_config = aws_config::load_from_env().await;
```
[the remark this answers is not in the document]
bbalser: This call cannot fail as it is just pulling the config from the env, the actual calls we make using this configuration will fail later if the config is not correct.

db_store/src/iam_auth_pool.rs
```diff
@@ -0,0 +1,187 @@
...
+        .body(SdkBody::empty())
+        .expect("valid request");
+    let _signature = signer.sign(
+        &operation_config,
+        &request_config,
+        credentials,
+        &mut request,
+    )?;
+    let mut uri = request.uri().to_string();
+    assert!(uri.starts_with("http://"));
```
jeffgrunewald, line 184: does this not support/expect `https://` since it's talking to the aws metadata service so it's expected to _always_ be private to the vpc the host is running on?

iot_config/src/main.rs
```diff
@@ -60,34 +60,36 @@ impl Daemon {
...
+
         let listen_addr = settings.listen_addr()?;
 
-        let gateway_svc = GatewayService::new(settings).await?;
-        let route_svc = RouteService::new(settings).await?;
-        let org_svc = OrgService::new(settings, route_svc.clone_update_channel()).await?;
+        let gateway_svc = GatewayService::new(settings, pool.clone()).await?;
+        let route_svc = RouteService::new(settings, pool.clone()).await?;
+        let org_svc =
+            OrgService::new(settings, route_svc.clone_update_channel(), pool.clone()).await?;
```
jeffgrunewald, line 79: i had to refactor the startup for all of these services so this file will want a rebase

iot_verifier/src/loader.rs
```diff
@@ -29,11 +29,6 @@ const REPORTS_META_NAME: &str = "report";
 /// cadence for how often to look for  reports from s3 buckets in minutes
 const REPORTS_POLL_TIME: i64 = 5;
 
-const STORE_WORKERS: usize = 100;
```
jeffgrunewald, line 32: was this already not in use or is this a typo?

jeffgrunewald (commented):
i think this is a rebase away from ready

db_store/src/iam_auth_pool.rs
bbalser: This function is copied from the library example and since it works, I haven't changed it at all.

https://docs.rs/aws-sig-auth/0.54.1/aws_sig_auth/#generate-rds-iam-token

iot_config/src/main.rs
bbalser: Rebased

db_store/src/iam_auth_pool.rs
```diff
@@ -0,0 +1,187 @@
...
+    let uri = uri.split_off("http://".len());
```
jeffgrunewald, line 185: wait, you're checking to see that the uri starts with `"http://"`, effectively panicking if it doesn't, and then immediately chopping off the prefix? 🤣

db_store/src/settings.rs
```diff
@@ -1,45 +1,70 @@
...
-            )
-            .connect_with(connect_options)
-            .await?;
+    pub async fn connect(
+        &self,
+        shutdown: triggered::Listener,
+    ) -> Result<(Pool<Postgres>, futures::future::BoxFuture<'static, Result>)> {
+        match self.auth_type {
+            AuthType::Postgres => match self.simple_connect().await {
+                Ok(pool) => Ok((pool, Box::pin(async move { Ok(()) }))),
```
jeffgrunewald, line 46: since this empty "handle" task isn't actually doing anything it won't hold up shutting down the app even though it's not listening for a shutdown signal? it'll just return immediately and the `try_join` in an app's main function will just keep on waiting for the _rest_ of the threads to complete or error?

jeffgrunewald (approved):

db_store/src/settings.rs
bbalser: This Future will immediately resolve to an `OK(())` Result and not cause the try_join to hang, since it is already resolved.

db_store/src/iam_auth_pool.rs
madninja: please don't be panicking in a function.. return an error..

bbalser: Removed b5bc865

bbalser: Fixed 56926e9

db_store/src/iam_auth_pool.rs
```diff
@@ -0,0 +1,227 @@
...
+                        Error::InvalidAssumedCredentials("no access_key_id".to_string())
+                    })?,
+                    creds.secret_access_key().ok_or_else(|| {
+                        Error::InvalidAssumedCredentials("no secret_access_key".to_string())
+                    })?,
+                    creds.session_token().map(|s| s.to_string()),
+                    creds
+                        .expiration()
+                        .map(|e| UNIX_EPOCH + Duration::from_secs(e.secs() as u64)),
+                    "Helium Foundation",
```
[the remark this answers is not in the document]
bbalser: It is the `provider_name`, I don't know how it is used, I put that string there and everything worked.

db_store/src/iam_auth_pool.rs
```diff
@@ -0,0 +1,227 @@
...
+        - Duration::from_secs(120);
+
+    loop {
+        let shutdown = shutdown.clone();
+
+        tokio::select! {
+            _ = shutdown => break,
+            _ = tokio::time::sleep(duration) => {
+                let connect_options = connect_parameters.connect_options(&client).await?;
+                pool.set_connect_options(connect_options);
```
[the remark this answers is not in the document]
bbalser: Not sure I understand the question, but if the pool wants to reconnect it will use the `ConnectOptions` to do so.  This overwrites the `ConnectOptions` the pool has before the auth token expires, so it should always have a valid auth token and can reconnect whenever it wants.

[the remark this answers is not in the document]
jeffgrunewald: oh, so the pool will always try to reconnect and this just ensures that it can?

jeffgrunewald (approved):

andymck (approved):

PR #393, opened 2023-02-27
verifier changes for helius integration

andymck:
WIP branch...

Initial requirements at replacing node follower usage in the IOT verifier

Region data continues to be pulled from node follower temporarily via RPC. This will be replaced with either a call to the config service or an RPC to helius ( tbd)  Previously the node data would have been returned from the follower as part of the gateway info and cached in the IOT verifier. Given the gateway info now originates from helius and that helius does not manage region data, deriving region data is now distinct from gateway info. Added a cache to store region data per gateway


andymck:
closing as verifier and other apps will no longer go direct to helius but instead use config service

PR #394, opened 2023-02-28
Config Service bugfixes, Admin API and Gateway Location RPC

jeffgrunewald:
! Depends on https://github.com/helium/proto/pull/288

This doozy introduces several bug fixes in the implementation of the streaming RPCs, retrieving records from the DB and decoding their types, integer size changes in the upstream proto dependency, and devaddr constraints records

- Introduces devaddr constraint bounds checking when attempting to add new devaddr ranges to routers
- Implements the Gateway service's Location rpc
- Implements the Admin service's rpcs and an AuthCache shared between all the services for efficient validation of administrative requests
- Restructures the region_params management rpc underneath the Admin service (formerly under the Gateway service)
- Extends the docker-compose stack to include a mock node-follower service for easier testing

### DIFF
--- a/iot_config/src/admin_service.rs
+++ b/iot_config/src/admin_service.rs
@@ -1,0 +1,90 @@
+use crate::{
+    region_map::{self, RegionMap},
+    GrpcResult, Settings,
+};
+use anyhow::Result;
+use file_store::traits::MsgVerify;
+use helium_crypto::PublicKey;
+use helium_proto::{
+    services::iot_config::{
+        self, AdminAddKeyReqV1, AdminKeyResV1, AdminLoadRegionReqV1, AdminLoadRegionResV1,
+        AdminRemoveKeyReqV1,
+    },
+    Region,
+};
+use sqlx::{Pool, Postgres};
+use tonic::{Request, Response, Status};
+
+pub struct AdminService {
+    admin_pubkey: PublicKey,
+    pool: Pool<Postgres>,
+    region_map: RegionMap,
+}
+
+impl AdminService {
+    pub fn new(settings: &Settings, pool: Pool<Postgres>, region_map: RegionMap) -> Result<Self> {
+        Ok(Self {
+            admin_pubkey: settings.admin_pubkey()?,
+            pool,
+            region_map,
+        })
+    }
+
+    fn verify_admin_signature<R>(&self, request: R) -> Result<R, Status>
+    where
+        R: MsgVerify,
+    {
+        request
+            .verify(&self.admin_pubkey)
+            .map_err(|_| Status::permission_denied("invalid admin signature"))?;
+        Ok(request)
+    }
+}
+
+#[tonic::async_trait]
+impl iot_config::Admin for AdminService {
+    async fn add_key(&self, _request: Request<AdminAddKeyReqV1>) -> GrpcResult<AdminKeyResV1> {
+        unimplemented!();
+    }
+
+    async fn remove_key(
+        &self,
+        _request: Request<AdminRemoveKeyReqV1>,
+    ) -> GrpcResult<AdminKeyResV1> {
+        unimplemented!();
+    }
+
+    async fn load_region(
+        &self,
+        request: Request<AdminLoadRegionReqV1>,
+    ) -> GrpcResult<AdminLoadRegionResV1> {
+        let request = request.into_inner();
+        let req = self.verify_admin_signature(request)?;
+
+        let region = Region::from_i32(req.region)
+            .ok_or_else(|| Status::invalid_argument("invalid region"))?;
+
+        let params = match req.params {
+            Some(params) => params,
+            None => return Err(Status::invalid_argument("missing region")),
+        };
+
+        let idz = if !req.hex_indexes.is_empty() {
+            Some(req.hex_indexes.as_ref())
+        } else {
+            None
+        };
+
+        let updated_region = region_map::update_region(region, &params, idz, &self.pool)
+            .await
+            .map_err(|_| Status::internal("region update failed"))?;
+
+        self.region_map.insert_params(region, params).await;
+        if let Some(region_tree) = updated_region {
+            tracing::debug!("New compacted region map with {} cells", region_tree.len());
+            self.region_map.replace_tree(region_tree).await;
+        }
+
+        Ok(Response::new(AdminLoadRegionResV1 {}))
+    }
+}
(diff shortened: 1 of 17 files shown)


iot_config/src/admin_service.rs
```diff
@@ -0,0 +1,153 @@
...
+#[tonic::async_trait]
+impl iot_config::Admin for AdminService {
+    async fn add_key(&self, request: Request<AdminAddKeyReqV1>) -> GrpcResult<AdminKeyResV1> {
+        let request = request.into_inner();
+
+        self.verify_request_signature(&request).await?;
+
+        let key_type = KeyType::from_i32(request.key_type)
+            .map_err(|_| Status::invalid_argument("invalid key type supplied"))?;
+        let pubkey = self
```
andymck, line 79: is this check not redundant, in that if the pub key is invalid wont it already be caught at the sig verification on line 75?

andymck (approved):
A lot going on in here but on an initial pass lgtm.  

One observation is that we dont appear to have any tests for the config service which may bite at the client integration phase

iot_config/src/admin_service.rs
jeffgrunewald: Good question but these are checking different keys/things.
They function on line 75 is checking the request to add the new key was signed by an already-registered key (the default key in the db or any admin keys saved to the db) so it would fail if it attempted to validate the request with the key being submitted to the `add_key` method. 
this check validates the key you’re trying to register is a valid key worth saving and also converts the bytes to a PublicKey in order to do the follow up check that it’s a valid key for the network this config instance is servicing (mainnet vs devnet)

jeffgrunewald:
> A lot going on in here but on an initial pass lgtm.
> 
> One observation is that we dont appear to have any tests for the config service which may bite at the client integration phase

I agree with you; looking forward to wrapping all of my DB accessors in a trait i can test from outside the rpc methods

iot_config/src/gateway_service.rs
```diff
@@ -17,48 +15,49 @@ use node_follower::{
...
 
 #[tonic::async_trait]
 impl iot_config::Gateway for GatewayService {
     async fn location(
         &self,
-        _request: Request<GatewayLocationReqV1>,
+        request: Request<GatewayLocationReqV1>,
     ) -> GrpcResult<GatewayLocationResV1> {
-        unimplemented!()
+        // Should this rpc be admin-authorized only or should a requesting pubkey
```
michaeldjeffrey, line 42: My understanding right now is only HPR's will be requesting gateway locations.

iot_config/src/main.rs
```diff
@@ -61,7 +62,7 @@ impl Daemon {
         poc_metrics::start_metrics(&settings.metrics)?;
 
         // Create database pool
-        let pool = settings.database.connect(10).await?;
+        let pool = settings.database.connect(50).await?;
```
michaeldjeffrey, line 65: Is this solving something? Should we make it configurable?

iot_config/src/route.rs
```diff
@@ -309,6 +308,13 @@ pub async fn delete_euis(
             break;
         }
     }
+
+    let id = Uuid::try_parse(id)?;
+    sqlx::query(" delete from route_eui_pairs where route_id = $1 ")
```
michaeldjeffrey, line 313: I'm not sure which is the correct order here. 

If we remove EUIs from HPR _then_ delete, if the delete query fails (don't know why it would, but...) we would be in a state where routing for EUIs would stop until any HPR in a region with one of those devices has to reconnect. 

If we delete first _then_ remove from HPR, we won't be removing anything from routing that could still exist in the DB. And if the removes fail for some reason, HPRs wouldn't receive them on a reconnect.

iot_config/src/route_service.rs
```diff
@@ -220,13 +290,22 @@ impl iot_config::Route for RouteService {
...
+        self.verify_request_signature(&request, OrgId::RouteId(&request.route_id))
+            .await?;
 
+        let pool = self.pool.clone();
         let (tx, rx) = tokio::sync::mpsc::channel(20);
+
         tokio::spawn(async move {
-            for eui in euis {
+            let euis = match route::list_euis_for_route(&request.route_id, &pool)
+                .map_err(|_| Status::internal("failed retrieving eui pairs"))
```
michaeldjeffrey, line 301: Should we be logging these types of errors?

iot_config/src/main.rs
jeffgrunewald: i wouldn't bother in this case; @bbalser is coming through and adjusting all of the database connections to allow for connecting with either postgres password _or_ aws iam role and this should get sorted out then

iot_config/src/gateway_service.rs
jeffgrunewald: @madninja and i are discussing whether or not this needs to become a more generic `metadata` method for retrieving any on-chain data about a hotspot, inclusive of the asserted location but that would be either primarily or exclusively for the other oracles so between that and the HPRs still under the umbrella of "network infrastructure" so likely authenticated but we don't currently authenticate any requests the oracles make (they only request from their host-local blockchain-node instance) so request signing would be a new requirement for them

iot_config/src/route.rs
jeffgrunewald: so treating this as a true "stream" from the database we more or less _have_ to remove from HPR before deleting them from the DB b/c if we did it in the reverse order we'd be deleting the records while there were still being streamed out of the DB; the previous impl only works b/c it's loading them all into memory at once before issuing the DB delete

however, after the changes we discussed on the HPR side where you're more durably persisting eui pairs and devaddr ranges and only adding/removing routes when they become enabled or disabled, i _think_ switching the order here is the correct way to process things in a failure scenario. since this RPC is serialized and won't return to the caller until the entire operation completes if the request fails midway during HPR eui removal but before deleting from the DB then the caller will receive the failure response and can re-request the deletion. if the caller received the `Ok` response that should signal that the full list was streamed for removal to the HPRs _and_ the db deletion completed successfully. if it were somehow possible to delete first and then stream for removal without deleting the source out from under the stream as you suggested the stream operation could fail after the DB delete succeeded and re-requesting would just return a no-op response b/c those eui pairs would already be gone from the DB and there wouldn't be anything to re-trigger the HPR removal stream to clean up routing

bbalser (approved):

iot_config/src/route_service.rs
jeffgrunewald: Change of plans: we’re returning them to the caller to decide what to do with them. We can also log if we decide we need to

iot_config/src/route.rs
jeffgrunewald: We just removed the `delete all` apis which makes this a moot point

PR #399, opened 2023-03-02
setup verifier for post current l1

andymck:
IOT verifier changes require to support the switch to solana.

- Add support for data transfer rewards
- Replaces blockchain node with new iot config client
- Drops invalid witnesses from final 14 selected, removes filtering out any invalids other than self witnesses
- Adds region param cache, avoid unnecessary RPC requests
- Adds iot price tracker support
- Adds iot valid packet loading from s3

Makes certain assumptions re the server side implementation of gateway info related RPCs
Depends on https://github.com/helium/proto/pull/296


iot_verifier/src/tx_scaler.rs
```diff
@@ -65,30 +74,31 @@ impl Server {
...
-            if let Some(h3index) = location {
-                if active_gateways.contains_key(&address) {
+        let mut gw_stream = self.iot_config_client.active_gateways().await?;
+        while let Some(gateway_info) = gw_stream.next().await {
+            if let Some(h3index) = gateway_info.location {
+                if active_gateways.contains_key(&gateway_info.address.as_ref().to_vec()) {
                     global_map.increment_unclipped(h3index)
                 }
             }
+            _ = gateway_cache.insert(gateway_info).await;
```
jeffgrunewald, line 101: i applaud the efficiency of using this operation to warm the cache even as i hate that it's happening in a module that's not _obviously_ supposed to be doing that 😆

iot_verifier/src/gateway_cache.rs
```diff
@@ -14,42 +14,53 @@ const CACHE_TTL: Duration = Duration::from_secs(12 * (60 * 60));
...
+    GatewayNotFound(PublicKeyBinary),
+    #[error("error querying iot config service")]
+    IotConfigClient(#[from] IotConfigClientError),
+}
 
 impl GatewayCache {
-    pub fn from_settings(settings: &Settings) -> Self {
-        let follower_service = FollowerService::from_settings(&settings.follower);
+    pub fn from_settings(settings: &Settings) -> Result<Self, GatewayCacheError> {
+        let iot_config_client = IotConfigClient::from_settings(&settings.iot_config_client)?;
```
jeffgrunewald, line 32: i wonder if this is going to cause any problems with the GatewayCache having an IotConfigClient since you're passing the gateway cache into the DensityScaler which _also_ has its own IotConfigClient being init'd from settings

jeffgrunewald: perhaps you can init the config client in the main run function and then clone it and pass it to both processes?

iot_config_client/src/region_params_resolver.rs
```diff
@@ -0,0 +1,43 @@
+use crate::iot_config_client::IotConfigClientError;
+use async_trait::async_trait;
+use helium_proto::{
+    services::iot_config::RegionParamsResV1 as RegionParamsProto, BlockchainRegionParamV1, Region,
```
jeffgrunewald, line 4:
```suggestion
    services::iot_config::GatewayRegionParamsResV1 as RegionParamsProto, BlockchainRegionParamV1, Region,
```

iot_config_client/src/iot_config_client.rs
```diff
@@ -0,0 +1,114 @@
...
+        let keypair = settings.keypair()?;
+        Ok(Self {
+            keypair: Arc::new(keypair),
+            gateway_client: settings.connect_iot_gateway_config(),
+            region_client: settings.connect_iot_region_config(),
+            batch_size: settings.batch,
+        })
+    }
+
+    pub async fn active_gateways(&mut self) -> Result<GatewayInfoStream, IotConfigClientError> {
```
jeffgrunewald, line 97: can we rename this function to something like `info_stream` or `gateway_stream`? it used to return `active_gateways` when we wrote the original stream coming from the blockchain-node since we could do the "active" filtering from the blockchain side but since we can't do that now and we only get the "active" status after comparing to the records in the verifier's DB this name seems like it will just add confusion anymore

iot_verifier/src/region_cache.rs
```diff
@@ -0,0 +1,72 @@
...
+    region_params_resolver::{RegionParamsInfo, RegionParamsResolver},
+};
+use retainer::Cache;
+use std::time::Duration;
+
+const CACHE_TTL: u64 = 86400;
+
+pub struct RegionCache {
+    pub iot_config_client: IotConfigClient,
+    pub cache: Cache<ProtoRegion, RegionParamsInfo>,
```
jeffgrunewald, line 15: can this just be `Region` if we don't need to distinguish it between the proto struct and a locally defined struct?

iot_verifier/src/region_cache.rs
```diff
@@ -0,0 +1,72 @@
...
+}
+
+#[derive(thiserror::Error, Debug)]
+#[error("error creating region cache: {0}")]
+pub struct NewGatewayCacheError(#[from] db_store::Error);
```
jeffgrunewald, line 20: this doesn't look to be in use anywhere

iot_config_client/src/gateway_resp.rs
```diff
@@ -0,0 +1,64 @@
...
+#[async_trait]
+pub trait GatewayInfoResolver {
+    async fn resolve_gateway_info(&mut self, address: &PublicKeyBinary) -> Result<GatewayInfo>;
+}
+
+#[derive(Debug, Clone)]
+pub struct GatewayInfo {
+    pub location: Option<u64>,
+    pub address: Vec<u8>,
+    pub owner: Vec<u8>,
```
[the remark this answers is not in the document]
andymck: looks like you reviewed an earlier draft, address is already a PublicKeyBinary

file_store/src/traits/msg_verify.rs
```diff
@@ -47,7 +47,6 @@ impl_msg_verify!(RouteListReqV1, signature);
 impl_msg_verify!(RouteGetReqV1, signature);
 impl_msg_verify!(RouteCreateReqV1, signature);
 impl_msg_verify!(RouteUpdateReqV1, signature);
-impl_msg_verify!(RouteDeleteReqV1, signature);
```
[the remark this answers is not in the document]
andymck: put back in here: 7f7555aab495ad2f224496c8994d4058104482b6
and confirmed working now...not sure why it was failing originally...maybe my proto branch was stale or the protos were in flux

iot_config_client/src/iot_config_client.rs
andymck: changed here: 7f7555aab495ad2f224496c8994d4058104482b6

iot_config_client/src/settings.rs
```diff
@@ -0,0 +1,58 @@
...
+            .connect_lazy();
+        iot_config::GatewayClient::new(channel)
+    }
+    pub fn connect_iot_region_config(&self) -> iot_config::admin_client::AdminClient<Channel> {
+        let channel = Endpoint::from(self.url.clone())
+            .connect_timeout(Duration::from_secs(self.connect))
+            .timeout(Duration::from_secs(self.rpc))
+            .connect_lazy();
+        iot_config::admin_client::AdminClient::new(channel)
+    }
```
jeffgrunewald, line 53:
```suggestion
    pub fn connect_config_clients(&self) -> (iot_config::GatewayClient<Channel>, iot_config::AdminClient<Channel>) {
        let channel = Endpoint::from(self.url.clone())
            .connect_timeout(Duration::from_secs(self.connect))
            .timeout(Duration::from_secs(self.rpc))
            .connect_lazy();
        (iot_config::GatewayClient::new(channel.clone()), iot_config::AdminClient::new(channel))
    }
```
since we're connecting to both anyway, would something like this work?

iot_config_client/src/iot_config_client.rs
```diff
@@ -0,0 +1,114 @@
...
+}
+
+impl IotConfigClient {
+    pub fn from_settings(settings: &Settings) -> Result<Self, IotConfigClientError> {
+        let keypair = settings.keypair()?;
+        Ok(Self {
+            keypair: Arc::new(keypair),
+            gateway_client: settings.connect_iot_gateway_config(),
+            region_client: settings.connect_iot_region_config(),
+            batch_size: settings.batch,
```
jeffgrunewald, line 93:
```suggestion
        let (gateway_client, region_client) = settings.connect_config_clients();
        Ok(Self {
            keypair: Arc::new(keypair),
            gateway_client,
            region_client,
            batch_size: settings.batch,
```
see below for the other half of this suggestion

iot_config_client/src/iot_config_client.rs
```diff
@@ -0,0 +1,114 @@
...
+        })
+    }
+
+    pub async fn gateway_stream(&mut self) -> Result<GatewayInfoStream, IotConfigClientError> {
+        let req = GatewayInfoStreamReqV1 {
+            batch_size: self.batch_size,
+            signature: vec![], // todo: are we signing requests or not?
```
jeffgrunewald, line 100: we are; especially the streaming one which would be bad if abused

iot_verifier/src/gateway_cache.rs
jeffgrunewald: not sure how tonic handles connections under the hood; since http2 is multiplexed i think you could wrap the underlying h2 connection in an Arc and share it between clients in which case cloning that is just cloning a smart pointer. if it's just a deep clone of the full connection though you're right that it wouldn't be an issue

iot_config_client/src/region_params_resolver.rs
jeffgrunewald: you're right; i forgot you're working off a branch of proto here so i just saw a reference to a message that didn't exist on master and was confused how this was even compiling; carry on!

jeffgrunewald:
looks like CI is currently failing due to a version problem with one of the rand crates (probably being pulled in transitively via helium-crypto/rustcrypto?) that I recall @maplant mentioning somewhere in discord a week ago

iot_verifier/src/gateway_cache.rs
andymck: i understand your point now, you are questioning whether we end up with two connections to the same server in the way i have it.   I would seriously doubt that is the case but cannot say for sure.  I will look into.

Tonic does recommend that connections are cloned in order to achieve concurrency, that will be via grpc streams which are multiplexed over the same connection.

iot_config_client/src/iot_config_client.rs
andymck: yeah the other requests are already signed, i forgot about this one

andymck: fixed

iot_verifier/src/gateway_cache.rs
andymck: addressed here: 9f1d2fbd2a617fea9c1d3429ddb6533caf848800

jeffgrunewald: I think you can still use same tonic Channel in the gateway cache, the region cache, and the density_scaler if you create a single instance of the IotConfigClient in main.rs with `IotConfigClient::from_settings(&settings.iot_config_client)` and then clone _that_ and pass it as the starting argument to `GatewayCache::from_settings()`, `RegionCache::from_settings()` and `DensityScaler::from_settings()` (the caches' could actually become `::new()` functions instead since the only thing they're doing with their settings is starting the config client

iot_verifier/src/tx_scaler.rs
```diff
@@ -75,17 +76,14 @@ impl Server {
         let refresh_start = Utc::now();
         tracing::info!("density_scaler: generating hex scaling map, starting at {refresh_start:?}");
         let mut global_map = GlobalHexMap::new();
-        let active_gateways = self
+        let gateways = self
```
jeffgrunewald, line 79: this could still stay `active_gateways` since it's the list of gateways that beaconed within the verifier's db within the inactivity threshold window

iot_verifier/migrations/10_packet.sql
```diff
@@ -0,0 +1,7 @@
+create table packet (
+    id serial primary key,
```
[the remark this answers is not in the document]
andymck: i am not sure but the max over a 7day period in feb amounted to 82m packets.  As we will be clearing these out ever reward period ( so max 24hours ) i32 likely gives plenty of headroom.  That said we dont loose much by defaulting to bigserial

iot_verifier/migrations/10_packet.sql
```diff
@@ -0,0 +1,7 @@
...
+    payload_size bigint not null,
+    gateway bytea not null,
```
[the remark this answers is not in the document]
andymck: verifier has always stored pubkeys are bytea

iot_verifier/src/poc.rs
```diff
@@ -400,7 +414,7 @@ fn verify_entropy(
 fn verify_beacon_payload(
     beacon_report: &IotBeaconReport,
     region: ProtoRegion,
-    region_params: &[BlockchainRegionParamV1],
+    region_params: &Vec<BlockchainRegionParamV1>,
```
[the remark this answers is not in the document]
andymck: yeah, not sure why i changed that

iot_verifier/src/packet.rs
```diff
@@ -0,0 +1,22 @@
+pub async fn insert<'c, E>(
+    executor: E,
+    payload_size: i32,
+    gateway: Vec<u8>,
+    payload_hash: Vec<u8>,
+) -> anyhow::Result<()>
```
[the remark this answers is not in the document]
andymck: this module is all very much WIP, i wouldnt pay any attention to it atm

iot_verifier/migrations/10_packet.sql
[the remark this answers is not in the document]
jeffgrunewald: if you’re just deleting the table contents every 24 hours I believe that will _not_ reset the serial integer since Postgres keeps track of the `last_value` in a separate associate `packet_seq` table and increment that with every record generated. So you’ll need to drop and recreate both tables periodically or perform special SEQUENCE update queries to reset the serial number back to 1 each time you drop the packet table contents

iot_config_client/src/settings.rs
```diff
@@ -0,0 +1,46 @@
...
+    /// RPC timeout for follower in seconds. Default 5
+    #[serde(default = "default_rpc_timeout")]
+    pub rpc: u64,
+    /// batch size for gateway stream results. Default 100
+    #[serde(default = "default_batch_size")]
+    pub batch: u32,
+}
+
+pub fn default_url() -> http::Uri {
+    http::Uri::from_static("http://127.0.0.1:8080")
```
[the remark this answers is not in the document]
andymck: removed

iot_packet_verifier/src/verifier.rs
```diff
@@ -475,15 +479,17 @@ mod test {
         assert_eq!(
             valid_packets,
             vec![
+                // TODO: some weird shit with the timestamp values in the tests
```
jeffgrunewald, line 482: i've never had luck testing things where timing is variable without setting a pre-determined timestamp at the beginning of the operation

iot_verifier/src/reward_share.rs
```diff
@@ -205,28 +351,65 @@ impl GatewayShares {
...
+    let iot_value = iot_amount * iot_price;
+    (iot_value / (*DC_USD_PRICE) / dec!(1_000_000))
+        .round_dp_with_strategy(0, RoundingStrategy::ToNegativeInfinity)
+}
+
+pub fn normalize_dc_transfer_rewards(
+    total_dc_transfer_rewards_used: Decimal,
+    total_dc_transfer_rewards: Decimal,
+) -> (Decimal, Decimal) {
+    match total_dc_transfer_rewards_used < total_dc_transfer_rewards {
```
jeffgrunewald, line 392: any specific reason this _shouldn't_ be `<=`? we won't have any dangling fractional amounts unaccounted for by leaving that scenario unaccounted for?

iot_verifier/src/reward_share.rs
```diff
@@ -61,8 +99,11 @@ impl GatewayShare {
             r#"
             insert into gateway_shares (hotspot_key, reward_type, reward_timestamp, hex_scale, reward_unit, poc_id)
             values ($1, $2, $3, $4, $5, $6)
-            on conflict (hotspot_key, poc_id) do update set
-            reward_type = EXCLUDED.reward_type, reward_timestamp = EXCLUDED.reward_timestamp, hex_scale = EXCLUDED.hex_scale, reward_unit = EXCLUDED.reward_unit
+            on conflict (id) do update set
```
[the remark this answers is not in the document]
andymck: artifact from when i had the pocs and data rewards combined into single table, addressed here: 6b9ee2328eae198a9a1e7426b775a8f7ca85bf92

iot_verifier/src/reward_share.rs
```diff
@@ -102,31 +143,78 @@ impl GatewayShare {
...
 }
 
+impl GatewayDCShare {
+    pub async fn save(
+        self,
+        db: &mut Transaction<'_, Postgres>,
+    ) -> Result<bool, SaveGatewayShareError> {
+        Ok(sqlx::query_as::<_, GatewayShareSaveResult>(
+            r#"
+            insert into gateway_dc_shares (hotspot_key, reward_type, reward_timestamp, num_dcs, id)
```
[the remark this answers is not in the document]
andymck: yeah no need to do this, addressed here: 6b9ee2328eae198a9a1e7426b775a8f7ca85bf92

iot_verifier/src/reward_share.rs
```diff
@@ -102,31 +143,78 @@ impl GatewayShare {
...
-            ReportType::Beacon => self.beacon_shares += rewards,
-            ReportType::Witness => self.witness_shares += rewards,
+            ReportType::Beacon => {
+                let rewards = share.hex_scale * share.reward_unit;
+                self.beacon_shares += rewards
+            }
+            ReportType::Witness => {
+                let rewards = share.hex_scale * share.reward_unit;
+                self.witness_shares += rewards
+            }
```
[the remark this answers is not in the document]
andymck: addressed here: 6b9ee2328eae198a9a1e7426b775a8f7ca85bf92

iot_verifier/src/reward_share.rs
andymck: addressed here: 6b9ee2328eae198a9a1e7426b775a8f7ca85bf92

iot_verifier/src/reward_share.rs
```diff
@@ -61,8 +99,11 @@ impl GatewayShare {
             r#"
             insert into gateway_shares (hotspot_key, reward_type, reward_timestamp, hex_scale, reward_unit, poc_id)
             values ($1, $2, $3, $4, $5, $6)
-            on conflict (hotspot_key, poc_id) do update set
-            reward_type = EXCLUDED.reward_type, reward_timestamp = EXCLUDED.reward_timestamp, hex_scale = EXCLUDED.hex_scale, reward_unit = EXCLUDED.reward_unit
+            on conflict (poc_id) do update set
```
jeffgrunewald, line 102:
```suggestion
            on conflict (hotspot_key, poc_id) do update set
```
i think this has to stay a conflict check against the entire `(hotspot_key, poc_id)` composite key because it's valid for two hotspots to have reward records in the database with with the same `poc_id` within the same reward period, one being for beaconing and any number for witnessing. because of that, this table also uses the composite of the hotspot key and the poc id as the primary key so not checking against both could inadvertently cause a collision if we did attempt to insert a "duplicate" record that didn't have a conflict resolution defined

jeffgrunewald (commented):
One issue I foresee with the gateway shares entries for PoC rewards but otherwise this looks ready to go

iot_verifier/src/reward_share.rs
andymck: darn, i was meant to put this entire sql sequence back to its original form.  Will fix...

iot_verifier/src/reward_share.rs
```diff
@@ -102,30 +143,66 @@ impl GatewayShare {
...
 
 impl RewardShares {
-    pub fn add_reward(&mut self, share: &GatewayShare) {
+    pub fn add_poc_reward(&mut self, share: &GatewayPocShare) {
         let rewards = share.hex_scale * share.reward_unit;
-        match share.reward_type {
-            ReportType::Beacon => self.beacon_shares += rewards,
-            ReportType::Witness => self.witness_shares += rewards,
-        }
+        self.beacon_shares += rewards
```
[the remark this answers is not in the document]
andymck: yeah had a brainfart. gonna remove `Packet` from ReportType enum altogether, its no longer necessary with the sep dc shares table

iot_verifier/src/reward_share.rs
andymck: addressed here: 95abec833a8dc0eba6ff9cf1fe4c961054d88d5a

jeffgrunewald (approved):

bbalser (approved):

PR #400, opened 2023-03-02
Add support for token price oracle

vihu:
- Add price oracle skeleton
- Update file_info
- Add settings
- Update Cargo.toml and Cargo.lock


price/Cargo.toml
```diff
@@ -0,0 +1,32 @@
...
+metrics-exporter-prometheus = { workspace = true }
+tokio = { workspace = true }
+chrono = { workspace = true }
+helium-proto = { workspace = true }
+file-store = { path = "../file_store" }
+poc-metrics = { path = "../metrics" }
+triggered = {workspace = true}
+pyth-sdk-solana = "0.7.1"
+solana-client = ">= 1.9, < 1.15"
+solana-program = ">= 1.9, < 1.15"
```
madninja, line 32: Do you need all of these?

price/docker-compose.yml
```diff
@@ -0,0 +1,67 @@
+version: "2.4"
```
madninja, line 1: Remove? I don't think we have docker compose for any of the other oracles

price/pkg/settings-template.toml
```diff
@@ -0,0 +1,56 @@
...
+# log = "price=debug"
+
+# RPC Endpoint for price oracles. Required.
+rpc_endpoint = "https://api.devnet.solana.com"
+
+# Price sink roll time (mins). Default = 3 mins. Optional.
+sink_roll_mins = 3
+
+# Price tick interval (secs). Default = 60s. Optional.
+tick_interval = 60
```
madninja, line 12: Rename to `interval`

price/pkg/settings-template.toml
```diff
@@ -0,0 +1,56 @@
+# log settings for the application (RUST_LOG format). Default below
+# 
+# log = "price=debug"
+
+# RPC Endpoint for price oracles. Required.
+rpc_endpoint = "https://api.devnet.solana.com"
+
+# Price sink roll time (mins). Default = 3 mins. Optional.
+sink_roll_mins = 3
```
madninja, line 9: move to constant.. it's not done like this in other oracles

price/pkg/settings-template.toml
```diff
@@ -0,0 +1,56 @@
+# log settings for the application (RUST_LOG format). Default below
+# 
+# log = "price=debug"
+
+# RPC Endpoint for price oracles. Required.
+rpc_endpoint = "https://api.devnet.solana.com"
```
madninja, line 6: Rename to source. 

In general config keys with a `_` in it mess with the ability to override them in env variables

price/src/error.rs
```diff
@@ -0,0 +1,9 @@
+#[derive(thiserror::Error, Debug)]
+pub enum PriceError {
+    #[error("unsupported token type: {0}")]
+    UnsupportedTokenType(i32),
+    #[error("unable to fetch price")]
+    UnableToFetch,
```
madninja, line 6: Is there an underlying error when UnableToFetch hits?

price/src/main.rs
```diff
@@ -0,0 +1,156 @@
...
+        let hst_watch = hst_price_generator.receiver();
+
+        // price service
+        let hnt_svc = PriceService::new(hnt_watch)?;
+        let mobile_svc = PriceService::new(mobile_watch)?;
+        let iot_svc = PriceService::new(iot_watch)?;
+        let hst_svc = PriceService::new(hst_watch)?;
+
+        // price servers
+        let hnt_server = PriceServer::new(hnt_svc);
```
madninja, line 96: Remove all servers and tonic dependencies.. The verifiers will get the price data from the bucket, remember?

price/src/price_service.rs
```diff
@@ -0,0 +1,42 @@
+use crate::{price_generator::MessageReceiver, GrpcResult};
```
madninja, line 1: Delete this file since we're not exposing GRPC services

price/pkg/settings-template.toml
madninja: This will need to be overridden to one of the better endpoints during deployment

price/docker-compose.yml
jeffgrunewald: i would remove this, but only because this is already available in the root of the project (we do have a single docker-compose file that stands up the iot-config service, minio with all the right buckets and a postgres database with all the right tables
you _can_ stand up all of the things at once by just typing `docker-compose up -d` but by chaining the depends_on key correctly you can just stand up the one or more services you care about by naming it as an argument to the command e.g. `docker-compose up -d iot-config`

vihu: Removed in [a454294](https://github.com/helium/oracles/pull/400/commits/a454294be181d1c7e47fa93265e510712be6cdc0)

price/src/main.rs
vihu: Removed in [a454294](https://github.com/helium/oracles/pull/400/commits/a454294be181d1c7e47fa93265e510712be6cdc0)

price/src/price_service.rs
vihu: Removed in [a454294](https://github.com/helium/oracles/pull/400/commits/a454294be181d1c7e47fa93265e510712be6cdc0)

price/src/error.rs
vihu: Doesn't look like it, `get_price_no_older_than` just returns None afaict.

price/src/price_service.rs
```diff
@@ -0,0 +1,26 @@
...
+
+pub struct PriceService {
+    price_watch: MessageReceiver,
+}
+
+impl PriceService {
+    pub fn new(price_watch: MessageReceiver) -> anyhow::Result<Self> {
+        Ok(Self { price_watch })
+    }
+}
```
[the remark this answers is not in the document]
vihu: This is not required I think, the price generator itself keeps track of the last price fetched, in case of an rpc failure we just update the timestamp and return that.

price/pkg/settings-template.toml
vihu: Fixed in [fe83739](https://github.com/helium/oracles/pull/400/commits/fe837396d363ec7094be174464b50e8b89359488)

price/pkg/settings-template.toml
vihu: Fixed in [fe83739](https://github.com/helium/oracles/pull/400/commits/fe837396d363ec7094be174464b50e8b89359488)

price/Cargo.toml
vihu: Yeah, client for the RpcClient and program for the pubkey. pyth-sdk doesn't seem to automatically fetch these.

file_store/src/cli/dump.rs
```diff
@@ -163,8 +164,14 @@ impl Cmd {
...
-                        "timestamp": packet_report.gateway_timestamp_ms,
-
+                        "timestamp": packet_report.gateway_timestamp_ms}))?;
+                }
+                FileType::PriceReport => {
+                    let manifest = PriceReportV1::decode(msg)?;
+                    print_json(&json!({
+                        "price": manifest.price,
+                        "timestamp": manifest.timestamp,
+                        "token_type": BlockchainTokenTypeV1::from_i32(manifest.token_type),
```
jeffgrunewald, line 174: is `from_i32` an infallible function call? i thought those all had to be result types

price/src/price_generator.rs
```diff
@@ -0,0 +1,176 @@
...
+                "got price {:?} via rpc for {:?} at {:?}",
+                value,
+                token_type,
+                current_time
+            );
+            Ok(Price {
+                price: value as u64,
+                timestamp: current_time,
+                token_type,
+            })
```
jeffgrunewald, line 173:
```suggestion
            Ok(Price::new(current_time, value as u64, token_type))
```

price/src/main.rs
```diff
@@ -0,0 +1,119 @@
...
+                .map_err(Error::from),
+            mobile_price_generator
+                .run(price_sink.clone(), &shutdown, BlockchainTokenTypeV1::Mobile)
+                .map_err(Error::from),
+            iot_price_generator
+                .run(price_sink.clone(), &shutdown, BlockchainTokenTypeV1::Iot)
+                .map_err(Error::from),
+            hst_price_generator
+                .run(price_sink, &shutdown, BlockchainTokenTypeV1::Hst)
+                .map_err(Error::from),
```
jeffgrunewald, line 107: since these generators are just going to be reaching a `None` else case when they check for a pricing account key do we even want to run them before those tokens have prices to be fetched? I would see running them if we were going to default to writing out a $0 pricing report but since we're not it seems like we could skip them.

for that matter, and since we are _only_ accessing these prices via the written reports would it be easier to have a single `PriceGenerator` struct that loops over all "active" tokens based on configuration, stores _only_ the last price in a map in its state and spawns a thread with a join handle for each active token, passing the previous price to the thread and A) checking for a newer price, B) writing out the report to the file sink with the current timestamp and C) returning either the previous or newer price via the join handle to the PriceGenerator?

then you only need to grab current time once and it doesn't need to be mutable, you simplify the state of the PriceGenerator process, there's only one continuous thread you need to worry about shutting down and it's just a matter of adding a token string to a config list in Settings to start checking for additional prices

file_store/src/cli/dump.rs
bbalser: I think `manifest.token_type()` will return the enum.

jeffgrunewald: Oh yeah; I keep forgetting about those functions

price/src/main.rs
vihu: Yeah we thought about having a single PriceGenerator, but it doesn't really matter if the timestamps for different token types (as they become available) within the reports are the same as long as they are there. I believe it's already pretty simple as is imo and the state is minimal.

price/src/price_generator.rs
vihu: Updated [f8ab0b3](https://github.com/helium/oracles/pull/400/commits/f8ab0b3f81d8ddf7315cdd800beaa15b769ab9af)

file_store/src/cli/dump.rs
vihu: Yup, ty. Updated [f8ab0b3](https://github.com/helium/oracles/pull/400/commits/f8ab0b3f81d8ddf7315cdd800beaa15b769ab9af)

price/src/settings.rs
```diff
@@ -0,0 +1,133 @@
...
+            BlockchainTokenTypeV1::Mobile => self
+                .cluster
+                .mobile_price_key
+                .as_ref()
+                .map(|key| SolPubkey::from_str(key).expect("unable to parse")),
+            BlockchainTokenTypeV1::Iot => self
+                .cluster
+                .iot_price_key
+                .as_ref()
+                .map(|key| SolPubkey::from_str(key).expect("unable to parse")),
```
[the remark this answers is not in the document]
vihu: Updated with panic!, cc: @madninja cuz panic! Open to other suggestions instead of panicking but as @jeffgrunewald said we want to catch this ASAP and fix it when tokens come alive. [f8ab0b3](https://github.com/helium/oracles/pull/400/commits/f8ab0b3f81d8ddf7315cdd800beaa15b769ab9af)

price/README.md
```diff
@@ -0,0 +1,19 @@
...
+- IOT
+
+Note that currently only HNT-USD prices are available.
+
+The price oracle server:
+
+- Requests price for HNT token at a regular interval (60s) from pyth via solana
+  RpcClient. In case of failure it uses the previously fetched price and stores
+  the same with an updated timestamp.
+- Stores and uploads [price_report](TODO) to an S3 bucket.
```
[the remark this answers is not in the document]
vihu: Fixed [f8ab0b3](https://github.com/helium/oracles/pull/400/commits/f8ab0b3f81d8ddf7315cdd800beaa15b769ab9af)

price/Cargo.toml
vihu: Removing just the solana-program doesn't seem to work either.

price/src/main.rs
andymck: Does it matter if one  price gets successfully updated and others do not? ie is there any intra dependencies on the prices being current on the consumer side ? If iot gets updated but for whatever reason hnt fails to update and continues failing for a while and whilst there may be steep changes in the hnt price

vihu: That's a good question but I'd say not something we have to solve for first pass at least. So in the scenario where HNT updated and IOT failed, we'll presumably continue to serve the same data for IOT till we:

a) either have some redundancy with our RPC endpoints (assuming it was an RPC failure to fetch IOT in the above example)
b) or we kind of "derive" IOT price from HNT in some way

@madninja What are your thoughts on this? I could look into adding support for multiple RPC endpoints I suppose. And _maybe_ do a mean of the prices fetched from multiple RPC endpoints?

bbalser (approved):

jeffgrunewald (approved):

PR #403, opened 2023-03-07
Add PriceTracker to Price Oracle

bbalser:


file_store/src/price_tracker.rs
```diff
@@ -0,0 +1,83 @@
+use crate::{Error, FileStore, FileType, Result, Settings};
+use chrono::{Duration, Utc};
+use futures::stream::{StreamExt, TryStreamExt};
+use helium_proto::{services::price_oracle::PriceOracleReportV1, BlockchainTokenTypeV1, Message};
```
jeffgrunewald, line 4: needs updated to latest report structure

file_store/src/price_tracker.rs
```diff
@@ -0,0 +1,83 @@
...
+
+        let prices_clone = prices.clone();
+        let settings_clone = settings.clone();
+        let handle = tokio::spawn(async move { run(settings_clone, prices_clone, shutdown).await });
+
+        Ok((PriceTracker { prices }, async move {
+            match handle.await {
+                Ok(Ok(())) => Ok(()),
+                Ok(Err(err)) => Err(err),
+                Err(err) => Err(Error::from(err)),
```
jeffgrunewald, line 29: for `Error::from` not return an `Err(value)` such that you still have to wrap it in an error result value?

file_store/src/price_tracker.rs
```diff
@@ -0,0 +1,83 @@
...
+            .await
+            .get(&token_type)
+            .cloned()
+            .ok_or(Error::PriceNotAvailable())
+    }
+}
+
+async fn run(settings: Settings, prices: Prices, shutdown: triggered::Listener) -> Result {
+    let file_store = FileStore::from_settings(&settings).await?;
+    let mut trigger = tokio::time::interval(std::time::Duration::from_secs(30));
```
jeffgrunewald, line 46: (feels like we need to check every time anymore) is the default tick behavior what we want here?

file_store/src/price_tracker.rs
```diff
@@ -0,0 +1,83 @@
...
+                    file_store.stream_file(file)
+                        .await?
+                        .map_err(Error::from)
+                        .and_then(|buf| async { PriceOracleReportV1::decode(buf).map_err(Error::from) })
+                        .map_err(|err| {
+                            tracing::warn!("skipping price report due to error {err:?}");
+                            err
+                        })
+                        .filter_map(|result| async { result.ok() })
+                        .map(|report| (report, prices.clone()))
```
jeffgrunewald, line 72: can you combine the filter map and the map together?

file_store/src/price_tracker.rs
```diff
@@ -0,0 +1,83 @@
...
+                tracing::info!("PriceTracker shutting down");
+                break;
+            }
+            _ = trigger.tick() => {
+                let files = file_store.list_all(FileType::PriceReport, after, None).await?;
+
+                for file in files {
+                    tracing::debug!("processing pricing report file {}", file.key);
+                    after = file.timestamp;
+                    file_store.stream_file(file)
```
jeffgrunewald, line 63: did we drop the "source unordered" from the file streams so these will always stream in oldest to newest and this mutating timestamp will be accurately moving forward with each file?

file_store/src/price_tracker.rs
bbalser: Default is burst and that seems like the correct one to me.

file_store/src/price_tracker.rs
bbalser: I can but it is easier to read as separate steps so my vote is leave it as is

file_store/src/price_tracker.rs
bbalser: This does not process files out of order or concurrently so we will always end up with latest prices

jeffgrunewald:
looks like this branch had an older commit from `rg/price` force-pushed to it; there are a bunch of small things that have been changed on that branch as a result of PR feedback that are still in play here so this should get rebased on the latest commit there unless it gets merged into main first

madninja (commented):
I don't quite get why price_tracker.rs is in the file_store at all.. We have none of the other services declare types here so why this one?

jeffgrunewald:
> I don't quite get why price_tracker.rs is in the file_store at all.. We have none of the other services declare types here so why this one?

i wondered about that; seems like a reasonable if imprecise place to put it since it's technically just a client for streaming price_report files and there was little enough code required that it seems overkill to put it in its own crate

madninja:
> > > I don't quite get why price_tracker.rs is in the file_store at all.. We have none of the other services declare types here so why this one?
> > 
> > 
> > i wondered about that; seems like a reasonable if imprecise place to put it since it's technically just a client for streaming price_report files and there was little enough code required that it seems overkill to put it in its own crate
> 
> I originally created a new `price_tracker` crate, but it felt like overkill, but I can go back to that if we feel that is a better home.

yes please

file_store/src/price_tracker.rs
bbalser: `Error::from` does not return a `Result` just the actual error, so it still needs to be made into a `Result` or do I not understand the question?

jeffgrunewald (approved):

price/src/price_tracker.rs
```diff
@@ -0,0 +1,135 @@
...
+                let timestamp = process_files(&file_store, &sender, after).await?;
+                after = timestamp.unwrap_or(after);
+            }
+        }
+    }
+
+    Ok(())
+}
+
+async fn get_initial_prices(
```
andymck, line 72: nit: the `get` in 'get_initial_prices' suggests that the function will be returning price information but instead a timestamp is returned.  I get that it is actually working out the initial prices but maybe renaming to calc_initial_prices is a better fit

andymck (approved):

price/src/price_tracker.rs
bbalser: Good call, fixed 23b8f58

price/src/price_tracker.rs
```diff
@@ -0,0 +1,135 @@
...
+}
+
+async fn calculate_initial_prices(
+    file_store: &FileStore,
+    sender: &watch::Sender<Prices>,
+) -> Result<DateTime<Utc>> {
+    tracing::debug!("PriceTracker: Updating initial prices");
+    let durations = vec![Duration::minutes(10), Duration::hours(4)];
+
+    for duration in durations {
```
maplant, line 79:
```suggestion
    for duration in [Duration::minutes(10), Duration::hours(4)] {
```

PR #405, opened 2023-03-10
Add ability to hardcode aws creds but only when file-store/local feature is enabled

bbalser:
This should enable local testing to read from an AWS bucket and write to a local minio bucket.   The hardcoded credentials will only be used if the "file-store/local" feature is enabled.

### DIFF
--- a/file_store/src/file_source.rs
+++ b/file_store/src/file_source.rs
@@ -65,7 +65,7 @@ mod test {
         // 2022-08-05 15:35:55     240363 cell_heartbeat.1658832527866.gz
         // 2022-08-05 15:36:08    6525274 cell_heartbeat.1658834120042.gz
         //
-        let file_store = FileStore::new(None, "us-east-1", "devnet-poc5g-rewards")
+        let file_store = FileStore::new(None, "us-east-1", "devnet-poc5g-rewards", None, None)
             .await
             .expect("file store");
         let stream = file_store.source(infos(&[
--- a/file_store/src/file_source.rs
+++ b/file_store/src/file_source.rs
@@ -46,7 +46,7 @@ where
 #[cfg(test)]
 mod test {
     use super::*;
-    use crate::{FileInfo, FileInfoStream, FileStore};
+    use crate::{FileInfo, FileInfoStream, FileStore, Settings};
     use std::str::FromStr;
 
     fn infos(names: &'static [&str]) -> FileInfoStream {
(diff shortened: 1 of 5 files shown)


file_store/src/file_source.rs
```diff
@@ -65,7 +65,7 @@ mod test {
         // 2022-08-05 15:35:55     240363 cell_heartbeat.1658832527866.gz
         // 2022-08-05 15:36:08    6525274 cell_heartbeat.1658834120042.gz
         //
-        let file_store = FileStore::new(None, "us-east-1", "devnet-poc5g-rewards")
+        let file_store = FileStore::new(None, "us-east-1", "devnet-poc5g-rewards", None, None)
```
maplant, line 68: Perhaps at this point a builder pattern would be better? We have quite a few of arguments that are normally going to be `None`

bbalser: Feels overkill to me at the moment, to my knowledge, we only call `new` in 2 places, this test and the `from_settings` function in file_store.rs.

file_store/src/file_store.rs
```diff
@@ -31,20 +31,40 @@ impl FileStore {
...
+        )
+        .await
     }
 
     pub async fn new(
         endpoint: Option<Endpoint>,
         region: impl ProvideRegion + 'static,
         bucket: impl Into<String>,
+        #[allow(unused_variables)] access_key_id: Option<String>,
+        #[allow(unused_variables)] secret_access_key: Option<String>,
```
jeffgrunewald, line 49: i'm guessing this whole thing would still work the same and these annotations would be unnecessary without the `local` feature cfg macro below but that's in place as an extra guard on top of just not supplying the `access_key_id` and `secret_access_key` as arguments outside of local testing?

jeffgrunewald (approved):

file_store/src/file_store.rs
madninja: this is one of the arguments for the builder pattern.. it removes the need for these allow/feature hacks?

bbalser: I removed the `FileStore::new` function since it was only used in a test and moved the functionality into `FileStore::from_settings`.  This removes the need for the allow directives.

Address in 93b33e5

madninja: nice fix

madninja (approved):

PR #406, opened 2023-03-10
Add price metrics

vihu:
Summary
----

This adds several price related metrics. Specifically a counter for update and a gauge to track the actual price value.

### DIFF
--- a/price/src/metrics.rs
+++ b/price/src/metrics.rs
@@ -1,0 +1,21 @@
+use helium_proto::BlockchainTokenTypeV1;
+
+const PRICE_UPDATE_COUNTER: &str = concat!(env!("CARGO_PKG_NAME"), "_", "price_update_counter");
+const PRICE_GAUGE: &str = concat!(env!("CARGO_PKG_NAME"), "_", "price_gauge");
+
+pub struct Metrics;
+
+impl Metrics {
+    pub fn update(token_type: BlockchainTokenTypeV1, price: f64) {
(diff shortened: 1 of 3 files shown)


bbalser (approved):

price/src/metrics.rs
```diff
@@ -0,0 +1,21 @@
...
+pub struct Metrics;
+
+impl Metrics {
+    pub fn update(token_type: BlockchainTokenTypeV1, price: f64) {
+        increment_counter(token_type);
+        set_gauge(token_type, price)
+    }
+}
+
+fn increment_counter(token_type: BlockchainTokenTypeV1) {
```
jeffgrunewald, line 15: should these only be called from within the `new price found from the pyth api` when you do a match on the `get_price` returning a result?
the observable behavior on the gauge should be unchanged but the `update_counter` seems like it's only a valuable metric if we're actually counting when the price is updated from pyth; incrementing regardless of the outcome seems like it's really just a counter of the update price handler ticking

vihu: Yeah, I too was thinking about this. I've separated the counters to be `update` and `stale`. This way on the dashboard we can run some query like "If the stale counter sums up to 5 or whatever within 5 mins, we need to trigger an alert". [09137a5](https://github.com/helium/oracles/pull/406/commits/09137a5b1a4fab756ecb8ec13bb80199698f1f37)

jeffgrunewald (approved):

PR #411, opened 2023-03-15
optimizing update queries

jeffgrunewald:
this is an attempt to optimize use of the config service database pool with regard to long-running streams as well as speed up and minimize the impact of large streaming updates to eui pair and session key filter records:

- when subscribers connect to the long-running update streams they receive all current records for eui pairs, devaddrs, routes, and session key filters and then sit in a loop receiving new updates as they arrive. here we are _explicitly dropping_ the initial results streams from the database in order to release the db pool connection when it is no longer needed by the spawned update task
- return update results from the insert/delete queries and only stream out updates that actually resulted in a change to the database
- move update results streaming out of the calling thread and into a spawned thread to return to the caller as soon as the database update has completed successfully


iot_config/src/route_service.rs
```diff
@@ -309,6 +309,10 @@ impl iot_config::Route for RouteService {
                 }
             }
 
+            drop(active_routes);
```
jeffgrunewald, line 312: i have tested this and received the desired result (all initial updates are streamed out followed by ongoing updates as they're processed by the config service) but particularly looking to @maplant or @madninja to confirm my suspicion that this should release the hold on the database pool by the long-running subscriber threads (the packet routers) so they're not continuously tying up a pool connection they'll never use after processing the initial set of results each time a packet router connects

maplant (approved):

iot_config/src/route.rs
```diff
@@ -241,43 +247,49 @@ pub async fn update_euis(
...
     let mut transaction = db.begin().await?;
 
-    if !to_add.is_empty() {
-        insert_euis(to_add, &mut transaction).await?;
-    }
+    let added_euis = if !to_add.is_empty() {
+        insert_euis(to_add, &mut transaction).await?
+    } else {
+        vec![]
+    };
```
JayKickliter, line 254: creating an empty vec is pretty low cost, so this isn't inherently bad, but it does stick out.

db_store/src/settings.rs
```diff
@@ -11,7 +11,7 @@ pub enum AuthType {
 
 #[derive(Debug, Deserialize, Clone)]
 pub struct Settings {
-    pub max_connections: u32,
+    pub connections: u32,
```
bbalser, line 14: I understand the reasoning for this change, but I would prefer we keep our settings descriptive so we can easily understand their purpose.

iot_config/src/route.rs
```diff
@@ -180,19 +183,27 @@ pub async fn update_route(
...
-        ));
+) -> Result<Vec<EuiPair>, RouteStorageError> {
+    if euis.is_empty() {
+        return Ok(vec![]);
     }
+    // We don't want to take any actions if a route_id cannot be parsed.
+    let eui_values = euis
+        .iter()
+        .map(|eui_pair| {
+            Uuid::try_parse(&eui_pair.route_id).map(|uuid| {
```
bbalser, line 194: does implementing `TryFrom` for `&EuiPair` make sense here?

jeffgrunewald: when i first thought about that i imagined putting it in `src/lora_field.rs` but that would break the encapsulation of the route ID since the only thing that actually cares that it's a Uuid is sqlx but then i figured i could just put the `From` impl in this module
and also it's pretty academic since this is not a library and the very idea of `From` sort of breaks encapsulation by intention...

bbalser (approved):

docker-compose.yml
```diff
@@ -12,14 +12,15 @@ services:
...
-      CFG_DATABASE_URL: postgres://postgres:postgres@postgres:5432/config_db
-      CFG_LISTEN: 0.0.0.0:8080
-      CFG_METRICS_ENDPOINT: 0.0.0.0:19000
-      CFG_FOLLOWER_URL: "http://node:8090"
-      CFG_NETWORK: mainnet
-      CFG_ADMIN: ${CONFIG_ADMIN_PUBKEY}
-      CFG_KEYPAIR: /config-signing-key.bin
-      CFG_LOG: info
+      CFG__DATABASE__URL: postgres://postgres:postgres@postgres:5432/config_db
+      CFG__DATABASE__MAX_CONNECTIONS: 50
```
vihu, line 16: Did you want this to be double underscore as well? `CFG__DATABASE__MAX__CONNECTIONS`?

jeffgrunewald: in this case you wouldn't; the double is _only_ between the levels of an application/crate whereas the single underscore is just part of the setting name

PR #412, opened 2023-03-15
db_store will now record metrics on db connection pool size and num_idle

bbalser:


db_store/src/settings.rs
```diff
@@ -39,14 +39,37 @@ fn default_auth_type() -> AuthType {
...
         &self,
+        app_name: impl Into<String>,
         shutdown: triggered::Listener,
     ) -> Result<(Pool<Postgres>, futures::future::BoxFuture<'static, Result>)> {
         match self.auth_type {
             AuthType::Postgres => match self.simple_connect().await {
-                Ok(pool) => Ok((pool, Box::pin(async move { Ok(()) }))),
+                Ok(pool) => Ok((
+                    pool.clone(),
+                    Box::pin(metric_tracker::start(app_name.into(), pool, shutdown).await?),
```
maplant, line 49: `metric_tracker::start` already returns a Box, so you're double boxing here. Either remove the box in the start function or here.

db_store/src/metric_tracker.rs
```diff
@@ -0,0 +1,41 @@
...
+
+    Ok(Box::pin(async move {
+        match join_handle.await {
+            Ok(()) => Ok(()),
+            Err(err) => Err(Error::from(err)),
+        }
+    }))
+}
+
+async fn run(app_name: String, pool: sqlx::Pool<sqlx::Postgres>, shutdown: triggered::Listener) {
```
maplant, line 22: Since you format the string, you don't really need to own it _I think_ (it's not across an await boundary, but there could be another issue), so you can probably just make this &str. Then you can get rid of the `impl Into<String>` and make that `&str` as well. Hopefully I'm right and this doesn't cause you to spend time on something that ends up not compiling

db_store/src/settings.rs
bbalser: Nice catch, Fixed 574c46a

db_store/src/metric_tracker.rs
bbalser: Address in 574c46a

maplant (approved):

andymck (approved):
lgtm

jeffgrunewald (approved):

PR #414, opened 2023-03-15
Update PriceTracker to have configurable duration 

bbalser:
PriceTracker now has a configuration duration that price is allowed to be used.

If now price available in that time range then the PriceTracker task is shutdown and an Err is returned.

### DIFF
--- a/price/src/price_tracker.rs
+++ b/price/src/price_tracker.rs
@@ -1,11 +1,10 @@
 use anyhow::{anyhow, Error, Result};
 use chrono::{DateTime, Duration, TimeZone, Utc};
-use config::{Config, File};
 use file_store::{FileInfo, FileStore, FileType};
 use futures::stream::{StreamExt, TryStreamExt};
 use helium_proto::{BlockchainTokenTypeV1, Message, PriceReportV1};
 use serde::{Deserialize, Serialize};
-use std::{collections::HashMap, path::Path};
+use std::collections::HashMap;
 use tokio::sync::{mpsc, watch};
 
 #[derive(Clone)]
--- a/price/src/price_tracker.rs
+++ b/price/src/price_tracker.rs
@@ -37,14 +37,6 @@ pub struct Settings {
 }
 
 impl Settings {
-    pub fn new<P: AsRef<Path>>(path: P) -> Result<Self> {
-        Config::builder()
-            .add_source(File::with_name(&path.as_ref().to_string_lossy()))
-            .build()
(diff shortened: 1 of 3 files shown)


andymck (approved):
lgtm, but admittedly my experience of channels is limited at this point..but trust the kill task does what it needs

vihu (approved):

vihu:
Looks good, but I'm wondering about the generous usage of `anyhow!`. Do we want to keep this style or have specific errors?

price/src/price_tracker.rs
```diff
@@ -111,21 +181,18 @@ async fn process_file(
...
-                |prices: &mut Prices| match prices.get(&report.token_type()) {
-                    Some(price) if price == &report.price => false,
-                    _ => {
-                        prices.insert(report.token_type(), report.price);
-                        true
-                    }
-                },
-            );
+        .for_each(|(token_type, price)| async move {
+            sender.send_modify(|prices| {
```
jeffgrunewald, line 193: so we're always updating now?

price/src/price_tracker.rs
```diff
@@ -1,31 +1,91 @@
...
+type Prices = HashMap<BlockchainTokenTypeV1, Price>;
+
+#[derive(Debug, Serialize, Deserialize, Clone)]
+pub struct Settings {
+    price_duration_minutes: u64,
+    file_store: file_store::Settings,
+}
+
+impl Settings {
+    pub fn new<P: AsRef<Path>>(path: P) -> Result<Self> {
```
jeffgrunewald, line 39: price tracker is still treated as a process running in the context of another application, right? how are we instantiating the settings for a price tracker in the context of the Verifier (for example) without passing a separate config file path to this function?

price/src/price_tracker.rs
bbalser: Yes, because we need to have the latest_timestamp when we try to get the price to verify that it is not too old.

price/src/price_tracker.rs
bbalser: that function may not be needed, I copied from another crate.

price/src/price_tracker.rs
jeffgrunewald: make sense

price/src/price_tracker.rs
bbalser: Removed 64f7e2a

jeffgrunewald (approved):

Issue #392, opened 2023-02-27
Rewards algorithm for PoC and data transfer work


meowshka:
1. Convert dollar amount for upload and download data transfer by each individual hotspot to MOBILE using Solana’s price oracle. $0.5 for 1GB according to HIP53. 
2. Calculate total % of MOBILE for data transfer work from total emitted MOBILE pool.
3. Subtract % of MOBILE used for data transfer work from total pool of rewards for PoC and data transfer to determine pool of MOBILE available for PoC. (PoC + Data transfer rewards make up 60% of all emitted MOBILE according to HIP53). Total pool of rewards of PoC should not be less than 20% of total emitted pool. Scream alerts and wake everyone up if it falls bellow. 
4. Calculate PoC rewards using total MOBILE available for PoC.
5. Sum up rewards for PoC and data transfer work for each wallet and submit rewards.

PR #416, opened 2023-03-20
Add data transfer rewards to mobile verifier 

maplant:
Someone please check my math!


mobile_verifier/src/owner_shares.rs
```diff
@@ -14,6 +15,80 @@ use rust_decimal::prelude::*;
...
+    }
+
+    pub async fn from_transfer_sessions(
+        dollars_per_mobile: u64,
+        transfer_sessions: impl Stream<Item = ValidDataTransferSession>,
+        epoch: &Range<DateTime<Utc>>,
+    ) -> Result<Self, TransferRewardsError> {
+        tokio::pin!(transfer_sessions);
+
+        let mut bytes = HashMap::<PublicKeyBinary, Decimal>::new();
```
bbalser, line 52: I have flipped to marc's side and I think it would be more readable as 
```rust
        let bytes: HashMap<PublicKeyBinary, Decimal> = transfer_sessions
            .fold(HashMap::new(), |mut acc, next| async move {
                *acc.entry(next.pub_key).or_default() +=
                    Decimal::from(next.upload_bytes + next.download_bytes);
                acc
            })
            .await;
```

mobile_verifier/src/owner_shares.rs
```diff
@@ -100,16 +176,17 @@ impl OwnerShares {
                     .into_iter()
                     .map(move |radio_share| proto::RadioRewardShare {
                         owner_key: owner_key.clone().into(),
-                        hotspot_key: radio_share.hotspot_key.into(),
                         cbsd_id: radio_share.cbsd_id,
                         amount: {
-                            let rewards = rewards_per_share * radio_share.amount;
+                            let rewards = rewards_per_share * radio_share.amount
```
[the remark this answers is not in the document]
maplant: This is for conversion to u64. When converting to u64, the value is truncated, so we multiply it by 10^scale to ensure that the value is in bones

mobile_verifier/src/owner_shares.rs
```diff
@@ -14,6 +15,80 @@ use rust_decimal::prelude::*;
...
+        {
+            *bytes.entry(pub_key).or_default() += Decimal::from(upload_bytes + download_bytes);
+        }
+
+        let mut reward_sum = Decimal::ZERO;
+        let dollars_per_mobile = Decimal::from(dollars_per_mobile);
+        let rewards: HashMap<_, _> = bytes
+            .into_iter()
+            .map(|(pub_key, bytes)| {
+                let mobiles = bytes / Decimal::from(100_000_000) * dollars_per_mobile;
```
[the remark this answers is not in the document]
andymck: still confused by this.  Dont you need to work out the DC costs of the data the hotspot transferred ?  From there you can then determine the value of that DC expressed as mobile tokens ( dividing DC costs by current mobile token price ) 

Also isnt `dollars_per_mobile` here the cost per mobile token, not the cost per half gigabyte

mobile_verifier/src/owner_shares.rs
```diff
@@ -3,17 +3,106 @@ use crate::{
...
+                data_transfer_reward_sum += mobiles;
+                (pub_key, mobiles)
+            })
+            .collect();
+
+        let total_rewards = get_scheduled_tokens(epoch.start, epoch.end - epoch.start)?;
+
+        let (scale, remaining_rewards) =
+            if data_transfer_reward_sum / total_rewards > *MAX_DATA_TRANSFER_REWARDS_PERCENT {
+                let scale = *MIN_POC_REWARDS_PERCENT * total_rewards / data_transfer_reward_sum;
```
andymck, line 70: why use `MIN_POC_REWARDS_PERCENT` to determine scale, shouldnt it be using `MAX_DATA_TRANSFER_REWARDS_PERCENT` ?
Example
total_rewards: 100000
total data rewards: 40000 ( 2x the cap of 0.2 )
total poc rewards: 80000

As the data rewards exceeds the 0.2 cap, scale should be 0.5

if we use the current formula:

MIN_POC_REWARDS_PERCENT * total_rewards / total data rewards = 2

whereas if we use 

MAX_DATA_TRANSFER_REWARDS_PERCENT * total_rewards / total data rewards = 0.5

andymck (commented):
Initial pass, will do another as I absorb things better

mobile_verifier/src/owner_shares.rs
```diff
@@ -100,16 +190,17 @@ impl OwnerShares {
                     .into_iter()
                     .map(move |radio_share| proto::RadioRewardShare {
                         owner_key: owner_key.clone().into(),
-                        hotspot_key: radio_share.hotspot_key.into(),
                         cbsd_id: radio_share.cbsd_id,
                         amount: {
-                            let rewards = rewards_per_share * radio_share.amount;
+                            let rewards = rewards_per_share * radio_share.amount
+                                + transfer_rewards.reward(&radio_share.hotspot_key);
                             let rewards = (rewards * Decimal::from(MOBILE_SCALE))
```
andymck, line 197: arent rewards already in bones and thus already at 10^6, i dont think MOBILE_SCALE is required here

mobile_verifier/src/owner_shares.rs
```diff
@@ -3,17 +3,106 @@ use crate::{
...
+const BYTES_PER_DC: u64 = 66;
+
+fn bytes_to_dc(bytes: u64) -> u64 {
+    let bytes = bytes.max(BYTES_PER_DC);
+    (bytes + BYTES_PER_DC - 1) / BYTES_PER_DC
+}
+
+lazy_static! {
+    static ref MAX_DATA_TRANSFER_REWARDS_PERCENT: Decimal = dec!(0.2);
+    static ref MIN_POC_REWARDS_PERCENT: Decimal = dec!(0.8);
```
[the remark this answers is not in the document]
andymck: k, i see that comes from the example listed here: https://github.com/helium/oracles/issues/392

which looks to be wrong, bbalser is verifying

mobile_verifier/src/owner_shares.rs
```diff
@@ -3,17 +3,106 @@ use crate::{
...
+    rewards: HashMap<PublicKeyBinary, Decimal>,
+    pub remaining_rewards: Decimal,
+}
+
+impl TransferRewards {
+    pub fn empty(epoch: &Range<DateTime<Utc>>) -> Self {
+        Self {
+            scale: Decimal::ONE,
+            rewards: HashMap::new(),
+            remaining_rewards: get_scheduled_tokens(epoch.start, epoch.end - epoch.start).unwrap(),
```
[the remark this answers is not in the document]
andymck: i see the 100_000_000 is defined in this issue:

https://github.com/helium/oracles/issues/392

bbalser is verifying this as it doesnt match what we get from hip53 rewards per day

mobile_verifier/src/owner_shares.rs
maplant: Yup, not sure how I got that math wrong :man_facepalming:

mobile_verifier/src/reward_shares.rs
```diff
@@ -183,9 +288,90 @@ mod test {
...
+
+        assert_eq!(
+            data_transfer_rewards.remaining_rewards,
+            dec!(20_000_000) * dec!(1_000_000)
+        );
+        assert_eq!(
+            // Rewards are automatically scaled
+            data_transfer_rewards.reward(&owner),
+            dec!(80_000_000) * dec!(1_000_000)
+        );
```
andymck, line 372: be good to assert the expected scale value here

mobile_verifier/src/reward_shares.rs
```diff
@@ -601,7 +789,7 @@ mod test {
             total += *val
         }
 
-        assert_eq!(total, 416_666_666_666_669); // total emissions for 1 hour
+        assert_eq!(total, 4_166_666_666_662); // total emissions for 1 hour
     }
```
andymck, line 794: Can you add a test to cover the dc_to_mobile_bones, given that is a key function we want to ensure any potential future behaviour changes is covered

mobile_verifier/src/reward_shares.rs
```diff
@@ -183,9 +288,90 @@ mod test {
...
+
+        let now = Utc::now();
+        let epoch = (now - Duration::hours(24))..now;
+
+        let data_transfer_rewards =
+            TransferRewards::from_transfer_sessions(dec!(1.0), data_transfer_sessions, &epoch)
+                .await
+                .expect("Could not fetch data transfer sessions");
+
+        assert_eq!(
```
andymck, line 364: difficult to understand the expected values here.  Maybe add a comment detailing how those are arrived at

andymck (commented):
looking good.  I need another quick pass on it

mobile_verifier/src/reward_shares.rs
maplant: I tried to make it as close to `0.5` to do just that

mobile_verifier/src/reward_shares.rs
```diff
@@ -84,14 +189,14 @@ impl OwnerShares {
...
     }
 
-    pub fn into_radio_shares(
+    pub fn into_radio_shares<'a>(
         self,
-        epoch: &'_ Range<DateTime<Utc>>,
-    ) -> Result<impl Iterator<Item = proto::RadioRewardShare> + '_, ClockError> {
+        transfer_rewards: &'a TransferRewards,
+        epoch: &'a Range<DateTime<Utc>>,
+    ) -> Result<impl Iterator<Item = proto::RadioRewardShare> + 'a, ClockError> {
```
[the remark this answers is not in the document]
maplant: Good idea

mobile_verifier/src/reward_shares.rs
maplant: I added a test for bytes -> bones, which goes through `dc_to_mobile_bones`, so it should test both

mobile_verifier/src/reward_shares.rs
```diff
@@ -3,17 +3,132 @@ use crate::{
...
     poc_mobile as proto, Channel,
 };
-use lazy_static::lazy_static;
 use rust_decimal::prelude::*;
+use rust_decimal_macros::dec;
 use std::collections::HashMap;
 use std::ops::Range;
 
+/// 100M genesis rewards per day
+const GENESIS_REWARDS_PER_DAY: Decimal = dec!(100_000_000_000_000);
```
andymck, line 19: now that you have TOTAL_EMISSIONS_POOL and data transfer is 40% of that, do we need to define GENESIS_REWARDS_PER_DAY ?

mobile_verifier/src/reward_shares.rs
```diff
@@ -134,8 +244,7 @@ pub fn get_scheduled_tokens(
 ) -> Result<Decimal, ClockError> {
     (*GENESIS_START <= start)
         .then(|| {
-            (Decimal::from(GENESIS_REWARDS_PER_DAY)
-                / Decimal::from(Duration::hours(24).num_seconds()))
+            (GENESIS_REWARDS_PER_DAY / Decimal::from(Duration::hours(24).num_seconds()))
```
andymck, line 247: Should you not be using TOTAL_EMISSIONS_POOL * 0.4 here ? not sure i understand why both TOTAL_EMISSIONS_POOL and GENESIS_REWARDS_PER_DAY is defined

Also suggest renaming `get_scheduled_tokens` and `get_scheduled_dc_tokens`

mobile_verifier/src/reward_shares.rs
```diff
@@ -3,17 +3,132 @@ use crate::{
...
+        //
+        let scale = if data_transfer_reward_sum / TOTAL_EMISSIONS_POOL
+            > MAX_DATA_TRANSFER_REWARDS_PERCENT
+        {
+            MAX_DATA_TRANSFER_REWARDS_PERCENT * TOTAL_EMISSIONS_POOL / data_transfer_reward_sum
+        } else {
+            Decimal::ONE
+        };
+        let remaining_poc_rewards = get_scheduled_tokens(epoch.start, epoch.end - epoch.start)?
+            - scale * data_transfer_reward_sum;
```
andymck, line 108:
```suggestion
        let remaining_poc_rewards = get_scheduled_tokens(epoch.start, epoch.end - epoch.start)?
            - (scale * data_transfer_reward_sum);
```

mobile_verifier/src/reward_shares.rs
```diff
@@ -3,17 +3,132 @@ use crate::{
...
+        tokio::pin!(transfer_sessions);
+
+        let mut data_transfer_reward_sum = Decimal::ZERO;
+        let rewards = transfer_sessions
+            // Accumulate bytes per hotspot
+            .fold(
+                HashMap::<PublicKeyBinary, Decimal>::new(),
+                |mut entries, session| async move {
+                    *entries.entry(session.pub_key).or_default() +=
+                        Decimal::from(bytes_to_dc(session.download_bytes + session.upload_bytes));
```
bbalser, line 72: Shouldn't we be reading the num dcs burned from the valid packet report rather than calculating it again  in the mobile verifier?

mobile_verifier/src/reward_shares.rs
jeffgrunewald: _GENESIS_ rewards should probably go away or change once we're out of the genesis period

mobile_verifier/src/reward_shares.rs
jeffgrunewald: should this guard against `remaining_poc_rewards` being a negative number in case we're missing some edge case in the math?

mobile_verifier/src/reward_shares.rs
```diff
@@ -42,14 +157,11 @@ impl RadioShares {
 }
 
 #[derive(Default)]
-pub struct OwnerShares {
+pub struct RewardShares {
```
jeffgrunewald, line 160: if we're renaming this should we be more specific and call it `PocShares` as distinct from `TransferShares` but both still falling under the umbrella of `RewardShares` since they're both rewards?

mobile_verifier/src/reward_shares.rs
```diff
@@ -84,14 +196,14 @@ impl OwnerShares {
...
-        epoch: &'_ Range<DateTime<Utc>>,
-    ) -> Result<impl Iterator<Item = proto::RadioRewardShare> + '_, ClockError> {
+        transfer_rewards: &'a TransferRewards,
+        epoch: &'a Range<DateTime<Utc>>,
+    ) -> Result<impl Iterator<Item = proto::RadioRewardShare> + 'a, ClockError> {
         let total_shares = self.total_shares();
-        let total_rewards = get_scheduled_tokens(epoch.start, epoch.end - epoch.start)?;
-        let rewards_per_share = (total_rewards / total_shares)
-            .round_dp_with_strategy(REWARDS_PER_SHARE_PREC, RoundingStrategy::ToPositiveInfinity);
+        let total_poc_rewards = transfer_rewards.remaining_poc_rewards;
```
jeffgrunewald, line 205: this doesn't seem right. shouldn't the total PoC rewards be 20% of that TOTAL_EMISSIONS_POOL number + any remainder from transfer rewards?

mobile_verifier/src/reward_shares.rs
jeffgrunewald: i concur on the use of both; i believe once we are issuing rewards from the emissions pool we should no longer be issuing rewards from the genesis period

mobile_verifier/src/verifier.rs
```diff
@@ -109,15 +114,34 @@ impl VerifierDaemon {
...
+                &self.verifier.file_store,
+                &scheduler.reward_period,
+            )
+            .await,
+            &scheduler.reward_period,
+        )
+        .await?;
+
+        // It's important to gauge the scale metric. If this value is < 1.0, we are in
+        // big trouble.
```
jeffgrunewald, line 136: if we're in _real_ trouble should we just bail the whole thing here? how much trouble are we talking? isn't that math dividing based on 40% of the total emissions available supposed to protect us from that trouble?

mobile_verifier/src/reward_shares.rs
```diff
@@ -3,17 +3,132 @@ use crate::{
...
 use rust_decimal::prelude::*;
+use rust_decimal_macros::dec;
 use std::collections::HashMap;
 use std::ops::Range;
 
+/// 100M genesis rewards per day
+const GENESIS_REWARDS_PER_DAY: Decimal = dec!(100_000_000_000_000);
+
+/// Total emissions pool
+const TOTAL_EMISSIONS_POOL: Decimal = dec!(166_666_666_666_666);
```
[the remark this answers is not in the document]
andymck: hip75 ( https://github.com/helium/HIP/blob/main/0075-mobile-poc-initiate-programmatic-minting-and-updated-emissions-curve.md ) superceeds hip53.  The total yearly emissions is 166B ( 50B premine, 116B over the year )

mobile_verifier/src/reward_shares.rs
andymck: should we rename `num_dcs` in the packet report then to `num_burnt_dcs`?

mobile_verifier/src/reward_shares.rs
andymck: I think we should start with the 30B total and work things out from there to the daily or epoch count

mobile_verifier/src/cli/generate.rs
```diff
@@ -48,7 +49,8 @@ impl Cmd {
 
         let mut total_rewards = 0_u64;
         let mut owner_rewards = HashMap::<_, u64>::new();
-        for reward in owner_shares.into_radio_shares(&epoch)? {
+        let transfer_rewards = TransferRewards::empty(&epoch);
```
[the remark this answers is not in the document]
maplant: Maybe? We don't really use these commands any more

mobile_verifier/src/reward_shares.rs
maplant: We confirmed that it is the same calculation so I think `num_dcs` is fine

mobile_verifier/src/reward_shares.rs
jeffgrunewald: i'm not sure how we get these numbers; this test establishes the epoch to be a 24 hour period and the test above establishes the _total_ rewards for a 24 period are `dec!(100_000_000)` so how are we getting these significantly larger numbers? and yes, a comment to indicate 33_333_333_333_333 is expected to be 20% of total rewards for a 24 hour period and 66_666_666_666_666 is 40% of the total rewards for a 24 hour period

mobile_verifier/src/reward_shares.rs
```diff
@@ -3,17 +3,127 @@ use crate::{
...
+        // [ scale * data_transfer_reward_sum ] / total_emissions_pool = 0.4
+        //
+        //   therefore:
+        //
+        // scale = [ 0.4 * total_emissions_pool ] / data_transfer_reward_sum
+        //
+        let scale = if data_transfer_reward_sum / total_emissions_pool
+            > MAX_DATA_TRANSFER_REWARDS_PERCENT
+        {
+            MAX_DATA_TRANSFER_REWARDS_PERCENT * total_emissions_pool / data_transfer_reward_sum
```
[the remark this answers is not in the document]
maplant: It technically should not matter but order of operations is same precedence for multiplication and division, but left-to-right associative https://doc.rust-lang.org/reference/expressions.html#expression-precedence

mobile_verifier/src/reward_shares.rs
```diff
@@ -3,17 +3,127 @@ use crate::{
...
+const DEFAULT_PREC: u32 = 15;
+
+pub struct TransferRewards {
+    scale: Decimal,
+    rewards: HashMap<PublicKeyBinary, Decimal>,
+    pub remaining_poc_rewards: Decimal,
+}
+
+impl TransferRewards {
+    pub fn empty(epoch: &Range<DateTime<Utc>>) -> Self {
```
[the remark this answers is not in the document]
maplant: It takes a range, and I want this to be explicit

mobile_verifier/src/reward_shares.rs
```diff
@@ -100,20 +95,18 @@ impl TransferRewards {
...
-        let remaining_poc_rewards = rewards_for_poc_and_dc - scale * data_transfer_reward_sum;
 
-        Ok(Self {
-            scale,
+        Self {
+            reward_scale,
             rewards,
-            remaining_poc_rewards,
-        })
+            reward_sum: reward_sum * reward_scale,
```
jeffgrunewald, line 108: in the event data transfer rewards _did_ exceed the allowed 40% at this point haven't you already allocated the overage to each gateway pubkey in the hashmap that made up the `reward_sum` total so it's too late to multiply by the scale percentage here? like, if we tallied up all of the individual data transfer rewards at the end and found they exceeded the allowable amount shouldn't we iterate over each entry in the map, multiplying _that_ by the reward_scale percentage and then re-sum the total?

jeffgrunewald (approved):

mobile_verifier/src/reward_shares.rs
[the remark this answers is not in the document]
jeffgrunewald: confusing my "long division" mnemonics for my "order of operations" mnemonics; you're right about order here

bbalser (approved):

PR #417, opened 2023-03-21
Refactor price service and enable stale price lifetime

bbalser:
Refactored price service to have 2 modes
1.  If token key is configured, then it will get new price from solana and output on configured cadence. It will also absorb downtimes and use the stale price up to the configured duration
2. If no key is configured, but a default value is, it will output the default price on the configured cadence
3. If no key or default price is configured, the price generator will stop with a value ` Ok(())` for that token.

### DIFF
--- a/price/src/price_generator.rs
+++ b/price/src/price_generator.rs
@@ -172,7 +172,7 @@ impl PriceGenerator {
             }
             Err(err) => {
                 tracing::error!(
-                    "error is retrieving new price for {:?}: {err:?}",
+                    "error in retrieving new price for {:?}: {err:?}",
                     self.token_type
                 );
 
--- a/price/src/price_generator.rs
+++ b/price/src/price_generator.rs
@@ -1,6 +1,6 @@
-use crate::{metrics::Metrics, PriceError, Settings};
-use anyhow::Result;
-use chrono::Utc;
+use crate::{metrics::Metrics, Settings};
+use anyhow::{anyhow, Error, Result};
+use chrono::{DateTime, Duration, TimeZone, Utc};
 use file_store::file_sink;
 use helium_proto::{BlockchainTokenTypeV1, PriceReportV1};
 use pyth_sdk_solana::load_price_feed_from_account;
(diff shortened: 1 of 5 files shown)


maplant:
Instead of using anyhow, since this an API we should have an thiserror enum error type for this crate

price/src/price_generator.rs
```diff
@@ -11,13 +11,13 @@ use tokio::time;
 
 #[derive(Debug, Clone, Serialize)]
 pub struct Price {
-    pub timestamp: i64,
+    pub timestamp: DateTime<Utc>,
```
maplant, line 14: Is it necessary for all of these fields to be public?

price/src/price_generator.rs
```diff
@@ -11,13 +11,13 @@ use tokio::time;
...
     pub price: u64,
     pub token_type: BlockchainTokenTypeV1,
 }
 
 impl Price {
-    pub fn new(timestamp: i64, price: u64, token_type: BlockchainTokenTypeV1) -> Self {
+    pub fn new(timestamp: DateTime<Utc>, price: u64, token_type: BlockchainTokenTypeV1) -> Self {
```
maplant, line 20: I'd recommend making this new function `pub(crate)`. I don't think there's ever a need to construct this outside of this crate

jeffgrunewald (approved):

price/src/price_generator.rs
bbalser: I just removed pub and everything compiled 29f472a

price/src/price_generator.rs
```diff
@@ -59,148 +67,153 @@ impl TryFrom<PriceReportV1> for Price {
...
-                    Metrics::update(
-                        "price_stale_counter".to_string(),
-                        token_type,
-                        old_price.price as f64,
-                    );
-                    tracing::warn!("rpc failure: {err:?}, reusing old price {:?}", old_price);
-                    Some(old_price)
+            Err(err) => {
+                tracing::error!(
+                    "error is retrieving new price for {:?}: {err:?}",
```
vihu, line 175:
```suggestion
                    "error in retrieving new price for {:?}: {err:?}",
```

PR #419, opened 2023-03-23
mobile config

jeffgrunewald:
Implement the initial functionality of the mobile-config service; it currently mirrors certain functionality internal to the oracle processing pipeline:
- the hotspot metadata service for providing the mobile verifier with on-chain metadata for mobile hotspots
- router service for verifying a given mobile packet router's authorization to burn data credits on behalf of the mobile network
- admin service for administering the mobile config service (starting with managing authorizing and signing keys)
relies on successfully merging:
- [x] https://github.com/helium/proto/pull/308
- [x] https://github.com/helium/gateway-rs/pull/399


mobile_config/src/hotspot_service.rs
```diff
@@ -0,0 +1,178 @@
...
+                Ok(signature) => Some(HotspotMetadataStreamResV1 {
+                    hotspots: res.hotspots,
+                    timestamp: res.timestamp,
+                    signature,
+                }),
+                Err(_) => None,
+            };
+            Ok(result)
+        })
+        .or_else(|err| async move {
```
bbalser, line 168: You should be able to use a `map_err` here which doesn't need to return a Future so it should increase readability a little

mobile_config/migrations/2_registered_keys.sql
```diff
@@ -0,0 +1,15 @@
...
+    'administrator',
+    'packet_router'
+    'oracle',
+);
+
+create table registered_keys (
+    pubkey text not null unique,
+    key_type key_type not null,
+
+    inserted_at timestamptz not null default now(),
```
andymck, line 11: nit: created_at rather than inserted_at a lil more descriptive

andymck (approved):
lgtm.  My only concern is the lack of tests which was same for the iot config service

mobile_config/src/hotspot_service.rs
jeffgrunewald: [eb412d3](https://github.com/helium/oracles/pull/419/commits/eb412d3d6af761ab3ec6bae92ef61e4f7792f01e) i swapped the first `or_else` out for a `map_err` but then in the followup i reworked the final error map into an `or_else` to catch the errors and resend them on to the client to decide how to handle them (or not) since the stream is technically a stream of `Result<HotspotInfo, Status>`

mobile_config/src/key_cache.rs
```diff
@@ -0,0 +1,222 @@
...
+
+#[derive(Clone, Copy, Debug, Serialize, sqlx::Type)]
+#[sqlx(type_name = "key_type", rename_all = "snake_case")]
+pub enum KeyType {
+    Administrator,
+    Oracle,
+    PacketRouter,
+}
+
+async fn fetch_stored_keys(
```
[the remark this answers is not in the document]
jeffgrunewald: done [f88093e](https://github.com/helium/oracles/pull/419/commits/f88093e7c9c967583113b43a7dc7dd9997eb6916)

mobile_config/migrations/2_registered_keys.sql
```diff
@@ -0,0 +1,15 @@
+create type key_type as enum (
+    'administrator',
+    'packet_router'
+    'oracle',
```
bbalser, line 4: postgres errors on this statement, can't have ending `,` .

jeffgrunewald: spoiled by rust; fixed [3e2ca43](https://github.com/helium/oracles/pull/419/commits/3e2ca43b54d5f0aaeaef51ca36ecb1aec3299408)

andymck (approved):

mobile_config/src/admin_service.rs
```diff
@@ -0,0 +1,159 @@
...
+};
+use sqlx::{Pool, Postgres};
+use tokio::sync::watch;
+use tonic::{Request, Response, Status};
+
+pub struct AdminService {
+    key_cache: KeyCache,
+    key_cache_updater: watch::Sender<CacheKeys>,
+    pool: Pool<Postgres>,
+    required_network: Network,
```
vihu, line 23: nit:  "required" seems redundant here? Just use network and either crash or don't do anything if the network is not present in settings?

Also, naming wise, should this be "network" or "cluster"?

mobile_config/src/client/mod.rs
```diff
@@ -0,0 +1,100 @@
...
+}
+
+#[async_trait::async_trait]
+impl gateway_info::GatewayInfoResolver for Client {
+    type Error = ClientError;
+
+    async fn resolve_gateway_info(
+        &mut self,
+        address: &PublicKeyBinary,
+    ) -> Result<Option<gateway_info::GatewayInfo>, Self::Error> {
```
vihu, line 50: Does the return type here need to be a `Result<Option<GatewayInfo>>`? Can it just be `Result<GatewayInfo>` instead and handled accordingly downstream?

jeffgrunewald: I believe this is a valid return type for general use and it’s up to the caller to determine how to interpret that. It mirrors the retrieval of the record from the helius database in that regard in that we allow a successful request return a None when the gateway is not found on chain but we can error for other failure scenarios that could be recovered from like a malformed pubkey request argument or an unauthorized request.

mobile_config/src/hotspot_service.rs
```diff
@@ -0,0 +1,179 @@
...
+
+        self.verify_request_signature(&request).await?;
+
+        tracing::debug!("fetching all hotspots' metadata");
+
+        let pool = self.metadata_pool.clone();
+        let signing_key = self.signing_key.clone();
+        let batch_size = request.batch_size;
+
+        let (tx, rx) = tokio::sync::mpsc::channel(20);
```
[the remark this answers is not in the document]
jeffgrunewald: increased; the client _is_ generally understood to be an oracle but this is a stream of a potentially large number of gateways so i bumped up the threshold.

iot_config/src/session_key_service.rs
```diff
@@ -199,94 +199,95 @@ impl iot_config::SessionKeyFilter for SessionKeyFilterService {
...
+                            self.update_validator(filter.oui).await.map_err(|err| {
+                                Status::internal(format!("unable to verify updates {err:?}"))
+                            })
+                        }
+                        None => Err(Status::invalid_argument("no session key filter provided")),
+                    },
+                    Err(_) => Err(Status::invalid_argument("no session key filter provided")),
+                }
+            })
+            .ok_or(Status::invalid_argument("no session key filter provided"))?
```
maplant, line 221:
```suggestion
            .ok_or_else(|| Status::invalid_argument("no session key filter provided"))?
```

mobile_config/src/key_cache.rs
```diff
@@ -0,0 +1,195 @@
...
+    }
+}
+
+pub(crate) mod db {
+    use super::{KeyType, PublicKey, PublicKeyBinary};
+    use sqlx::Row;
+
+    pub(crate) async fn fetch_stored_keys(
+        db: impl sqlx::PgExecutor<'_>,
+    ) -> anyhow::Result<Vec<(PublicKey, KeyType)>> {
```
maplant, line 148: You can just have this return a `CachedKey` instead of into_itering the result and re-collecting

mobile_config/src/key_cache.rs
```diff
@@ -0,0 +1,195 @@
...
+    Administrator,
+    Oracle,
+    PacketRouter,
+}
+
+impl KeyType {
+    pub fn from_i32(v: i32) -> anyhow::Result<Self> {
+        ProtoKeyType::from_i32(v)
+            .map(|kt| kt.into())
+            .ok_or(anyhow!("unsupported key type {}", v))
```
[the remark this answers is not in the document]
jeffgrunewald: already fixed

mobile_config/src/key_cache.rs
jeffgrunewald: good call; i probably _thought_ i wanted a generic result here at one point but i'm never using this function anywhere else so that's a good recommendation

maplant (approved):

PR #424, opened 2023-03-30
handle gateway info metadata struct

andymck:
*** NOTE ***  this has a dep on a yet to be created proto branch which will return a gateway metadata struct containing asserted related fields.  Until that proto is available it maps the existing gateway info proto to a newly updated internal GatewayInfo struct.

With this change, when resolving gateway info for a beaconer or a witness, we only return the gateway info if it contains the expected asserted related fields.  These will be contained within an Option field of gateway info, if this is None it is assumed the gateway is not yet asserted and an appropriate error is returned.

The verifier will only proceed to verifying a beacon or witness if the asserted data is available, otherwise a verification fail is returned immediately.


andymck:
superseded by #428

PR #428, opened 2023-04-04
Iot Config service use Helius for on-chain metadata

jeffgrunewald:
Switches the iot-config service (and the rest of oracles) to use the helius database for retrieving on-chain gateway metadata instead of the blockchain-node follower client service.

This also adds additional optimizations for signature verification of sent and received messages with the iot-config service, primarily enforcing messages include the binary of the signing pubkey for O(1) verification of the signature verification instead of O(n) when checking the signature against _all_ potentially valid signing keys; instead of attempting to verifying the signature of a message against all potential keys we validate against the key as self-reported in the message and then simply check for membership of that key in the list of potential authorized signers. This also introduces additional message signing of _outgoing_ messages sent by the config service itself

Depends on:
- [x] https://github.com/helium/proto/pull/308


file_store/src/traits/msg_verify.rs
```diff
@@ -40,29 +32,35 @@ impl_msg_verify!(SpeedtestReqV1, signature);
...
-impl_msg_verify!(GatewayLocationReqV1, signature);
-impl_msg_verify!(GatewayRegionParamsReqV1, signature);
-impl_msg_verify!(AdminAddKeyReqV1, signature);
-impl_msg_verify!(AdminLoadRegionReqV1, signature);
-impl_msg_verify!(AdminRemoveKeyReqV1, signature);
-impl_msg_verify!(SessionKeyFilterGetReqV1, signature);
-impl_msg_verify!(SessionKeyFilterListReqV1, signature);
-impl_msg_verify!(SessionKeyFilterStreamReqV1, signature);
-impl_msg_verify!(SessionKeyFilterUpdateReqV1, signature);
+impl_msg_verify!(iot_config::OrgCreateHeliumReqV1, signature);
```
jeffgrunewald, line 35: i've started pathing these message types here since they will have very similar names between other services (like the mobile config service)

iot_config/src/client/mod.rs
```diff
@@ -0,0 +1,141 @@
...
+            .info_stream(request)
+            .await?
+            .into_inner()
+            .filter_map(|resp| async move { resp.ok() })
+            .map(move |resp| (resp, pubkey.clone()))
+            .filter_map(|(resp, pubkey)| async move {
+                match resp.verify(&pubkey) {
+                    Ok(()) => Some(resp),
+                    Err(_) => None,
+                }
```
bbalser, line 121:
```rust
  resp.verify(&pubkey).map(|_| resp).ok()
```

jeffgrunewald: done

iot_config/src/client/settings.rs
```diff
@@ -0,0 +1,47 @@
+use serde::Deserialize;
+use std::{str::FromStr, sync::Arc};
+
+#[derive(Clone, Debug, Deserialize)]
+pub struct Settings {
+    /// grpc url to the iot config oracle server
+    #[serde(with = "http_serde::uri")]
+    pub url: http::Uri,
+    /// File from which to load keypair for signing config server requests
```
andymck, line 9:
```suggestion
    /// File from which to load keypair for signing config client requests
```

iot_config/src/admin_service.rs
```diff
@@ -82,29 +101,51 @@ impl iot_config::Admin for AdminService {
...
+        resp.signature = self.sign_response(&resp.encode_to_vec())?;
+
+        Ok(Response::new(resp))
     }
 
     async fn remove_key(&self, request: Request<AdminRemoveKeyReqV1>) -> GrpcResult<AdminKeyResV1> {
         let request = request.into_inner();
 
-        self.verify_request_signature(&request).await?;
+        let signer = self.verify_public_key(&request.signer)?;
```
andymck, line 139: why verify the public key, if we are subsequently verifying the signature ?  the sig verification will fail if the pub key is invalid anyway

andymck (commented):
Initial first pass, will need to take a second run

iot_config/src/gateway_info.rs
```diff
@@ -0,0 +1,176 @@
...
+    pub location: u64,
+    pub elevation: i32,
+    pub gain: i32,
+    pub region: Region,
+}
+
+#[derive(Clone, Debug)]
+pub struct GatewayInfo {
+    pub address: PublicKeyBinary,
+    pub metadata: Option<GatewayMetadata>,
```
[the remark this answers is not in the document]
jeffgrunewald: Yeah, I thought about both options but there are other consumers of the GatewayInfo type that still find value in a successful response for a non-asserted gateway

iot_config/src/admin_service.rs
jeffgrunewald: `verify_public_key` is how we turn the proto’s bytes field into a helium_crypto `PublicKey`, so we have to do some version of that verification in order to perform the signature verification

jeffgrunewald:
This is _ready to review_ it's in "draft" status to prevent merging before updating the dependencies once they get updated

iot_config/src/admin_service.rs
andymck: k, i checked the actual verify_public_key function, its simply doing a try_from. I had originally thought there was more going on with this function

iot_config/src/client/mod.rs
```diff
@@ -0,0 +1,136 @@
...
+        &mut self,
+        address: &PublicKeyBinary,
+    ) -> Result<Option<gateway_info::GatewayInfo>, Self::Error> {
+        let mut request = iot_config::GatewayInfoReqV1 {
+            address: address.clone().into(),
+            signer: self.signing_key.public_key().into(),
+            signature: vec![],
+        };
+        request.signature = self.signing_key.sign(&request.encode_to_vec())?;
+        tracing::debug!(pubkey = address.to_string(), "fetching gateway info");
```
andymck, line 86: nit: not sure what value this log statement provide

iot_config/src/gateway_info.rs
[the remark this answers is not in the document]
andymck: i closed #424, after 428 gets merged i will maybe open a new PR with a few tweaks here and there

iot_config/src/gateway_info.rs
```diff
@@ -0,0 +1,176 @@
...
+        })
+    }
+}
+
+pub(crate) mod db {
+    use futures::stream::{Stream, StreamExt};
+    use helium_crypto::PublicKeyBinary;
+    use sqlx::{PgExecutor, Row};
+
+    pub struct IotMetadata {
```
andymck, line 129: slightly confusing having both GatewayMetadata and IotMetadata structs

andymck (approved):
this lgtm, my eyes hurt from all the scrolling but other than that lg...
Once merged i may tweak a few things on the iot verifier side

iot_config/src/admin.rs
```diff
@@ -1,182 +1,105 @@
...
-    router_keys: Arc<RwLock<HashSet<PublicKey>>>,
+    cache_receiver: watch::Receiver<CacheKeys>,
 }
 
 impl AuthCache {
     pub async fn new(
         settings: &Settings,
         db: impl sqlx::PgExecutor<'_> + Copy,
-    ) -> Result<Self, AdminAuthError> {
+    ) -> anyhow::Result<(watch::Sender<CacheKeys>, Self)> {
```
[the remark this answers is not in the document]
jeffgrunewald: this is a stylistic choice that i'm up for debate over; i don't feel _overly_ strongly about it, but given the `std::result::Result` is part of the typical prelude unless you override it with `use anyhow::Result` then it's easier at a glance, particularly in a large file that what you're returning is _not_ the std Result type

iot_config/src/client/mod.rs
jeffgrunewald: this is probably a better place for a metric on grpc requests then a log statement, good point

iot_config/src/gateway_info.rs
jeffgrunewald: i had hoped scoping the `IotMetadata` struct inside the `db` submodule would help with that, but i couldn't think of a better way to do unless i skipped the `FromRow` impl altogether; since the `GatewayInfo` requires the hextree map to look up a Region from the asserted location and i can't change the `from_row` function signature to pass that in i needed the intermediary data structure

iot_config/src/gateway_service.rs
```diff
@@ -40,33 +75,50 @@ impl iot_config::Gateway for GatewayService {
...
+                    )))
+                },
+                |info| {
+                    if let Some(location) = info.location {
+                        Ok(location)
+                    } else {
+                        Err(Status::not_found(format!(
+                            "gateway unasserted: pubkey = {address:}"
+                        )))
+                    }
```
bbalser, line 101: NIT:
```rust
                    info.location.ok_or_else(|| {
                        Status::not_found(format!("gateway unasserted: pubkey = {address:}"))
                    })
```

iot_config/src/gateway_service.rs
```diff
@@ -40,33 +75,50 @@ impl iot_config::Gateway for GatewayService {
...
+        let signer = self.verify_public_key(&request.signer)?;
+        self.verify_request_signature(&signer, &request)?;
 
-        let location = self
-            .follower_service
-            .clone()
-            .resolve_gateway_info(gateway_address)
+        let address: &PublicKeyBinary = &request.gateway.into();
+
+        let location = gateway_info::db::get_info(&self.metadata_pool, address)
```
bbalser, line 85: NIT: I think you can get location in one step and make it more readable by not using the question mark until the end of the chain.
```rust
        let location = gateway_info::db::get_info(&self.metadata_pool, address)
            .await
            .map_err(|_| Status::internal("error fetching gateway info"))
            .and_then(|opt| {
                opt.ok_or_else(|| {
                    Status::not_found(format!("gateway not found: pubkey = {address:}"))
                })
            })
            .and_then(|iot_metadata| {
                iot_metadata.location.ok_or_else(|| {
                    Status::not_found(format!("gateway unasserted: pubkey = {address:}"))
                })
            })
            .and_then(|location| {
                Cell::from_raw(location).map_err(|_| {
                    Status::internal(format!(
                        "invalid h3 index location {location} for {address}"
                    ))
                })
            })
            .map(|cell| cell.to_string())?;
```

iot_config/src/route.rs
```diff
@@ -263,16 +301,25 @@ pub async fn update_euis(
...
-                    update_tx.clone(),
-                )
+                let mut update_res = proto::RouteStreamResV1 {
+                    action: i32::from(action),
+                    data: Some(proto::route_stream_res_v1::Data::EuiPair(update.into())),
+                    timestamp,
+                    signer: signer.clone(),
+                    signature: vec![],
+                };
+                futures::future::ready(signing_key.sign(&update_res.encode_to_vec()))
```
bbalser, line 316: Nice! I need to remember this.

iot_config/src/route_service.rs
```diff
@@ -80,49 +92,54 @@ impl RouteService {
...
+            .map_err(|_| Status::internal("response signing error"))
+    }
+
     async fn update_validator(
         &self,
         route_id: &str,
         check_constraints: bool,
     ) -> Result<DevAddrEuiValidator, DbOrgError> {
-        let admin_keys = self.auth_cache.get_keys(KeyType::Administrator).await;
+        // let admin_keys = self.auth_cache.get_keys().into_iter().filter(|(_, keytype)| keytype == &KeyType::Administrator).map(|(pubkey, _)| pubkey).collect();
```
bbalser, line 141: I am assuming this should be removed

iot_config/src/gateway_info.rs
bbalser: I understand andymck's point, but I can't think of a better way either and I think scoping `IotMetadata` to the `db` submodule does help.

bbalser (approved):
Couple Nits, but otherwise nice job!!!

iot_config/src/route_service.rs
jeffgrunewald: good catch

iot_config/src/admin.rs
```diff
@@ -1,182 +1,105 @@
...
-#[error("unsupported key type {0}")]
-pub struct UnsupportedKeyTypeError(i32);
-
 impl KeyType {
-    pub fn from_i32(v: i32) -> Result<Self, UnsupportedKeyTypeError> {
+    pub fn from_i32(v: i32) -> anyhow::Result<Self> {
         ProtoKeyType::from_i32(v)
             .map(|kt| kt.into())
-            .ok_or(UnsupportedKeyTypeError(v))
+            .ok_or(anyhow!("unsupported key type {}", v))
```
maplant, line 102:
```suggestion
            .ok_or_else(|| anyhow!("unsupported key type {}", v))
```

iot_config/src/gateway_info.rs
```diff
@@ -0,0 +1,176 @@
...
+    }
+}
+
+fn h3index_to_region(
+    location: u64,
+    region_map: &region_map::RegionMapReader,
+) -> anyhow::Result<Region> {
+    hextree::Cell::from_raw(location)
+        .map(|cell| region_map.get_region(cell))?
+        .ok_or(anyhow!("invalid region"))
```
maplant, line 65:
```suggestion
        .ok_or_else(|| anyhow!("invalid region"))
```

iot_config/src/gateway_service.rs
```diff
@@ -75,107 +121,181 @@ impl iot_config::Gateway for GatewayService {
...
+        let signer = self.verify_public_key(&request.signer)?;
+        self.verify_request_signature(&signer, &request)?;
+
+        let address = &request.address.into();
+        let metadata_info = gateway_info::db::get_info(&self.metadata_pool, address)
+            .await
+            .map_err(|_| Status::internal("error fetching gateway info"))?
+            .ok_or(Status::not_found(format!(
+                "gateway not found: pubkey = {address:}"
+            )))?;
```
maplant, line 211:
```suggestion
            .ok_or_else(|| Status::not_found(format!(
                "gateway not found: pubkey = {address:}"
            )))?;
```

PR #429, opened 2023-04-04
Refactor to support IotRewardShare and multiple sub reward types.  Add support for operation fund rewards

andymck:
Depends on proto branch: https://github.com/helium/proto

Refactor to replace GatewayRewardShare with IotRewardShare.  IotRewardShare now supports multiple reward types each with their own data requirements.

Adds output of operational fund rewards

### DIFF
--- a/iot_verifier/src/reward_share.rs
+++ b/iot_verifier/src/reward_share.rs
@@ -4,7 +4,7 @@ use file_store::{iot_packet::IotValidPacket, iot_valid_poc::IotPoc, traits::Time
 use futures::stream::TryStreamExt;
 use helium_crypto::PublicKeyBinary;
 use helium_proto::services::poc_lora as proto;
-
+use helium_proto::services::poc_lora::iot_reward_share::Reward as ProtoReward;
 use lazy_static::lazy_static;
 use rust_decimal::prelude::*;
 use rust_decimal_macros::dec;
@@ -23,6 +23,8 @@ lazy_static! {
     static ref WITNESS_REWARDS_PER_DAY_PERCENT: Decimal = dec!(0.24);
     // Data transfer is allocated 50% of daily rewards
     static ref DATA_TRANSFER_REWARDS_PER_DAY_PERCENT: Decimal = dec!(0.50);
+    // Operations fund is allocated 7% of daily rewards
+    static ref OPERATIONS_REWARDS_PER_DAY_PERCENT: Decimal = dec!(0.07);
     // dc remainer distributed at ration of 4:1 in favour of witnesses
     // ie WITNESS_REWARDS_PER_DAY_PERCENT:BEACON_REWARDS_PER_DAY_PERCENT
     static ref WITNESS_DC_REMAINER_PERCENT: Decimal = dec!(0.80);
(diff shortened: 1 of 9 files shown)


iot_verifier/src/reward_share.rs
```diff
@@ -62,6 +64,16 @@ fn get_scheduled_dc_tokens(duration: Duration) -> Decimal {
     )
 }
 
+fn get_scheduled_ops_fund_tokens(duration: Duration) -> u64 {
+    get_tokens_by_duration(
+        *REWARDS_PER_DAY * *OPERATIONS_REWARDS_PER_DAY_PERCENT,
+        duration,
+    )
+    .round_dp_with_strategy(0, RoundingStrategy::ToZero)
```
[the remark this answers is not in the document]
bbalser: We use this rounding strategy to round any fractional bones right before we write the amounts to the output, because operation funds have no other calculations, we use this strategy here.

iot_verifier/src/rewarder.rs
```diff
@@ -1,4 +1,4 @@
-use crate::{reward_share::GatewayShares, scheduler::Scheduler};
+use crate::{reward_share::operational_rewards, reward_share::GatewayShares, scheduler::Scheduler};
```
jeffgrunewald, line 1:
```suggestion
use crate::{reward_share::{operational_rewards, GatewayShares}, scheduler::Scheduler};
```

reward_index/migrations/5_add_type_to_index.sql
```diff
@@ -0,0 +1,3 @@
+CREATE TYPE reward_type as enum('mobile', 'iot_gateway', 'iot_operational');
```
jeffgrunewald, line 1: no `'mobile_operations'`? i don't see an equivalent operational funding percentage in the hip53 breakdown; is that accurate?

iot_verifier/src/rewarder.rs
bbalser: Addressed 0d43f6f

reward_index/migrations/5_add_type_to_index.sql
bbalser: yes there is no `mobile_operations`

iot_verifier/src/reward_share.rs
```diff
@@ -23,6 +23,8 @@ lazy_static! {
     static ref WITNESS_REWARDS_PER_DAY_PERCENT: Decimal = dec!(0.24);
     // Data transfer is allocated 50% of daily rewards
     static ref DATA_TRANSFER_REWARDS_PER_DAY_PERCENT: Decimal = dec!(0.50);
+    // Operations fund is allocated 7% of daily rewards
+    static ref OPERATIONS_REWARDS_PER_DAY_PERCENT: Decimal = dec!(0.07);
     // dc remainer distributed at ration of 4:1 in favour of witnesses
     // ie WITNESS_REWARDS_PER_DAY_PERCENT:BEACON_REWARDS_PER_DAY_PERCENT
     static ref WITNESS_DC_REMAINER_PERCENT: Decimal = dec!(0.80);
```
maplant, line 30: All of these can be constants instead of lazy statics

maplant (approved):

PR #433, opened 2023-04-10
Add mobile_reward_share output to mobile verifier

maplant:


mobile_verifier/src/verifier.rs
```diff
@@ -139,17 +137,22 @@ impl VerifierDaemon {
...
+                // Await the returned one shot to ensure that we wrote the file
+                .await??;
+            self.mobile_rewards
+                .write(mobile_reward_share, [])
                 .await?
                 // Await the returned one shot to ensure that we wrote the file
                 .await??;
         }
 
-        let written_files = self.radio_rewards.commit().await?.await??;
```
jeffgrunewald, line 152: do these not _both_ need to be committed for as long as we're writing out the legacy `radio_reward_shares`?

maplant: I changed `radio_reward_shares` to auto commit

PR #436, opened 2023-04-11
IOT Verifier fixes

bbalser:
Misc fixes from local testing of the iot verifier using mainnet POC data.  Tests were run with a backlog of 24hrs worth of data

### DIFF
--- a/iot_verifier/migrations/10_gateway_dc_shares.sql
+++ b/iot_verifier/migrations/10_gateway_dc_shares.sql
@@ -6,6 +6,4 @@ create table gateway_dc_shares (
     id bytea primary key not null
 );
 
-create index idx_hotspot_key on gateway_dc_shares (hotspot_key);
-
-create index idx_reward_type on gateway_dc_shares (reward_type);
+create index idx_gds_hotspot_key on gateway_dc_shares (hotspot_key);
--- a/iot_verifier/src/loader.rs
+++ b/iot_verifier/src/loader.rs
@@ -145,7 +145,7 @@ impl Loader {
             .unwrap_or(window_default_lookback)
             .max(window_max_lookback);
         let before_max = after + self.window_width;
-        let before = (now - (self.window_width * 3)).min(before_max);
+        let before = (now - (self.window_width * 3)).max(before_max);
         let window_width = (before - after).num_minutes() as u64;
         tracing::info!("sliding window, after: {after}, before: {before}, width: {window_width}");
         self.process_window(gateway_cache, after, before).await?;
--- a/iot_verifier/src/loader.rs
+++ b/iot_verifier/src/loader.rs
@@ -154,7 +154,7 @@ impl Loader {
             "sliding window, after: {after}, before: {before}, cur width: {:?}, required width: {:?}",
             cur_window_width.num_minutes(), self.window_width.num_minutes()
         );
-        // if the current window width is less that our expected width
+        // if the current window width is less than our expected width
         // then do nothing
         // this likely means our loader tick interval is set to a value
         // less than our width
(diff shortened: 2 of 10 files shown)


iot_verifier/migrations/10_gateway_dc_shares.sql
```diff
@@ -6,6 +6,4 @@ create table gateway_dc_shares (
     id bytea primary key not null
 );
 
-create index idx_hotspot_key on gateway_dc_shares (hotspot_key);
-
-create index idx_reward_type on gateway_dc_shares (reward_type);
+create index idx_gds_hotspot_key on gateway_dc_shares (hotspot_key);
```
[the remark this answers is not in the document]
andymck: what does `gds` stand for here ?  Also dont we need to drop the existing `idx_hotspot_key` index ?

iot_verifier/src/loader.rs
```diff
@@ -145,7 +145,7 @@ impl Loader {
             .unwrap_or(window_default_lookback)
             .max(window_max_lookback);
         let before_max = after + self.window_width;
-        let before = (now - (self.window_width * 3)).min(before_max);
+        let before = (now - (self.window_width * 3)).max(before_max);
```
andymck, line 148: i have a vague recollection of setting this to max deliberately but I have went through the logic again and cant see any reason why it should be.  

So the change is probably good but I will look at this flow further

andymck: When set to max it would have made for a very wide load window in the scenario of backloading the reports after being down for a period of time.  That would be suboptimal and resulted in a potential memory issue as it would result in all beacons and witnesses being loaded for the period.

As such this change is good

iot_verifier/src/reward_share.rs
```diff
@@ -342,22 +341,29 @@ impl GatewayShares {
...
-                witness_amount: compute_rewards(
-                    witness_rewards_per_share,
-                    reward_shares.witness_shares,
-                ),
-                dc_transfer_amount: compute_rewards(
-                    dc_transfer_rewards_per_share,
-                    reward_shares.dc_shares,
-                ),
+            .map(move |(hotspot_key, reward_shares)| {
+                tracing::info!(
```
madninja, line 348: use the structured form please

```
tracing::info!(%hotspot_key, dc_shares = ?reward_shares.dc_shares, "dc shares for hotspot");
```

Note the shorthand `%` for using Display vs `?` for Debug. Also note the shorthand of not needing the = for hotspot_key if the name for the key you're outputting is what you want to see in the output

iot_verifier/src/reward_share.rs
```diff
@@ -342,22 +341,29 @@ impl GatewayShares {
         let dc_transfer_rewards_per_share =
             rewards_per_share(total_dc_transfer_rewards_capped, total_dc_shares);
         // compute the awards per hotspot
+        tracing::info!("dc_transfer_rewards_per_share: {dc_transfer_rewards_per_share}");
```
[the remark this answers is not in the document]
andymck: noted

iot_verifier/src/reward_share.rs
andymck: this was added for debug purposes only, but noted

iot_verifier/src/reward_share.rs
```diff
@@ -237,17 +235,20 @@ impl GatewayShares {
...
+            .bind(period_end)
+            .execute(&mut *tx)
+            .await
+            .map(|_| ())?;
+
+        sqlx::query("delete from gateway_dc_shares where reward_timestamp <= $1")
+            .bind(period_end)
+            .execute(&mut *tx)
+            .await
+            .map(|_| ())
```
maplant, line 251: Personally, I would do the following: 
```suggestion
        let _ = sqlx::query("delete from gateway_shares where reward_timestamp <= $1")
            .bind(period_end)
            .execute(&mut *tx)
            .await?;

        let _ = sqlx::query("delete from gateway_dc_shares where reward_timestamp <= $1")
            .bind(period_end)
            .execute(&mut *tx)
            .await?;

        Ok(())
```

maplant (approved):

PR #437, opened 2023-04-11
Validate heartbeats for gateway ownership in mobile_verifier

maplant:


mobile_verifier/src/heartbeats.rs
```diff
@@ -145,27 +146,45 @@ pub struct SaveHeartbeatError(#[from] sqlx::Error);
...
-                reward_weight,
-                cbsd_id: heartbeat_report.cbsd_id.clone(),
-                timestamp: heartbeat_report.timestamp.naive_utc(),
-                validity,
-            }
-        })
+        heartbeats
+            .map(move |heartbeat_report| {
+                let (reward_weight, validity) =
+                    match validate_heartbeat(&heartbeat_report, &valid_epoch) {
```
vihu, line 157: Would it make sense to put the gateway owner check in the `validate_heartbeat` function itself. It already checks other errors (not_operational, bad_cbsd_id and outside_range). You could probably pass in the resolver to `validate_heartbeat` function and have it do the ownership check. That way the subsequent `.then` won't be necessary.

maplant: I considered that, but I liked how that function returned a Result. But I'll look back at in and see if there's a good way of combining them

jeffgrunewald (approved):
i could go either way on where to perform the on-chain asserted location lookup but this looks functionally correct to me

andymck (approved):

mobile_verifier/src/heartbeats.rs
```diff
@@ -145,27 +146,45 @@ pub struct SaveHeartbeatError(#[from] sqlx::Error);
...
+                    reward_weight,
+                    cbsd_id: heartbeat_report.cbsd_id.clone(),
+                    timestamp: heartbeat_report.timestamp.naive_utc(),
+                    validity,
+                }
+            })
+            .then(|mut heartbeat| {
+                let mut config_client = config_client.clone();
+                async move {
+                    // If we get back some gateway info for the given address, it's a valid address
```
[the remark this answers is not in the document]
jeffgrunewald: that's a good point; if the gateway needs to be _asserted_ we'll have to resolve the gateway to a `Some(gateway)` and then resolve the `gateway.metadata` field to a `Some(metadata)` as well. the gateway can be resolved by its address to be found on-chain but not yet asserted (if that's relevant). might not be at the moment but could/should eventually for future iterations of Mobile PoC?

PR #441, opened 2023-04-12
Separate data transfer ingest bucket

maplant:


jeffgrunewald (approved):
i think this could get confusing with the generic "ingest" name vs. the "data_transfer_ingest"; at some point would like to see "ingest" maybe renamed to "poc_ingest" similar to the way the iot verifier makes the two broad distinctions between the datasets it processes

bbalser:
> i think this could get confusing with the generic "ingest" name vs. the "data_transfer_ingest"; at some point would like to see "ingest" maybe renamed to "poc_ingest" similar to the way the iot verifier makes the two broad distinctions between the datasets it processes

If we want to change the name, now is the right time to do it.

bbalser (approved):

PR #442, opened 2023-04-12
Validate speedtest hotspots

maplant:


jeffgrunewald (approved):
only question again is whether or not we need the gateway to be _asserted_ in order to validate the reported speedtest and qualify for rewards or if the gateway just needs to registered to the chain

mobile_verifier/src/speedtests.rs
```diff
@@ -89,13 +91,21 @@ impl SpeedtestRollingAverage {
             .await;
 
         futures::stream::iter(tests_by_publickey.into_iter())
-            .then(move |(pubkey, cell_speedtests)| async move {
-                let rolling_average = exec
-                    .fetch(&pubkey)
-                    .await?
-                    .unwrap_or_else(|| SpeedtestRollingAverage::new(pubkey.clone()));
-                Ok((rolling_average, cell_speedtests))
+            .then(move |(pubkey, cell_speedtests)| {
```
bbalser, line 94: why not just do a`filter` step before this then that eliminiates any that are not valid?

maplant: The client can return an Error

maplant:
> 

I believe @madninja can verify, but it just needs to registered to the chain

bbalser (approved):

PR #443, opened 2023-04-12
Add cache to mobile_config client with default ttl of 60 minutes

bbalser:


mobile_config/src/client/mod.rs
```diff
@@ -48,23 +68,32 @@ impl gateway_info::GatewayInfoResolver for Client {
...
-                response.verify(&self.config_pubkey)?;
-                response.info.map(gateway_info::GatewayInfo::from)
-            }
-            Err(status) if status.code() == tonic::Code::NotFound => None,
-            Err(status) => Err(status)?,
-        };
-        Ok(response)
+        if let Some(cached_response) = self.cache.get(address).await {
+            Ok(cached_response.value().clone())
+        } else {
```
maplant, line 73: I think this is one place where an early return would be preferred

bbalser: Agreed cb96028

maplant (approved):

mobile_config/src/client/mod.rs
```diff
@@ -21,22 +24,39 @@ pub enum ClientError {
...
 }
 
-#[derive(Clone, Debug)]
+#[derive(Clone)]
 pub struct Client {
     pub client: mobile_config::GatewayClient<Channel>,
     signing_key: Arc<Keypair>,
     config_pubkey: PublicKey,
     batch_size: u32,
+    cache: Arc<Cache<PublicKeyBinary, Option<gateway_info::GatewayInfo>>>,
```
jeffgrunewald, line 33: the cache doesn't implement Debug i take it...

jeffgrunewald (approved):

mobile_config/src/client/mod.rs
bbalser: It does not

PR #444, opened 2023-04-12
Manually commit heartbeats and speedtests

maplant:


jeffgrunewald (approved):
were these just not getting committed at all before?

jeffgrunewald (approved):

PR #449, opened 2023-04-13
Register file-sink metric with status="ok" label

bbalser:


file_store/src/file_sink.rs
```diff
@@ -123,6 +123,7 @@ impl FileSinkBuilder {
         };
 
         metrics::register_counter!(client.metric);
+        metrics::register_counter!(client.metric, vec![OK_LABEL]);
```
jeffgrunewald, line 126: does this need both or can you just tack the `vec![OK_LABEL]` on to the existing counter registration? won't this just end up with two metrics registered, one with and one without the label as-is?

bbalser: yes, isn't that want we wanted?

jeffgrunewald: maybe it doesn't matter? i thought we more wanted a single metric that tracks the data across restarts. this feels like it will just leave a second dangling counter metrics _without_ the label that never gets used or incremented

bbalser: so just get rid of `  metrics::register_counter!(client.metric);`??

jeffgrunewald: That’s what I thought

bbalser: Removed 125aa52

jeffgrunewald (approved):

andymck (approved):

PR #454, opened 2023-04-14
Add minimum allowed balance before we disable orgs

maplant:


iot_packet_verifier/src/verifier.rs
```diff
@@ -110,11 +108,14 @@ where
                     })
                     .await
                     .map_err(VerificationError::InvalidPacketWriterError)?;
-                self.config_server
-                    .disable_org(report.oui)
-                    .await
-                    .map_err(VerificationError::ConfigError)?;
             }
+            match remaining_balance {
```
andymck, line 112: could this match and the if clause at line 86 not be merged ?

iot_packet_verifier/src/verifier.rs
```diff
@@ -110,11 +108,14 @@ where
...
+                Some(remaining_balance) if remaining_balance >= minimum_allowed_balance => {
+                    self.config_server.enable_org(report.oui).await
```
andymck, line 114: does it matter if the org is already enabled and we keep calling enable_org each time there is a balance change ?  Is there any performance or other consideration with this ?

andymck (approved):

iot_packet_verifier/src/verifier.rs
bbalser: The `config_server` has a cache of enabled_clients and only calls `iot_config` if it needs to.

bbalser (approved):

iot_packet_verifier/src/verifier.rs
maplant: Not cleanly, the match has an if guard

iot_packet_verifier/src/balances.rs
```diff
@@ -67,13 +67,13 @@ where
 {
     type Error = S::Error;
 
-    /// Debits the balance from the cache, returning true if there was enough
-    /// balance and false otherwise.
+    /// Debits the balance from the cache, returning the remaining balance if there
+    /// was enough balance and false otherwise.
```
jeffgrunewald, line 71:
```suggestion
    /// Debits the balance from the cache, returning the remaining balance as an
    /// option if there was enough and none otherwise.
```

iot_packet_verifier/src/verifier.rs
jeffgrunewald: iot_config also treats this idempotently so it's safe to call directly against the service as well

jeffgrunewald (approved):

PR #456, opened 2023-04-16
implement gateway info cache in mainline

jeffgrunewald:
implement the retainer cache on the iot-config service in mainline ahead of the solana switchover. this will reduce pressure on the database both when gateways are calling in requesting region params as well as when the iot-verifier is requesting large volumes of gateway info for evaluating beacon and witness validity.

this does beg the question if the iot-config client should still implement a cache of its own on behalf of oracle callers like the iot-verifier if the server-side is caching as well (and must implement caching since the gateways themselves cannot)


madninja:
> this does beg the question if the iot-config client should still implement a cache of its own on behalf of oracle callers like the iot-verifier if the server-side is caching as well (and must implement caching since the gateways themselves cannot)

Saving a bunch of network calls sounds like a good thing for speed?

jeffgrunewald:
> > this does beg the question if the iot-config client should still implement a cache of its own on behalf of oracle callers like the iot-verifier if the server-side is caching as well (and must implement caching since the gateways themselves cannot)
> 
> Saving a bunch of network calls sounds like a good thing for speed?

Agreed in general, but my concern would be in the two caches stacking and resulting in poc checks being made against outdated records so might be worth dialing down one or both to avoid that

iot_config/src/gateway_service.rs
```diff
@@ -36,8 +41,13 @@ impl GatewayService {
         region_map: RegionMapReader,
         auth_cache: AuthCache,
     ) -> Result<Self> {
+        let gateway_cache = Arc::new(Cache::new());
+        let cache_clone = gateway_cache.clone();
+        tokio::spawn(async move { cache_clone.monitor(4, 0.25, CACHE_EVICTION_FREQUENCY).await });
```
madninja, line 46: This needs to go into the run loop.. with something that can hopefully also tell it to shut down cleanly.. spawning in new has a bad smell

iot_config/src/gateway_service.rs
```diff
@@ -127,48 +152,26 @@ impl iot_config::Gateway for GatewayService {
...
-                        default_region
-                    }
-                };
-                (region, gain)
-            } else {
+        let (region, gain) = match self.resolve_gateway_info(address).await {
+            Err(_) => {
                 tracing::debug!(
                     pubkey = address.to_string(),
                     default_region = default_region.to_string(),
```
madninja, line 159: Not in this PR, but you could just make this line `%default_region`

iot_config/src/gateway_service.rs
jeffgrunewald: I want to explore this more in our team call tomorrow; I personally have had bad experiences with spawned tasks in these grpc handlers not shutting down but others swear tokio handles this for them cleanly

iot_config/src/gateway_service.rs
```diff
@@ -59,6 +69,25 @@ impl GatewayService {
             .map_err(|_| Status::permission_denied("invalid admin signature"))?;
         Ok(())
     }
+
+    async fn resolve_gateway_info(&self, pubkey: &PublicKeyBinary) -> Result<GatewayInfo, Status> {
```
jeffgrunewald, line 73: i'm curious about thoughts on having this instead return `Result<(GatewayInfo, bool), Status>` where the bool is whether or not we had a cache hit. since this function is hiding the result of that call it means i lose the ability to measure in the request duration histogram metric (different PR) whether or not a request was processed through the cache via a label

andymck (approved):
the implementation here looks fine but I am surprised there is a need for the gateway cache, given its only pulling data from the DB and also as the iot verifier implements its own cache which will substantially reduce the server side load from that oracle.   I believe a similar cache was also added to the mobile verifier.

Perhaps some of this can be explained by something i noticed whilst testing locally using mainnet poc data, we get a lot of beacons and witnesses from unknown gateways.  Currently the IOT verifier does not cache the response for these and as such we keep hitting the server for any next request.  I can look at addressing this and it should help to further reduce load on the server side

andymck:
> > > this does beg the question if the iot-config client should still implement a cache of its own on behalf of oracle callers like the iot-verifier if the server-side is caching as well (and must implement caching since the gateways themselves cannot)
> > 
> > 
> > Saving a bunch of network calls sounds like a good thing for speed?
> 
> Agreed in general, but my concern would be in the two caches stacking and resulting in poc checks being made against outdated records so might be worth dialing down one or both to avoid that

the client side caching eliminates the network round trip and as such will still be beneficial

bbalser (approved):

iot_config/src/gateway_service.rs
JayKickliter: Sorry, I wasn't clear on the call earlier. Despite the fact that `Cache::monitor` doesn't complete (I said it doesn't return), it does yield. So it _could_ still make sense to to also select on a shutdown monitor to ensure this thread shuts down cleanly. I don't have a strong opinion on whether spawning is good or bad in the constructor however.

PR #465, opened 2023-04-18
Use jemalloc for long-running, high-churn services

JayKickliter:
We've noticed poor-memory reuse in iot_config, particularly when updating regions. There is a step-change in memory utilization when regenerating region params, as can be expected, but memory goes up a little bit on subsequent regens, when in theory there should be enough deallocated memory laying around to handle the needs. There's no guarantee that this change will reduce memory usage, so it should be tested on a problem service. However, it would almost certainly speed up heap allocations in a tight lip, like construction regions.

Let me know if I missed some other targets. Or if we'd prefer to just add `jemalloc` to every `main.rs` in this repo.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3281,6 +3281,7 @@ dependencies = [
  "hextree",
  "http",
  "http-serde",
+ "jemallocator",
  "libflate",
  "metrics",
  "metrics-exporter-prometheus",
@@ -3316,6 +3317,7 @@ dependencies = [
  "helium-proto",
  "http",
  "http-serde",
+ "jemallocator",
  "metrics",
  "poc-metrics",
  "prost",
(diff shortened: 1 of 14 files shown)


jeffgrunewald (approved):

maplant (approved):

JayKickliter:
@jeffgrunewald Did you just rebase this branch? Have you test-deployed it?

jeffgrunewald:
> @jeffgrunewald Did you just rebase this branch? Have you test-deployed it?

yes; we have a couple of feature-critical deploys to get out the door so we haven't had a chance to test-deploy this but i did a rebase to get it back to a mergable state

JayKickliter:
I suggest testing this first if possible

PR #466, opened 2023-04-18
ensure packet ids are unique, handle dups

andymck:
The iot packet verifier can output dups, as such using only the fields contained within the verified packet will result in dups being dropped when being inserted into the DB.

This adds a now timestamp component to the packet ID field in the DB to ensure dups get inserted correctly


madninja:
hmm.. that means you'd never be able to look up a packet anymore without knowing a timestamp you can never guess?

bbalser (approved):

madninja (approved):
Approved for now with the caveat that that timestamp should go into the packet db to be able to reconstruct which packet happened where and when

PR #481, opened 2023-04-20
Monitor funds and re-enable disabled orgs

maplant:


iot_packet_verifier/src/settings.rs
```diff
@@ -46,7 +48,11 @@ pub fn default_log() -> String {
 }
 
 pub fn default_minimum_allowed_balance() -> u64 {
-    35_000_000
+    3_500_000
+}
+
+pub fn default_monitor_funds_period() -> u64 {
```
jeffgrunewald, line 54: seconds or minutes?

iot_packet_verifier/src/verifier.rs
```diff
@@ -180,74 +182,160 @@ pub trait ConfigServer {
...
+        self.disabled_clients.remove(&oui);
+        let mut req = OrgEnableReqV1 {
+            oui,
+            timestamp: Utc::now().timestamp_millis() as u64,
+            signer: self.keypair.public_key().into(),
+            signature: vec![],
+        };
+        let signature = self.keypair.sign(&req.encode_to_vec())?;
+        req.signature = signature;
+        let _ = self.client.enable(req).await?;
```
jeffgrunewald, line 221: we should make a note to have the responses validated at some point but not necessary for this PR

jeffgrunewald (approved):
having the default "re-enable" monitor period be in minutes without a comment or using "min" in the name is waiting to trip someone up (especially when i think we typically use seconds as our duration number).
i would minimum make a comment or name that setting to indicate minutes is the expect time scale but preferably i would document that setting is expected in seconds and make the default settings function `60 * 30`

iot_packet_verifier/src/verifier.rs
maplant: definitely

iot_packet_verifier/src/settings.rs
maplant: minutes

jeffgrunewald (approved):

iot_packet_verifier/src/verifier.rs
```diff
@@ -180,74 +182,150 @@ pub trait ConfigServer {
...
+                    .await
+                    .client
+                    .list(OrgListReqV1 {})
+                    .await
+                    .map_err(OrgClientError::RpcError)?
+                    .into_inner();
+                for org in orgs.orgs {
+                    if org.locked {
+                        disabled_clients.insert(org.oui);
+                        let payer = client.fetch_org(org.oui, &mut org_cache).await?;
```
jeffgrunewald, line 252: i don't see you using `disabled_clients` anywhere, you're just inserting into it and then dropping it at the end of each loop?
also, i don't think you need the outer `org_cache` any longer since you're calling `org list` on each run and that's sending you back the org record, including the `payer` field you're using to query solana. unlike the other, you _are_ using this cache from one run to the next, but since you're always calling the `org list` command and getting back the same info you need the cache isn't doing anything for you

bbalser (approved):

bbalser (approved):

PR #488, opened 2023-04-24
migrate skf under route

jeffgrunewald:
accommodate changes on the iot-config service to migrate storage and organization of session key filters from the oui level to the route level.

requires merging of https://github.com/helium/proto/pull/319 and update of the oracle cargo toml


iot_config/src/route_service.rs
```diff
@@ -135,6 +137,36 @@ impl RouteService {
...
+                    Status::invalid_argument(format!("unable to parse route_id: {route_id}"))
+                }
+                _ => Status::internal("error retrieving devaddrs for route"),
+            })?
+            .filter_map(|range| async move { range.ok() })
+            .collect()
+            .await;
+
+        'update_addrs: for update in updates {
+            for range in &ranges {
```
bbalser, line 158: You can avoid the `continue` by using `any` on an iterator rather than the inside for loop.

jeffgrunewald:
This change is tested and working correctly but it is blocked on https://github.com/helium/oracles/pull/493 which introduced a protobuf change that is preventing this branch from passing clippy lint due to the main branch still using a field that has been marked deprecated as of https://github.com/helium/proto/pull/322

andymck (approved):
lgtm

iot_config/src/route_service.rs
```diff
@@ -677,8 +895,8 @@ struct DevAddrEuiValidator {
 
 #[derive(thiserror::Error, Debug)]
 enum DevAddrEuiValidationError {
-    #[error("devaddr range outside of constraint bounds {0}")]
-    RangeOutOfBounds(String),
+    #[error("devaddr or range outside of bounds {0}")]
```
michaeldjeffrey, line 898: From what I can follow, this error is only used when a Route's Devaddr Range falls outside the Org's constraints. 
So I _think_ this might be a typo?

michaeldjeffrey (approved):

iot_config/src/route_service.rs
jeffgrunewald: not a typo, just a holdover. when i started out i was going unify all of these under a single update validator but i switched to handling the skf updates differently since they aren't being streamed; i just forgot to set this error message back to what it had originally been. good catch

PR #489, opened 2023-04-24
introducing lamport balance monitor for packet verifiers

jeffgrunewald:
allow monitoring the lamport balance of oracles that burn lamports in order to process transactions directly on the solana chain. this will allow the oracles to avoid running too low on a balance to maintain operations without advance notice allowing the service maintainer to top up the account

### DIFF
--- a/solana/src/balance_monitor.rs
+++ b/solana/src/balance_monitor.rs
@@ -9,7 +9,8 @@ pub async fn start(
     app_account: &str,
     solana: Option<Arc<SolanaRpc>>,
     shutdown: triggered::Listener,
-) -> Result<futures::future::BoxFuture<'static, Result<(), tokio::task::JoinError>>, SolanaRpcError> {
+) -> Result<futures::future::BoxFuture<'static, Result<(), tokio::task::JoinError>>, SolanaRpcError>
+{
     Ok(match solana {
         None => Box::pin(async move { Ok(()) }),
         Some(rpc_client) => {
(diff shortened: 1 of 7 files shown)


bbalser (approved):

solana/src/balance_monitor.rs
```diff
@@ -0,0 +1,61 @@
...
+                match handle.await {
+                    Ok(()) => Ok(()),
+                    Err(err) => Err(Error::from(err)),
+                }
+            })
+        }
+    })
+}
+
+async fn run(metric_name: String, solana: Arc<SolanaRpc>, shutdown: triggered::Listener) {
```
maplant, line 35: You should be able to make these generic over `SolanaNetwork` and not have to match on an Option

maplant: ehhh I guess not

solana/src/balance_monitor.rs
```diff
@@ -0,0 +1,61 @@
...
+    loop {
+        let shutdown = shutdown.clone();
+
+        tokio::select! {
+            _ = shutdown => {
+                tracing::info!("sol monitor: shutting down");
+                break
+            }
+            _ = trigger.tick() => {
+                let verifier_pubkey = match Keypair::from_bytes(&solana.keypair) {
```
maplant, line 47: why do this every tick?

bbalser (approved):

jeffgrunewald:
Screenshot 2023-04-24 at 10 39 30 PM
With this change, the `iot_packet_verifier_sol_balance` metric can be seen here being exported

andymck (approved):

bbalser (approved):

PR #494, opened 2023-04-25
Update cache when we re-enable

maplant:


jeffgrunewald (approved):

iot_packet_verifier/src/verifier.rs
```diff
@@ -174,9 +182,89 @@ pub trait ConfigServer {
...
+
+#[async_trait]
+impl BalanceStore for crate::balances::BalanceStore {
+    async fn set_balance(&self, payer: &PublicKeyBinary, balance: u64) {
+        self.lock().await.entry(payer.clone()).or_default().balance = balance;
+    }
+}
+
+#[async_trait]
+impl BalanceStore for Arc<Mutex<HashMap<PublicKeyBinary, u64>>> {
```
jeffgrunewald, line 253:
```suggestion
// differs from the BalanceStore in the value stored in the contained HashMap; a u64 here instead of a Balance {} struct
impl BalanceStore for Arc<Mutex<HashMap<PublicKeyBinary, u64>>> {
```
took me a moment to realize why we need this separate trait impl

jeffgrunewald (approved):

andymck (approved):

iot_packet_verifier/src/verifier.rs
```diff
@@ -163,9 +163,17 @@ impl Debiter for Arc<Mutex<HashMap<PublicKeyBinary, u64>>> {
     }
 }
 
+// TODO: Move these to a separate module
```
andymck, line 166: this todo still applicable ?

maplant: Yeah, I should move all the config client stuff into its own module. But later

PR #497, opened 2023-04-26
halt rewards if data below threshold

andymck:
This adds a threshold check to the rewards generation.  If the volume of any of the 3 rewardable entities ( beacons, witnesses or packets ) for the rewardable period is below a minimum level then the rewards will be suspended for the epoch.  

The minimum volumes are derived from the average volume for each entity type over the last 90 days.   Each time rewards are run we store the volumes for that epoch in the DB and thus feeds into the dataset for the average calculations.

When rewards are suspended, they will be retried every hour until a max duration ( defaults to 3 hours but customisable via settings ) has passed.  At that point the rewards will proceed as normal for the original epoch *unless* the key 'halt_rewards_override' in the meta table has been set too true.  The rewards will not run whilst halt_rewards_override is 'true'.

The intent with this change is to suspend rewards if volume is unexpectedly low and to allow the dataset to be reviewed manually.  Rather than stopping the verifier and forcing a backlog of loader events, the verifier will continue other operations and provide a window for manual review and possible intervention. Once that window expires the verifier will assume all is good and subsequently process the rewards.  If manual review highlights some sort of issue with which we should not process the rewards, the user can intervene and set the meta key `halt_rewards_override` too true.  This will force rewards to remain suspended until the key is subsequently set to false ( or a non true value ).

### DIFF
--- a/iot_verifier/migrations/12_bootstrap_reward_halt_override.sql
+++ b/iot_verifier/migrations/12_bootstrap_reward_halt_override.sql
@@ -1,0 +1,17 @@
+CREATE TYPE reward_type as enum('beacon', 'witness', 'packet');
+
+insert into meta
+    (key, value)
+values
+    ('halt_rewards_override', 'false')
+on conflict
+    (key)
+do nothing;
+
+CREATE TABLE rewards_history (
+    reward_type reward_type,
+    count bigint NOT NULL,
+    epoch_ts timestamptz NOT NULL,
+    created_at timestamptz  NOT NULL default now()
+);
+
--- a/iot_verifier/src/rewarder.rs
+++ b/iot_verifier/src/rewarder.rs
@@ -3,6 +3,7 @@ use std::time::Duration;
 use crate::{
     reward_share::{operational_rewards, GatewayShares},
     scheduler::Scheduler,
+    Settings,
 };
 use chrono::{DateTime, Duration as ChronoDuration, TimeZone, Utc};
 use db_store::meta;
@@ -12,18 +13,11 @@ use helium_proto::RewardManifest;
 use price::PriceTracker;
 use rust_decimal::prelude::*;
 use rust_decimal_macros::dec;
+use sqlx::PgPool;
 use sqlx::{PgExecutor, Pool, Postgres};
 use tokio::time::sleep;
 
 const HALT_REWARDS_DB_KEY: &str = "halt_rewards_override";
-// set min counts for the number of beacon, witnesses and packets we expect
-// to be within a rewardable period
-// note: these are floor levels, the averages are expected to be higher
-//       but if we drop below the levels defined here, rewards will be delayed
-//       enabling a manual review
-const BEACON_MIN_REWARDABLE_COUNT: u64 = 800_000; // 200K gateways @ 4 beacons per day
-const WITNESS_MIN_REWARDABLE_COUNT: u64 = 4_000_000; // 5 witnesses per above beacon
-const PACKET_MIN_REWARDABLE_COUNT: u64 = 100_000; // TODO: work out a sensible packet floor level
 const AVG_SCALING_FACTOR: Decimal = dec!(0.8);
 
 pub struct Rewarder {
@@ -34,6 +28,9 @@ pub struct Rewarder {
     pub reward_offset: ChronoDuration,
     pub rewards_retry_interval: Duration,
     pub rewards_max_delay_duration: ChronoDuration,
+    pub beacon_min_rewardable_count: u64,
+    pub witness_min_rewardable_count: u64,
+    pub packet_min_rewardable_count: u64,
 }
 
 #[derive(sqlx::Type, Debug, Clone, PartialEq, Eq, Hash)]
@@ -45,6 +42,26 @@ pub enum RewardType {
 }
 
 impl Rewarder {
+    pub async fn from_settings(
+        settings: &Settings,
+        pool: PgPool,
+        rewards_sink: file_sink::FileSinkClient,
+        reward_manifests_sink: file_sink::FileSinkClient,
+    ) -> anyhow::Result<Self> {
+        Ok(Self {
+            pool,
+            rewards_sink,
+            reward_manifests_sink,
+            reward_period_hours: settings.rewards,
+            reward_offset: settings.reward_offset_duration(),
+            rewards_retry_interval: settings.rewards_retry_interval(),
+            rewards_max_delay_duration: settings.rewards_max_delay_duration(),
+            beacon_min_rewardable_count: settings.beacon_min_rewardable_count,
+            witness_min_rewardable_count: settings.witness_min_rewardable_count,
+            packet_min_rewardable_count: settings.packet_min_rewardable_count,
+        })
+    }
+
     pub async fn run(
         mut self,
         price_tracker: PriceTracker,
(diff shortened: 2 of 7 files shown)


iot_verifier/src/rewarder.rs
```diff
@@ -120,8 +189,72 @@ impl Rewarder {
...
+        let duration_since_last_rewards_end =
+            scheduler.time_since_last_reward_period_ended(Utc::now());
+        let halt_rewards_override = fetch_halt_rewards_override(&self.pool).await?;
+        // get the current averages from historic data in the db
+        let (scaled_beacon_avg, scaled_witness_avg, scaled_packet_avg) =
+            self.get_historic_averages().await?;
+
+        tracing::info!(%gateway_reward_shares.total_beacon_count,
+            %gateway_reward_shares.total_witness_count,
+            %gateway_reward_shares.total_packet_count,
```
jeffgrunewald, line 210: how does the "key = value" short-hand syntax treat these when they're struct fields being dot-accessed directly?

iot_verifier/src/rewarder.rs
```diff
@@ -141,3 +274,108 @@ async fn save_rewarded_timestamp(
...
+        sqlx::query(
+            r#"
+        INSERT INTO rewards_history(reward_type, count, epoch_ts) VALUES($1, $2, $3)
+        "#,
+        )
+        .bind(reward_type)
+        .bind(count)
+        .bind(epoch_ts)
+        .execute(tx)
+        .await?;
```
jeffgrunewald, line 344:
```suggestion
        beacon_count: i64,
        witness_count: i64,
        packet_count: i64,
        epoch_ts: DateTime<Utc>,
    ) -> anyhow::Result<()> {
        sqlx::query(
            r#"
        INSERT INTO rewards_history (reward_type, count, epoch_ts) VALUES ('beacon', $1, $4), ('witness', $2, $4), ('packet', $3, $4)
        "#,
        )
        .bind(beacon_count)
        .bind(witness_count)
        .bind(packet_count)
        .bind(epoch_ts)
        .execute(tx)
        .await?;
```
since this is a fixed operation you could do this with a single DB call

iot_verifier/src/rewarder.rs
```diff
@@ -141,3 +274,108 @@ async fn save_rewarded_timestamp(
...
+        .bind(reward_type)
+        .fetch_one(db)
+        .await?;
+        let result = (result * AVG_SCALING_FACTOR).to_u64().unwrap_or(default);
+        if result < default {
+            Ok(default)
+        } else {
+            Ok(result)
+        }
+    }
```
jeffgrunewald, line 368:
```suggestion
    pub async fn get_avgs(
        db: impl sqlx::PgExecutor<'_>,
        default_beacon: u64,
        default_witness: u64,
        default_packet: u64
    ) -> anyhow::Result<(u64, u64, u64)> {
        let row = sqlx::query(
            r#"
        SELECT COALESCE(AVG(count), 0) FROM rewards_history where reward_type = 'beacon' as beacon_avg,
            COALESCE(AVG(count), 0) FROM rewards_history where reward_type = 'witness' as witness_avg,
            COALESCE(AVG(count), 0) FROM rewards_history where reward_type = 'packet' as packet_avg,
        "#,
        )
        .fetch_one(db)
        .await?;
        let beacon_avg = row.try_get::<Decimal, &str>("beacon_avg").map(|ba| (ba * AVG_SCALING_FACTOR).to_u64()).unwrap_or(default_beacon);
        let witness_avg = row.try_get::<Decimal, &str>("witness_avg").map(|wa| (wa * AVG_SCALING_FACTOR).to_u64()).unwrap_or(default_witness);
        let packet_avg = row.try_get::<Decimal, &str>("packet_avg").map(|pa| (pa * AVG_SCALING_FACTOR).to_u64()).unwrap_or(default_packet);
        Ok(
            std::cmp::max(beacon_avg, default_beacon),
            std::cmp::max(witness_avg, default_witness),
            std::cmp::max(packet_avg, default_packet),
        )
    }
```
you should be able to get the averages out with a single query and simplify that calculation as well. may be able to do it _even_ more simply by querying against the rewards_history table with a single select and using group_by but not sure about that one

iot_verifier/src/settings.rs
```diff
@@ -2,6 +2,7 @@ use chrono::Duration;
 use config::{Config, Environment, File};
 use serde::Deserialize;
 use std::path::Path;
+use std::time::Duration as StdDuration;
```
jeffgrunewald, line 5: when we have multiple types of duration in the same module should we be consistent with which ones we alias and which ones we take from the original name? in the other module we're aliasing `chrono::Duration as ChronoDuration` but here we're flipping that

iot_verifier/src/rewarder.rs
```diff
@@ -69,10 +111,23 @@ impl Rewarder {
         &mut self,
         scheduler: &Scheduler,
         iot_price: Decimal,
-    ) -> anyhow::Result<()> {
+    ) -> anyhow::Result<Duration> {
```
jeffgrunewald, line 114: not sure how i feel about this function taking on the additional responsibility of processing rewards _and_ computing the delay duration until the next attempt. since it's already returning a result, it seems like it would be more semantically correct to keep the function focused entirely on that and have it instead return a std::Result where the error could include any of the previous errors + "RewardsHalted" and/or "RewardsDelayed" and then compute the sleep duration outside of this function the way it currently does in main. In either case the rewards method still needs to be passed the Scheduler struct, so it _can_ compute the next duration, but not sure it _should_

iot_verifier/src/rewarder.rs
```diff
@@ -120,8 +189,72 @@ impl Rewarder {
...
+        let sleep_duration = scheduler.sleep_until_next_epoch(Utc::now())?;
+        Ok(sleep_duration)
+    }
 
-        Ok(())
+    async fn threshold_check(
+        &self,
+        gateway_reward_shares: &GatewayShares,
+        scheduler: &Scheduler,
+    ) -> anyhow::Result<bool> {
```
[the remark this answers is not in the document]
andymck: well this call is fallible, hence the result

iot_verifier/migrations/12_bootstrap_reward_halt_override.sql
```diff
@@ -0,0 +1,17 @@
...
+    ('halt_rewards_override', 'false')
+on conflict
+    (key)
+do nothing;
+
+CREATE TABLE rewards_history (
+    reward_type reward_type,
+    count bigint NOT NULL,
+    epoch_ts timestamptz NOT NULL,
+    created_at timestamptz  NOT NULL default now()
```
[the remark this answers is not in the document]
andymck: the records do not get updated after the initial insert

iot_verifier/src/rewarder.rs
```diff
@@ -120,8 +189,72 @@ impl Rewarder {
...
+
+        tracing::info!(%gateway_reward_shares.total_beacon_count,
+            %gateway_reward_shares.total_witness_count,
+            %gateway_reward_shares.total_packet_count,
+            %scaled_beacon_avg,
+            %scaled_witness_avg,
+            %scaled_packet_avg,
+            %halt_rewards_override);
+
+        // if we are above our thesholds then we are good to reward
```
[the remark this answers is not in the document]
andymck: the rewards volume would need to be greater than the computed average - 20%, hence it is a 20% buffer

iot_verifier/src/rewarder.rs
andymck: i had something like that in an earlier version and reverted to individual queries but i think the code reads better as it is currently and its not like this flow needs to be optimised.  The queries are run once per day under normal conditions or once per hour for a fixed number of hours when rewards are suspended

iot_verifier/src/rewarder.rs
andymck: i agree the current incarnation is not ideal, i added this originally as i didnt want to deal with propagating errors at the time.  Meant to come back to it and then got used to it

iot_verifier/src/settings.rs
andymck: I took the approach of aliasing the minority usage

andymck:
Closing this as madninja doesnt want to proceed with self healing techniques at this time.  Instead focus will be on metrics and actioning those metrics

PR #502, opened 2023-05-02
filter hotspots not registered on chain from data transfer rewards

jeffgrunewald:
addresses an edge case in the the flow of data verification from the iot-packet-verifier to the iot-verifier wherein an otherwise valid iot packet report may have had its data routed by a gateway not registered to the solana chain, thereby making the packet report valid but the gateway still invalid for claiming rewards for the transfer.

gateways may receive data transfer rewards without having been _location asserted_ but they may not if they have never even registered on solana and received an on-chain identity.

### DIFF
--- a/iot_verifier/src/main.rs
+++ b/iot_verifier/src/main.rs
@@ -182,7 +182,7 @@ impl Server {
             ),
             entropy_loader.run(entropy_loader_receiver, &shutdown),
             loader.run(&shutdown, &gateway_cache),
-            packet_loader.run(pk_loader_receiver, &shutdown),
+            packet_loader.run(pk_loader_receiver, &shutdown, &gateway_cache),
             purger.run(&shutdown),
             rewarder.run(price_tracker, &shutdown),
             density_scaler.run(&shutdown).map_err(Error::from),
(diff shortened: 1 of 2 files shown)


madninja:
> addresses an edge case in the the flow of data verification from the iot-packet-verifier to the iot-verifier wherein an otherwise valid iot packet report may have had its data routed by a gateway not registered to the solana chain, thereby making the packet report valid but the gateway still invalid for claiming rewards for the transfer.
> 

don't use "claiming". The gateway can't be rewarded because it is not on chain with an owner to receive data rewards. 

> gateways may receive data transfer rewards without having been _location asserted_ but they may not if they have never even registered on solana and received an on-chain identity.

andymck (commented):
A concern I have with this is that we have no paper trail of a valid packet being discarded from rewards,  the current iot_verifier's packet loader assumes that all valid packets are going to be rewarded...that was the contract with the iot packet verifier, it outputs valid packets the iot verifier rewards those valid packets.

If a question was asked at a later stage ( such as by the gateway owner ) as to why their data packets were not rewarded it may be time consuming to determine the reason (especially if the gateway was subsequently asserted )

The need for a paper trail is also reinforced here when it is considered that any decisions that rely on the config service data is unlikely to be replayable at a later data, ie we cant rerun the data and be sure to get the same decision again ( for example if the gateway has been asserted/registered after the first run of the data )

Adding a new output file such as 'non_rewardable_packet' should be considered here, leaving an auditable paper trail.

andymck (approved):

PR #504, opened 2023-05-02
Bump to latest hextree release

JayKickliter:
HexTree still has some minor churn, and this PR updates oracles to the latest v0.3.1 release with the goal of making sure oracles doesn't camp on an old version for too long. This should have no _noticeable_ effect on performance.

### DIFF
--- a/iot_config/src/region_map.rs
+++ b/iot_config/src/region_map.rs
@@ -46,7 +46,7 @@ impl RegionMap {
     }
 
     pub fn get_region(&self, location: Cell) -> Option<Region> {
-        self.region_hextree.get(location).cloned()
+        self.region_hextree.get(location).unzip().1.copied()
     }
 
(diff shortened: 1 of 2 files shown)


iot_config/src/region_map.rs
```diff
@@ -46,7 +46,7 @@ impl RegionMap {
     }
 
     pub fn get_region(&self, location: Cell) -> Option<Region> {
-        self.region_hextree.get(location).cloned()
+        self.region_hextree.get(location).unzip().1.copied()
```
JayKickliter, line 49: `get` now returns the actual cell, which is likely a parent cell of `location`. Since we don't seem to need that information, `unzip().1` throws away the actual cell and possible gets optimized away.

madninja (approved):

jeffgrunewald (approved):

PR #505, opened 2023-05-03
converting helius-fed database to new foundation schema

jeffgrunewald:
the new database being fed from the solana chain stores the helium crypto public key address of the hotspots as a binary buffer of the b58 decoded address string, as well as splitting the hotspot metadata info between the `iot_hotspot_infos` table where the metadata is stored and the `key_to_assets` table where the encoded pubkey address is stored. this change enables querying the new schema structure by the iot_config server (mobile_config server pending) without changing the RPC API the rest of the oracles use to lookup or stream hotspot on-chain metadata

### DIFF
--- a/iot_config/src/gateway_info.rs
+++ b/iot_config/src/gateway_info.rs
@@ -130,6 +130,7 @@ pub(crate) mod db {
     use futures::stream::{Stream, StreamExt};
     use helium_crypto::PublicKeyBinary;
     use sqlx::{PgExecutor, Row};
+    use std::str::FromStr;
 
     pub struct IotMetadata {
         pub address: PublicKeyBinary,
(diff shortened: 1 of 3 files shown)


iot_config/src/gateway_info.rs
```diff
@@ -143,13 +144,16 @@ pub(crate) mod db {
         db: impl PgExecutor<'_>,
         address: &PublicKeyBinary,
     ) -> anyhow::Result<Option<IotMetadata>> {
+        let entity_key = bs58::decode(address.to_string()).into_vec()?;
```
andymck, line 147: are you sure this gives you what you need ? seems odd that the a public key in a format not directly supported by crypto-rs is being used here

andymck (approved):
looks ok to me but i havent seen the schema so making some assumptions here

iot_config/src/gateway_info.rs
jeffgrunewald: I’d like to go through it with folks today but I’ve tested this locally with a dump of data and this is unfortunately what we need ☹️

bbalser (approved):

PR #506, opened 2023-05-03
write out non rewardable packet paper trail to s3

andymck:
depends on: https://github.com/helium/proto/pull/325

Writes out a paper trail to S3 for verified packets ( outputted by the iot packet verifier ) that are deemed not rewardable.  Currently this would only be in the scenario whereby the gateway does not exist but may in the future extend to additional scenarios ( such as duplicate or free packets )

Also adds metrics to the packet loader

### DIFF
--- a/iot_verifier/src/packet_loader.rs
+++ b/iot_verifier/src/packet_loader.rs
@@ -1,11 +1,22 @@
-use crate::{gateway_cache::GatewayCache, reward_share::GatewayDCShare};
-use file_store::{file_info_poller::FileInfoStream, iot_packet::IotValidPacket};
+use crate::{
+    gateway_cache::GatewayCache, metrics::LoaderMetricTracker, reward_share::GatewayDCShare,
+    Settings,
+};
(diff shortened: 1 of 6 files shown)


iot_verifier/src/packet_loader.rs
```diff
@@ -17,52 +28,113 @@ pub enum NewLoaderError {
...
+                FileType::NonRewardablePacket,
+                store_base_path,
+                concat!(env!("CARGO_PKG_NAME"), "_non_rewardable_packet"),
+                shutdown.clone(),
+            )
+            .deposits(Some(file_upload_tx.clone()))
+            .roll_time(ChronoDuration::minutes(5))
+            .create()
+            .await?;
+        tokio::spawn(async move { non_rewardable_packet_server.run().await });
```
jeffgrunewald, line 57:
```suggestion
        tokio::spawn(non_rewardable_packet_server.run());
```
i believe you can remove the async block when you're only doing a single awaitable async operation within the spawned task

jeffgrunewald (approved):

PR #507, opened 2023-05-03
Real time verification of mobile heartbeats and speedtests

maplant:


mobile_verifier/src/cli/server.rs
```diff
@@ -1,10 +1,11 @@
-use crate::{
-    verifier::{Verifier, VerifierDaemon},
-    Settings,
-};
+use crate::{verifier::VerifierDaemon, Settings};
```
andymck, line 1: remind me, why is this module hosted under the cli dir ? 

nit: why is it called server.rs, i thought we previously moved away from this convention and named modules based on task/activity

mobile_verifier/src/verifier.rs
```diff
@@ -1,148 +1,219 @@
...
+                        .monitor(4, 0.25, std::time::Duration::from_secs(60 * 60 * 3))
+                        .await
+                });
+
+                loop {
+                    let Some(file) = heartbeats.recv().await else {
+                        bail!("Heartbeat report file stream was dropped unexpectedly");
+                    };
+                    let epoch =
+                        (file.file_info.timestamp - Duration::hours(1))..file.file_info.timestamp;
```
andymck, line 80: I see this range is used to validate if the heartbeat received_timestamp is within its range but under what circumstances would the received_timestamp be outside of this range, i mean if the roll up files are written out every 5 mins or so how could the received_timestamp not be in this range ?

mobile_verifier/src/verifier.rs
```diff
@@ -1,148 +1,223 @@
...
+            mobile_rewards,
+            reward_manifests,
+            price_tracker,
+            data_transfer_ingest,
+            reward_period_hours,
+        } = self;
+
+        let heartbeat_pool = pool.clone();
+        let heartbeat_config_client = config_client.clone();
+        let mut heartbeat_verification_task: JoinHandle<anyhow::Result<()>> =
```
andymck, line 64: Would it not be better structure to break each task out to its own module, ie a heartbeat loader, a speedtest loader, a rewarder.  Grouping them all into a single module makes the code somewhat more difficult to follow and maintain

mobile_verifier/src/verifier.rs
```diff
@@ -1,148 +1,223 @@
...
+                    let reports = file.into_stream(&mut transaction).await?;
+
+                    let mut validated_heartbeats = pin!(
+                        Heartbeat::validate_heartbeats(&heartbeat_config_client, reports, &epoch)
+                            .await
+                    );
+
+                    while let Some(heartbeat) = validated_heartbeats.next().await.transpose()? {
+                        heartbeat.write(&valid_heartbeats).await?;
+                        /*
```
andymck, line 91: should these commented out sections relating to cache usage be removed ?

mobile_verifier/src/reward_shares.rs
```diff
@@ -586,9 +558,9 @@ mod test {
         owners.insert(gw8.clone(), owner4.clone());
 
         // init cells and cell_types
-        let c1 = "P27-SCE4255W2107CW5000014".to_string();
+        // let c1 = "P27-SCE4255W2107CW5000014".to_string();
```
jeffgrunewald, line 561:
```suggestion
```

mobile_verifier/src/speedtests.rs
```diff
@@ -423,7 +438,7 @@ pub enum FetchError {
 #[async_trait::async_trait]
 pub trait SpeedtestStore {
     async fn fetch(
-        self,
+        &mut self,
```
jeffgrunewald, line 441: this trait fully deprecated?

mobile_verifier/src/verifier.rs
```diff
@@ -179,21 +249,24 @@ impl VerifierDaemon {
         let written_files = self.mobile_rewards.commit().await?.await??;
 
         let mut transaction = self.pool.begin().await?;
-        // Clear the heartbeats table:
-        sqlx::query("TRUNCATE TABLE heartbeats;")
+
+        // Clear the heartbeats of old heartbeats:
```
jeffgrunewald, line 253:
```suggestion
        // Clear the heartbeats table of old heartbeats:
```

mobile_verifier/src/verifier.rs
```diff
@@ -179,21 +249,24 @@ impl VerifierDaemon {
...
+        sqlx::query("DELETE FROM heartbeats WHERE truncated_timestamp < $3")
```
jeffgrunewald, line 254: should this be `$1`? will postgres automatically convert the input timestamp to a truncated timestamp (i.e. they are comparable?)

mobile_verifier/src/heartbeats.rs
```diff
@@ -52,87 +24,52 @@ pub struct HeartbeatReward {
...
-                .or_default();
-            entry.reward_weight = heartbeat.reward_weight;
-            entry.hours_seen[heartbeat.timestamp.hour() as usize] = true;
-        }
+impl HeartbeatReward {
+    pub fn validated<'a>(
+        exec: impl sqlx::PgExecutor<'a> + Copy + 'a,
+        epoch: &'a Range<DateTime<Utc>>,
+    ) -> impl Stream<Item = Result<HeartbeatReward, sqlx::Error>> + 'a {
+        sqlx::query_as::<_, HeartbeatKey>(
```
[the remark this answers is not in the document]
bbalser: The insert statement does ensure that there is only ever one record per radio per hour, but I agree this does feel like we could do this without causing N+1 queries.

mobile_verifier/src/verifier.rs
bbalser: This cache still feels very valuable as it will stop us from hitting the database for a large number of heartbeats.

mobile_verifier/src/settings.rs
```diff
@@ -69,7 +60,9 @@ impl Settings {
             .and_then(|config| config.try_deserialize())
     }
 
-    pub fn verification_offset_duration(&self) -> Duration {
-        Duration::minutes(self.verification_offset_minutes)
+    pub fn start_after(&self) -> DateTime<Utc> {
```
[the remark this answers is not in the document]
bbalser: this setting sets the FileInfoPoller look back behavior, so it will use this value if there are no records in the `files_processed` table to tell it where to start from.

mobile_verifier/src/heartbeats.rs
```diff
@@ -24,6 +32,77 @@ pub struct HeartbeatReward {
...
+            tokio::spawn(async move {
+                cache_clone
+                    .monitor(4, 0.25, std::time::Duration::from_secs(60 * 60 * 3))
+                    .await
+            });
+
+            loop {
+                tokio::select! {
+                    _ = shutdown.clone() => break,
+                    Some(file) = self.heartbeats.recv() => {
```
andymck, line 71: be good to push the handling out into its own function

mobile_verifier/src/speedtests.rs
```diff
@@ -41,6 +53,63 @@ impl Speedtest {
...
+            file_sink,
+        }
+    }
+
+    pub async fn run(mut self, shutdown: triggered::Listener) -> anyhow::Result<()> {
+        tokio::spawn(async move {
+            loop {
+                tokio::select! {
+                    _ = shutdown.clone() => break,
+                    Some(file) = self.speedtests.recv() => {
```
andymck, line 83: as with heartbeats, be good to have the handling of the reports in its own function

mobile_verifier/src/verifier.rs
bbalser: Fixed 9962212

mobile_verifier/src/speedtests.rs
bbalser: Addressed 0c7e805

mobile_verifier/src/heartbeats.rs
```diff
@@ -84,7 +97,8 @@ impl HeartbeatDaemon {
         file: FileInfoStream<CellHeartbeatIngestReport>,
         cache: &Arc<Cache<(String, DateTime<Utc>), ()>>,
     ) -> anyhow::Result<()> {
-        let epoch = (file.file_info.timestamp - Duration::hours(1))..file.file_info.timestamp;
+        let epoch = (file.file_info.timestamp - Duration::hours(3))
```
jeffgrunewald, line 100: any special reason this is a 33 hour window or was that intended to be `Duration::hours(3)...Duration::minutes(30)`?

andymck:
Something to note, this adds realtime processing of heartbeats and speedtest but does not do same for data transfer sessions ?  Is that deliberate or an oversight ?

mobile_verifier/src/speedtests.rs
maplant: let me try to bring something like this back for tests

mobile_verifier/src/heartbeats.rs
```diff
@@ -52,87 +24,52 @@ pub struct HeartbeatReward {
...
-        T: IntoIterator<Item = Heartbeat>,
-    {
-        let mut heartbeats = Self::default();
-        heartbeats.extend(iter);
-        heartbeats
-    }
+#[derive(Clone)]
+pub struct Heartbeat {
+    pub cbsd_id: String,
+    pub cell_type: Option<CellType>,
```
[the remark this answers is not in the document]
maplant: yes, but only if the validity is `HeartbeatValidity::BadCbsdId`, so it will be written as an invalid heartbeat

mobile_verifier/src/rewarder.rs
```diff
@@ -0,0 +1,226 @@
...
+            .ok_or(db_store::Error::DecodeError)
+    }
+
+    async fn next_rewarded_end_time(&self) -> db_store::Result<DateTime<Utc>> {
+        Utc.timestamp_opt(meta::fetch(&self.pool, "next_rewarded_end_time").await?, 0)
+            .single()
+            .ok_or(db_store::Error::DecodeError)
+    }
+
+    async fn disable_complete_data_checks_until(&self) -> db_store::Result<DateTime<Utc>> {
```
bbalser, line 88: Shouldn't we default this if the key is not found in the meta table?

mobile_verifier/src/heartbeats.rs
```diff
@@ -213,68 +138,55 @@ impl Heartbeat {
...
-                                  heartbeats.hotspot_key
-                          END,
-            reward_weight = EXCLUDED.reward_weight
-            RETURNING (xmax = 0) as inserted;
-            "#,
+        let truncated_timestamp = self.truncated_timestamp()?;
+        Ok(
+            sqlx::query_as::<_, HeartbeatSaveResult>(
+                r#"
+                INSERT INTO heartbeats (cbsd_id, hotspot_key, cell_type, latest_timestamp, truncated_timestamp)
```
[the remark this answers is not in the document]
maplant: Exactly

mobile_verifier/src/heartbeats.rs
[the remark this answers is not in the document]
maplant: Fixed by @bbalser

mobile_verifier/src/rewarder.rs
maplant: I have two opinions about this. The first is that we should probably _not_ default it, as it's a very important setting. The second is that _if_ we want to default this, we might as well default it by inserting the key in a migration

mobile_verifier/src/heartbeats.rs
maplant: Indeed, it is supposed to be 30 minutes

mobile_verifier/src/rewarder.rs
bbalser: I can get behind setting a default via migration, but I think we have to have a default or catch the error where it does not exist.

maplant: Added default in a new migration

mobile_verifier/src/rewarder.rs
```diff
@@ -0,0 +1,249 @@
...
+                next_rewarded_end_time,
+                self.reward_offset,
+            );
+            let now = Utc::now();
+            let sleep_duration = if scheduler.should_reward(now) {
+                if self.is_data_current(&scheduler.reward_period).await? {
+                    self.reward(&scheduler).await?;
+                    continue;
+                } else {
+                    tracing::info!(
```
jeffgrunewald, line 69: do we log at `info` level by default? if not this should probably be a warning

mobile_verifier/src/rewarder.rs
```diff
@@ -0,0 +1,249 @@
...
+    }
+
+    pub async fn is_data_current(
+        &self,
+        reward_period: &Range<DateTime<Utc>>,
+    ) -> anyhow::Result<bool> {
+        // Check if we have heartbeats and speedtests past the end of the reward period
+        if reward_period.end >= self.disable_complete_data_checks_until().await? {
+            if sqlx::query_scalar::<_, i64>(
+                "SELECT COUNT(*) FROM heartbeats WHERE latest_timestamp >= $1",
```
jeffgrunewald, line 120: are we guaranteed that when these checks are being run there are no records being written to the database for the _next_ reward period? and if we haven't exhausted the results for the reward period when the rewarder first runs and it goes to sleep we won't pick up any records for the following period during that time?

mobile_verifier/migrations/12_disable_complete_data_checks_until.sql
```diff
@@ -0,0 +1,7 @@
+INSERT INTO
+   meta
+VALUES
+   ('disable_complete_data_checks_until', '0')
```
jeffgrunewald, line 4: i didn't see this getting updated; is this only intended to be manually modified by a human operator?

mobile_verifier/src/heartbeats.rs
```diff
@@ -52,150 +32,182 @@ pub struct HeartbeatReward {
...
-    where
-        T: IntoIterator<Item = Heartbeat>,
-    {
-        for heartbeat in iter.into_iter() {
-            if heartbeat.validity != proto::HeartbeatValidity::Valid {
-                continue;
+    async fn process_file(
+        &self,
+        file: FileInfoStream<CellHeartbeatIngestReport>,
+        cache: &Arc<Cache<(String, DateTime<Utc>), ()>>,
```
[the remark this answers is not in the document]
maplant: Brian made this change and I didn't notice :-)

mobile_verifier/src/heartbeats.rs
```diff
@@ -52,150 +32,184 @@ pub struct HeartbeatReward {
...
-                reward_weight: value.reward_weight,
-            })
+    pub async fn run(mut self, shutdown: triggered::Listener) -> anyhow::Result<()> {
+        tokio::spawn(async move {
+            let cache = Arc::new(Cache::<(String, DateTime<Utc>), ()>::new());
+
+            let cache_clone = cache.clone();
+            tokio::spawn(async move {
+                cache_clone
+                    .monitor(4, 0.25, std::time::Duration::from_secs(60 * 60 * 3))
```
andymck, line 74: this sets the eviction check to 3 hours but the items in the cache are set to a ttl of 2 hours.  I would suggest bringing the eviction check down to something like 30 mins

andymck (approved):

mobile_verifier/src/rewarder.rs
maplant: yeah, I don't really think this is a warning. Although maybe it is? I don't expect this to happen too often, if at all

mobile_verifier/src/heartbeats.rs
```diff
@@ -52,150 +32,184 @@ pub struct HeartbeatReward {
...
-    {
-        for heartbeat in iter.into_iter() {
-            if heartbeat.validity != proto::HeartbeatValidity::Valid {
-                continue;
+    async fn process_file(
+        &self,
+        file: FileInfoStream<CellHeartbeatIngestReport>,
+        cache: &Arc<Cache<(String, DateTime<Utc>), ()>>,
+    ) -> anyhow::Result<()> {
+        tracing::info!("Processing heartbeat file {}", file.file_info.key);
```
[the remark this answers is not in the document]
bbalser: We have metrics for every write to a file_sink, so both heartbeats and speedtests have metrics already.

bbalser (approved):

PR #510, opened 2023-05-08
add receiver_nsid field to http protocol

michaeldjeffrey:
requires https://github.com/helium/proto/pull/327


jeffgrunewald (commented):
looks good in theory once the proto repo is updated so it will build

jeffgrunewald (approved):

andymck (approved):

PR #514, opened 2023-05-11
Implement Org update API and expand Helium NetID management

jeffgrunewald:
Implements the expanded management of Helium org NetIDs:
- Allow selecting one of three (Type 0, 3, 6) Helium NetIDs from which to draw an org's devaddrs (constraints), defaulting to the largest (Type 0) address
- Allow tracking the issuance of available devaddrs from each of the Helium NetID address pools when they are assigned to org's via the org's devaddr constraint ranges and returning addresses to the pool when the org removes the associated constraints
- Allow updating org's: owner_pubkey, payer_pubkey, delegate routing pubkeys, and adding or removing devaddr constraints (various operations authorized or not according to requestor role)

### DIFF
--- a/iot_config/src/gateway_service.rs
+++ b/iot_config/src/gateway_service.rs
@@ -93,13 +93,19 @@ impl GatewayService {
         match self.gateway_cache.get(pubkey).await {
             Some(gateway) => Ok(gateway.value().clone()),
             None => {
-                let metadata = gateway_info::db::get_info(&self.metadata_pool, pubkey)
-                    .await
-                    .map_err(|_| Status::internal("error fetching gateway info"))?
-                    .ok_or_else(|| {
-                        telemetry::count_gateway_info_lookup("not-found");
-                        Status::not_found(format!("gateway not found: pubkey = {pubkey:}"))
-                    })?;
+                let metadata = tokio::select! {
+                    query_result = gateway_info::db::get_info(&self.metadata_pool, pubkey) => {
+                        query_result.map_err(|_| Status::internal("error fetching gateway info"))?
+                            .ok_or_else(|| {
+                                telemetry::count_gateway_info_lookup("not-found");
+                                Status::not_found(format!("gateway not found: pubkey = {pubkey:}"))
+                            })?
+                    }
+                    _ = tokio::time::sleep(tokio::time::Duration::from_secs(3)) => {
+                        tracing::warn!("gateway info request query timed out");
+                        return Err(Status::deadline_exceeded("query timed out requesting info"))
+                    }
+                };
                 let gateway = GatewayInfo::chain_metadata_to_info(metadata, &self.region_map);
                 self.gateway_cache
                     .insert(pubkey.clone(), gateway.clone(), CACHE_TTL)
--- a/iot_config/src/helium_netids.rs
+++ b/iot_config/src/helium_netids.rs
@@ -101,13 +101,16 @@ impl AddressStore for sqlx::Transaction<'_, sqlx::Postgres> {
         net_id: HeliumNetId,
         released_addrs: &[u32],
     ) -> Result<(), Self::Error> {
+        let net_id = i32::from(net_id.id());
+        let released_addrs = released_addrs
+            .iter()
+            .map(|addr| (*addr, net_id))
+            .collect::<Vec<(u32, i32)>>();
         let mut query_builder: sqlx::QueryBuilder<sqlx::Postgres> = sqlx::QueryBuilder::new(
             " delete from helium_used_devaddrs where (devaddr, net_id) in ",
         );
-        query_builder.push_values(released_addrs, |mut builder, addr| {
-            builder
-                .push_bind(*addr as i32)
-                .push_bind(i32::from(net_id.id()));
+        query_builder.push_tuples(released_addrs, |mut builder, (addr, id)| {
+            builder.push_bind(addr as i32).push_bind(id);
         });
         Ok(query_builder.build().execute(self).await.map(|_| ())?)
     }
--- a/iot_config/src/helium_netids.rs
+++ b/iot_config/src/helium_netids.rs
@@ -113,6 +113,10 @@ impl AddressStore for sqlx::Transaction<'_, sqlx::Postgres> {
     }
 }
 
+pub fn is_helium_netid(net_id: &NetIdField) -> bool {
+    [TYPE_0_ID, TYPE_3_ID, TYPE_6_ID].contains(net_id)
+}
+
 pub async fn checkout_devaddr_constraints<S>(
     addr_store: &mut S,
     count: u64,
@@ -126,23 +130,23 @@ where
         .get_used_addrs(net_id)
         .await
         .map_err(DevAddrConstraintsError::AddressStore)?;
+
     let range_start = *addr_range.start();
     let range_end = *addr_range.end();
-    let last_used = if let Some(last) = used_addrs.last() {
-        *last
-    } else {
-        range_start
-    };
+    let last_used = used_addrs.last().copied().unwrap_or(range_start);
     let used_range = (range_start..=last_used).collect::<HashSet<u32>>();
     let used_addrs = used_addrs.into_iter().collect::<HashSet<u32>>();
+
     let mut available_diff = used_range
         .difference(&used_addrs)
         .copied()
         .collect::<Vec<_>>();
     available_diff.sort();
+
     let mut claimed_addrs = available_diff
         .drain(0..(count as usize).min(available_diff.len()))
         .collect::<Vec<_>>();
+
     let mut next_addr = last_used + 1;
     while claimed_addrs.len() < count as usize {
         if next_addr <= range_end {
@@ -152,10 +156,12 @@ where
             return Err(DevAddrConstraintsError::NoAvailableAddrs);
         }
     }
+
     addr_store
         .claim_addrs(net_id, &claimed_addrs)
         .await
         .map_err(DevAddrConstraintsError::AddressStore)?;
+
     let new_constraints = constraints_from_addrs(claimed_addrs)?;
     Ok(new_constraints)
 }
--- a/iot_config/src/helium_netids.rs
+++ b/iot_config/src/helium_netids.rs
@@ -1,0 +1,205 @@
+#![allow(dead_code, unused)]
+use crate::lora_field::{DevAddrConstraint, LoraField, NetIdField, self};
+use helium_proto::services::iot_config::org_create_helium_req_v1::HeliumNetId as ProtoNetId;
+use std::{collections::HashSet, ops::RangeInclusive};
+
+const TYPE_0_ID: NetIdField = LoraField(0x00003c);
+const TYPE_3_ID: NetIdField = LoraField(0x60002d);
+const TYPE_6_ID: NetIdField = LoraField(0xc00053);
+const TYPE_0_RANGE: RangeInclusive<u32> = 2_013_265_920..=2_046_820_351;
+const TYPE_3_RANGE: RangeInclusive<u32> = 3_763_994_624..=3_764_125_695;
+const TYPE_6_RANGE: RangeInclusive<u32> = 4_227_943_424..=4_227_944_447;
+
+pub enum HeliumNetId {
+    Type0Id,
+    Type3Id,
+    Type6Id,
+}
+
+impl HeliumNetId {
+    pub fn id(&self) -> NetIdField {
+        match *self {
+            HeliumNetId::Type0Id => TYPE_0_ID,
+            HeliumNetId::Type3Id => TYPE_3_ID,
+            HeliumNetId::Type6Id => TYPE_6_ID,
+        }
+    }
+
+    pub fn addr_range(&self) -> RangeInclusive<u32> {
+        match *self {
+            HeliumNetId::Type0Id => TYPE_0_RANGE,
+            HeliumNetId::Type3Id => TYPE_3_RANGE,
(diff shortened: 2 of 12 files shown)


iot_config/src/helium_netids.rs
```diff
@@ -0,0 +1,291 @@
...
+        range_start
+    };
+    let used_range = (range_start..=last_used).collect::<HashSet<u32>>();
+    let used_addrs = used_addrs.into_iter().collect::<HashSet<u32>>();
+    let mut diff = used_range
+        .difference(&used_addrs)
+        .copied()
+        .collect::<Vec<_>>();
+    let mut next_addr = last_used + 1;
+    while diff.len() < count as usize {
```
michaeldjeffrey, line 109: If there is an unallocated gap larger than the requested amount of devaddrs, the whole gap will be allocated.

```rust
    #[tokio::test]
    async fn allocate_fewer_than_existing_gap() {
        let used_addrs = Mutex::new(vec![]);
        checkout_devaddr_constraints(&used_addrs, 8, HeliumNetId::Type0)
            .await
            .expect("allocate first round");
        checkout_devaddr_constraints(&used_addrs, 32, HeliumNetId::Type0)
            .await
            .expect("allocate second round");
        checkout_devaddr_constraints(&used_addrs, 8, HeliumNetId::Type0)
            .await
            .expect("allocate third round");

        // round 2 goes out of business, and their devaddrs are released back to the wild
        let remove: Vec<u32> = used_addrs
            .lock()
            .await
            .clone()
            .into_iter()
            .skip(8)
            .take(32)
            .collect();
        assert_eq!(Ok(()), used_addrs.release_addrs(&remove).await);
        assert_eq!(8 + 8, used_addrs.lock().await.len());
        checkout_devaddr_constraints(&used_addrs, 8, HeliumNetId::Type0)
            .await
            .expect("allocate fourth round");
        assert_eq!(8 + 8 + 8, used_addrs.lock().await.len());
    }
```

iot_config/src/helium_netids.rs
```diff
@@ -0,0 +1,291 @@
+#![allow(dead_code, unused)]
```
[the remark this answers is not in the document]
jeffgrunewald: yes, removed

iot_config/src/helium_netids.rs
jeffgrunewald: good catch! fixed [c6cca67](https://github.com/helium/oracles/pull/514/commits/c6cca67e1acb8c529c73920ea6e11d9ba437c6ee)

iot_config/src/helium_netids.rs
```diff
@@ -0,0 +1,377 @@
...
+        .await
+        .map_err(DevAddrConstraintsError::AddressStore)?;
+    let range_start = *addr_range.start();
+    let range_end = *addr_range.end();
+    let last_used = if let Some(last) = used_addrs.last() {
+        *last
+    } else {
+        range_start
+    };
+    let used_range = (range_start..=last_used).collect::<HashSet<u32>>();
```
michaeldjeffrey, line 147: `used_addrs` is not Net ID aware. 
If you attempt to checkout devaddrs in a Type0 NetID after an existing Type6, this function will never finish.

```rust
    #[tokio::test]
    async fn allocate_across_net_id() {
        let mut used_addrs = vec![];
        checkout_devaddr_constraints(&mut used_addrs, 8, HeliumNetId::Type6)
            .await
            .expect("testing allocation");
        checkout_devaddr_constraints(&mut used_addrs, 8, HeliumNetId::Type3)
            .await
            .expect("special request allocation");
        checkout_devaddr_constraints(&mut used_addrs, 8, HeliumNetId::Type0)
            .await
            .expect("average allocation");

        assert_eq!(8 + 8 + 8, used_addrs.len());
    }
```

iot_config/src/org.rs
```diff
@@ -148,6 +333,7 @@ async fn insert_constraints(
         r#"
         insert into organization_devaddr_constraints (oui, net_id, start_addr, end_addr)
         values ($1, $2, $3, $4)
+        on conflict net_id do nothing
```
michaeldjeffrey, line 336: I don't see any constraints on this table that could cause a conflict.

iot_config/src/org_service.rs
```diff
@@ -168,14 +198,28 @@ impl iot_config::Org for OrgService {
 
         tracing::debug!("create helium org request: {request:?}");
 
-        let requested_addrs = request.devaddrs;
-        let devaddr_constraint = org::next_helium_devaddr(&self.pool)
+        let net_id = request.net_id();
+        let requested_addrs = if request.devaddrs >= 8 {
```
michaeldjeffrey, line 202: I'm not sure what the correct answer is here, but if there's a way to consolidate more checks here...
If they pass less than 8, we tell them "insufficient", which I first read as "not enough available".

If they pass 9, this check is good, but you can't checkout an uneven number of devaddrs. And that error message get's returned as "helium addresses unavailable".

iot_config/src/org.rs
```diff
@@ -138,7 +151,179 @@ pub async fn create_org(
...
+                    &mut txn,
+                    addr_count,
+                    helium_net_id,
+                )
+                .await
+                .map_err(|err| OrgStoreError::SaveConstraints(format!("{err}")))?;
+                insert_helium_constraints(oui, net_id, &constraints, &mut txn).await?
+            }
+            Some(proto::Update::Constraint(constraint_update))
+                if authorizer == UpdateAuthorizer::Admin && is_helium_org =>
```
michaeldjeffrey, line 192: This check prevents roaming orgs from allowing others access to their devaddr ranges.

iot_config/src/org.rs
```diff
@@ -276,6 +462,8 @@ pub enum OrgStoreError {
     DecodeKey(#[from] helium_crypto::Error),
     #[error("Route Id parse error: {0}")]
     RouteIdParse(#[from] sqlx::types::uuid::Error),
+    #[error("Invalid or Unauthorized update: {0}")]
+    InvalidUpdate(String),
```
michaeldjeffrey, line 466: There's enough ways to fail an update it might make sense to differentiate invalid vs unauthorized.
Especially since unauthorized checks fall through to a generic message.

iot_config/src/helium_netids.rs
```diff
@@ -0,0 +1,377 @@
...
+pub trait AddressStore {
+    type Error;
+
+    async fn get_used_addrs(&mut self) -> Result<Vec<u32>, Self::Error>;
+    async fn claim_addrs(&mut self, new_addrs: &[u32]) -> Result<(), Self::Error>;
+    async fn release_addrs(&mut self, released_addrs: &[u32]) -> Result<(), Self::Error>;
+}
+
+#[async_trait::async_trait]
+impl AddressStore for sqlx::Pool<sqlx::Postgres> {
```
michaeldjeffrey, line 60: Nothing appears to be using the Pool impl.

iot_config/src/helium_netids.rs
```diff
@@ -0,0 +1,377 @@
+use crate::lora_field::{self, DevAddrConstraint, LoraField, NetIdField};
+use helium_proto::services::iot_config::org_create_helium_req_v1::HeliumNetId as ProtoNetId;
+use std::{collections::HashSet, ops::RangeInclusive};
+
+const TYPE_0_ID: NetIdField = LoraField(0x00003c);
```
michaeldjeffrey, line 5: I'm not sure how to properly identify it, but we still have to support `000024` as a NetID as well. At least until everyone can be migrated away to `00003c`.

jeffgrunewald: i assumed that if these are the three official helium netids then these would be the ones we would need to make extensible and manage ongoing; i don't believe there are any changes in this PR that _don't_ support `000024` in the same fashion they've already been supported: orgs can continue to assign devaddrs within slices of that address space, they just can't extend within that space. once everyone is migrated they're just database entries we can keep or delete as needed

iot_config/src/helium_netids.rs
jeffgrunewald: indeed; i could have sworn i deleted that

iot_config/src/org.rs
jeffgrunewald: beyond the initial creation rpc that allocates the devaddr contraints/netid we don't currently distinguish between roamers and helium orgs and the current requirement was that orgs could modify their own delegate keys and payer key while a (foundation) admin must authorize a change of the owner pubkey and adding/removing devaddr constraints

iot_config/src/org_service.rs
jeffgrunewald: sure thing; fixed

iot_config/src/org.rs
jeffgrunewald: dang; good catch

iot_config/src/org.rs
jeffgrunewald: additionally, i don't disagree with letting roaming orgs manage the devaddr constraints within their owned non-helium netids or even helium orgs subletting portions of their own leased constraint ranges but since this wasn't an immediately necessary feature i'd rather not support it until we have time to do a pretty substantial reworking of how to do authorization and permissioning of access to rpcs and records; i can't really conceive of a way to allow those kinds of distinctions without reworking the entire auth structure and probably a fair portion of the org/route database table schemas

iot_config/src/helium_netids.rs
jeffgrunewald: fixed: [56ba443](https://github.com/helium/oracles/pull/514/commits/56ba4431b2ce7adcc0d2bb52d98b9fe80cfb3936)

iot_config/src/org.rs
michaeldjeffrey: I was only trying to point out the `&& is_helium_org` prevents any non-helium org from being updated by this rpc. Whether that be a roaming org subletting part of their address space, or a typo on creation. 

But it's not something needed right now. Just mentioning it in hopes that someone will remember when reworking is done. :+1:

michaeldjeffrey (approved):

iot_config/src/gateway_service.rs
```diff
@@ -93,13 +93,19 @@ impl GatewayService {
...
             Some(gateway) => Ok(gateway.value().clone()),
             None => {
-                let metadata = gateway_info::db::get_info(&self.metadata_pool, pubkey)
-                    .await
-                    .map_err(|_| Status::internal("error fetching gateway info"))?
-                    .ok_or_else(|| {
-                        telemetry::count_gateway_info_lookup("not-found");
-                        Status::not_found(format!("gateway not found: pubkey = {pubkey:}"))
-                    })?;
+                let metadata = tokio::select! {
```
bbalser, line 96: nice!

iot_config/src/org.rs
```diff
@@ -122,14 +126,29 @@ pub async fn create_org(
...
-            OrgStoreError::SaveConstraints(format!(
-                "{} - {}",
-                devaddr_range.start_addr, devaddr_range.end_addr
-            ))
-        })?;
+    if [
+        HeliumNetId::Type0_0x00003c.id(),
+        HeliumNetId::Type3_0x60002d.id(),
+        HeliumNetId::Type6_0xc00053.id(),
+    ]
```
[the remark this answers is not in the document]
jeffgrunewald: i can; that `id()` method just does a match and returns a const so i'd need to either make this a static ref or else make those consts public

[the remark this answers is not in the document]
jeffgrunewald: made a `is_helium_netid()` function for this

iot_config/src/org.rs
```diff
@@ -138,7 +157,192 @@ pub async fn create_org(
...
+                    .try_into()
+                    .map_err(|err: &'static str| OrgStoreError::InvalidUpdate(err.to_string()))?;
+                let constraints = helium_netids::checkout_devaddr_constraints(
+                    &mut txn,
+                    addr_count,
+                    helium_net_id,
+                )
+                .await
+                .map_err(|err| OrgStoreError::SaveConstraints(format!("{err}")))?;
+                insert_helium_constraints(oui, net_id, &constraints, &mut txn).await?
```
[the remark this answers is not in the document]
jeffgrunewald: this has been broken further into smaller component functions

iot_config/src/org.rs
```diff
@@ -122,14 +124,23 @@ pub async fn create_org(
...
+            ))?;
+        if check_roamer_constraint_count(net_id, &mut txn).await? == 0 {
+            insert_roamer_constraint(oui as u64, net_id, constraint, &mut txn).await
+        } else {
+            return Err(OrgStoreError::SaveConstraints(format!(
+                "constraint already in use {constraint:?}"
+            )));
+        }
+    }
+    .map_err(|_| OrgStoreError::SaveConstraints(format!("{devaddr_ranges:?}")))?;
```
andymck, line 143: you sure no important error details are being masked with this type of remapping  ( here and in other similar instances )?

iot_config/src/org.rs
```diff
@@ -138,7 +149,255 @@ pub async fn create_org(
...
+
+    Ok(updated_org)
+}
+
+pub async fn get_org_netid(
+    oui: u64,
+    db: impl sqlx::PgExecutor<'_>,
+) -> Result<NetIdField, sqlx::Error> {
+    let netid = sqlx::query_scalar::<_, i32>(
+        " select net_id from organization_devaddr_constraints where oui = $1 limit 1 ",
```
andymck, line 218: using limit without order by here, what about side effects from potential inconsistent results ?   Or is net_id guaranteed to be a consistent value here ?

andymck (approved):

PR #516, opened 2023-05-12
Gateway cache related tweaks

andymck:
Various improvements identified after an outage between iot verifier and the config service.

1. fire cache miss metric irrespective of result of config service request
2. Add metric to track config service request errors
3. add jitter to the TTL for each entry to the cache

### DIFF
--- a/iot_verifier/src/gateway_cache.rs
+++ b/iot_verifier/src/gateway_cache.rs
@@ -73,7 +73,7 @@ impl GatewayCache {
                 Ok(hit.value().clone())
             }
             _ => {
-                tracing::debug!("cache miss: {:?}", address);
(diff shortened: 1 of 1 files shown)


iot_verifier/src/gateway_cache.rs
```diff
@@ -65,28 +73,35 @@ impl GatewayCache {
                 Ok(hit.value().clone())
             }
             _ => {
+                tracing::debug!("cache miss: {:?}", address);
```
jeffgrunewald, line 76:
```suggestion
                tracing::debug!(?address, "cache miss");
```
structured log format with the short-hand "assigned log key is the same as the variable name being logged"

PR #518, opened 2023-05-19
Change file_sink and file_info_poller to handle shutdowns better while send data across mpsc channel

bbalser:
During testing it was discovered 2 issues in the file_store crate that could cause shutdowns to hang.

in file_sink, we learned that calling `sender.send` and the channel is at capacity, it will not error when the receiver is dropped.  Address this by using `sender.send_timeout` and including a shutdown_listener in a `tokio::select`

In file_info_poller, changed to using `sender.try_send` and handling the Full Error by breaking the for loop and waiting for the next interval tick.gg

### DIFF
--- a/file_store/src/file_info_poller.rs
+++ b/file_store/src/file_info_poller.rs
@@ -4,7 +4,7 @@ use derive_builder::Builder;
 use futures::{stream::BoxStream, StreamExt};
 use retainer::Cache;
 use std::marker::PhantomData;
-use tokio::sync::mpsc::{Receiver, Sender};
+use tokio::sync::mpsc::{error::TrySendError, Receiver, Sender};
 
 const DEFAULT_POLL_DURATION_SECS: i64 = 30;
 const DEFAULT_POLL_DURATION: std::time::Duration =
@@ -87,10 +87,9 @@ where
         loop {
             let after = self.after(latest_ts);
             let before = Utc::now();
-            let shutdown = shutdown.clone();
 
             tokio::select! {
-                _ = shutdown => {
+                _ = shutdown.clone() => {
                     tracing::info!("FileInfoPoller shutting down");
                     break;
                 }
@@ -99,9 +98,13 @@ where
                     let files = self.store.list_all(self.file_type, after, before).await?;
                     for file in files {
                         if !is_already_processed(&self.db, &cache, &file).await? {
-                            latest_ts = Some(file.timestamp);
-                            cache_file(&cache, &file).await;
-                            send_stream(&sender, &self.store, file).await?;
+                            if send_stream(&sender, &self.store, file.clone()).await? {
+                                latest_ts = Some(file.timestamp);
+                                cache_file(&cache, &file).await;
+                            } else {
+                                tracing::info!("FileInfoPoller: channel full");
+                                break;
+                            }
                         }
                     }
                 }
@@ -136,7 +139,7 @@ async fn send_stream<T>(
     sender: &Sender<FileInfoStream<T>>,
     store: &FileStore,
     file: FileInfo,
-) -> Result
+) -> Result<bool>
 where
     T: MsgDecode + TryFrom<T::Msg, Error = Error> + Send + Sync + 'static,
 {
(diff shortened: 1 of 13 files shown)


file_store/src/file_sink.rs
```diff
@@ -490,16 +514,20 @@ mod tests {
...
-        let (file_sink_client, mut file_sink_server) =
-            FileSinkBuilder::new(FileType::EntropyReport, tmp_dir.path(), "fake_metric")
-                .roll_time(chrono::Duration::milliseconds(100))
-                .create()
-                .await
-                .expect("failed to create file sink");
+        let (file_sink_client, mut file_sink_server) = FileSinkBuilder::new(
+            FileType::EntropyReport,
+            tmp_dir.path(),
+            "fake_metric",
```
andymck, line 520: whats with these 'fake_metric' refs?

file_store/src/file_sink.rs
```diff
@@ -145,10 +157,12 @@ impl FileSinkBuilder {
 pub struct FileSinkClient {
     sender: MessageSender,
     metric: &'static str,
+    shutdown_listener: triggered::Listener,
 }
 
 const OK_LABEL: Label = Label::from_static_parts("status", "ok");
 const ERROR_LABEL: Label = Label::from_static_parts("status", "error");
+const SEND_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(10);
```
andymck, line 165: 10 secs seems a very long time to wait for an internal msg send

andymck (approved):

file_store/src/file_sink.rs
bbalser: That is in a test, so the metric name is irrelevant.

file_store/src/file_sink.rs
bbalser: yeah, I just picked something to get a conversation going, what do you think it should be?

file_store/src/file_info_poller.rs
```diff
@@ -171,8 +174,11 @@ where
         stream,
     };
 
-    let _ = sender.send(incoming_data_stream).await;
-    Ok(())
+    match sender.try_send(incoming_data_stream) {
+        Ok(_) => Ok(true),
+        Err(TrySendError::Full(_)) => Ok(false),
```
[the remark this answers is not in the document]
bbalser: It is not treating the channel being at capacity as an error. Returning false causes it to stop trying to send any more file across the channel and just wait for the next interval tick to try again.  I could return an Error here and than catch it in the calling function, but it felt simpler to just return a boolean rather than add another variant to the file_store/error.rs just to catch it in the caller and not bubble it up.

file_store/src/file_sink.rs
andymck: 5 secs ? good old erlang default sync timeout

jeffgrunewald, line 165: We still want this to try and send for 10 seconds? Not drop this down to 3-4?

file_store/src/file_sink.rs
```diff
@@ -160,27 +174,36 @@ impl FileSinkClient {
...
-                    self.metric,
-                    labels
-                        .chain(std::iter::once(ERROR_LABEL))
-                        .collect::<Vec<Label>>()
-                );
-                tracing::error!("file_sink write failed for {:?} with {e:?}", self.metric);
-                Err(Error::channel())
+        tokio::select! {
+            _ = self.shutdown_listener.clone() => {
+                Err(Error::Shutdown)
```
jeffgrunewald, line 179: What’s the logic behind labeling a shutdown as an error? Just in the context of a send that won’t complete because of the shutdown ?

file_store/src/file_sink.rs
bbalser: Fixed bff5af5

file_store/src/file_sink.rs
jeffgrunewald: best i got is `Err(Error::ItIsWhatItIs)` 🤷

jeffgrunewald (approved):

PR #519, opened 2023-05-19
Add gateway info producer and consumer data in gateway cache and density scaler

andymck:
Replaces the gateway info retainer cache and the density scaler's separate streaming of gateway info from config service with a single tokio watch producer.  The producer will periodically ( default 30 mins ) stream all gateways from config service and the resulting gateway map will be consumed by the gateway info cache and the density scalers.

This way the gateway cache and the density scaler are always in sync.  It also eliminates any individual requests to the config service for gateway info.

### DIFF
--- a/iot_verifier/src/gateway_cache.rs
+++ b/iot_verifier/src/gateway_cache.rs
@@ -1,93 +1,37 @@
-use futures::stream::StreamExt;
+use crate::gateway_updater::MessageReceiver;
 use helium_crypto::PublicKeyBinary;
-use iot_config::{
-    client::{Client as IotConfigClient, ClientError as IotConfigClientError},
-    gateway_info::{GatewayInfo, GatewayInfoResolver},
-};
-use retainer::Cache;
-use std::{sync::Arc, time::Duration};
-use tokio::task::JoinHandle;
-
-/// how long each cached items takes to expire ( 12 hours in seconds)
-const CACHE_TTL: Duration = Duration::from_secs(12 * (60 * 60));
-/// how often to evict expired items from the cache ( every 1 hour)
-const CACHE_EVICTION_FREQUENCY: Duration = Duration::from_secs(60 * 60);
+use iot_config::gateway_info::GatewayInfo;
 
 pub struct GatewayCache {
-    pub iot_config_client: IotConfigClient,
-    pub cache: Arc<Cache<PublicKeyBinary, GatewayInfo>>,
-    pub cache_monitor: JoinHandle<()>,
+    gateway_cache_receiver: MessageReceiver,
 }
 
 #[derive(Debug, thiserror::Error)]
 pub enum GatewayCacheError {
     #[error("gateway not found: {0}")]
     GatewayNotFound(PublicKeyBinary),
-    #[error("error querying iot config service")]
-    IotConfigClient(#[from] IotConfigClientError),
 }
 
 impl GatewayCache {
-    pub fn from_settings(iot_config_client: IotConfigClient) -> Self {
-        let cache = Arc::new(Cache::<PublicKeyBinary, GatewayInfo>::new());
-        let clone = cache.clone();
-        // monitor cache to handle evictions
-        let cache_monitor =
-            tokio::spawn(async move { clone.monitor(4, 0.25, CACHE_EVICTION_FREQUENCY).await });
+    pub fn new(gateway_cache_receiver: MessageReceiver) -> Self {
         Self {
-            iot_config_client,
-            cache,
-            cache_monitor,
+            gateway_cache_receiver,
         }
-    }
-
-    pub async fn prewarm(&self) -> anyhow::Result<()> {
-        tracing::info!("starting prewarming gateway cache");
-        let mut gw_stream = self
-            .iot_config_client
-            .clone()
-            .stream_gateways_info()
-            .await?;
-        while let Some(gateway_info) = gw_stream.next().await {
-            _ = self.insert(gateway_info).await;
-        }
-        tracing::info!("completed prewarming gateway cache");
-        Ok(())
     }
 
     pub async fn resolve_gateway_info(
         &self,
         address: &PublicKeyBinary,
     ) -> Result<GatewayInfo, GatewayCacheError> {
-        match self.cache.get(address).await {
+        match self.gateway_cache_receiver.borrow().get(address) {
             Some(hit) => {
-                metrics::increment_counter!("oracles_iot_verifier_gateway_cache_hit");
-                Ok(hit.value().clone())
+                metrics::increment_counter!("oracles_iot_verifier_gateway_found");
+                Ok(hit.clone())
             }
             _ => {
-                match self
-                    .iot_config_client
-                    .clone()
-                    .resolve_gateway_info(address)
-                    .await
-                {
-                    Ok(Some(res)) => {
-                        tracing::debug!("cache miss: {:?}", address);
-                        metrics::increment_counter!("oracles_iot_verifier_gateway_cache_miss");
-                        _ = self.insert(res.clone()).await;
-                        Ok(res)
-                    }
-                    Ok(None) => Err(GatewayCacheError::GatewayNotFound(address.clone())),
-                    Err(err) => Err(GatewayCacheError::IotConfigClient(err)),
-                }
+                metrics::increment_counter!("oracles_iot_verifier_gateway_not_found");
+                Err(GatewayCacheError::GatewayNotFound(address.clone()))
             }
         }
-    }
-
-    pub async fn insert(&self, gateway: GatewayInfo) -> anyhow::Result<()> {
-        self.cache
-            .insert(gateway.address.clone(), gateway, CACHE_TTL)
-            .await;
-        Ok(())
     }
 }
--- a/iot_verifier/src/gateway_cache.rs
+++ b/iot_verifier/src/gateway_cache.rs
@@ -25,11 +25,11 @@ impl GatewayCache {
     ) -> Result<GatewayInfo, GatewayCacheError> {
         match self.gateway_cache_receiver.borrow().get(address) {
             Some(hit) => {
-                metrics::increment_counter!("oracles_iot_verifier_gateway_found");
+                metrics::increment_counter!("oracles_iot_verifier_gateway_cache_hit");
                 Ok(hit.clone())
             }
-            _ => {
-                metrics::increment_counter!("oracles_iot_verifier_gateway_not_found");
+            None => {
+                metrics::increment_counter!("oracles_iot_verifier_gateway_cache_miss");
                 Err(GatewayCacheError::GatewayNotFound(address.clone()))
             }
         }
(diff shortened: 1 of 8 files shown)


iot_verifier/src/gateway_updater.rs
```diff
@@ -0,0 +1,95 @@
...
+
+    async fn handle_refresh_tick(&mut self) -> Result<(), GatewayStoreError> {
+        let updated_gateway_map = refresh_gateways(self.iot_config_client.clone()).await?;
+        _ = self.sender.send(updated_gateway_map);
+        Ok(())
+    }
+}
+
+pub async fn refresh_gateways(
+    iot_config_client: IotConfigClient,
```
bbalser, line 81: Change this to and you should be able to remove all the cloning of iot_config_client
```rust
iot_config_client: &mut IotConfigClient,
```

iot_verifier/src/gateway_updater.rs
```diff
@@ -0,0 +1,95 @@
...
+) -> Result<GatewayMap, GatewayStoreError> {
+    tracing::info!("refreshing gateway_store");
+    let mut map = HashMap::<PublicKeyBinary, GatewayInfo>::new();
+    let mut gw_stream = iot_config_client.clone().stream_gateways_info().await?;
+    while let Some(gateway_info) = gw_stream.next().await {
+        _ = map.insert(gateway_info.address.clone(), gateway_info)
```
bbalser, line 87:
```rust
        map.insert(gateway_info.address.clone(), gateway_info);
```

iot_verifier/src/gateway_cache.rs
```diff
@@ -1,93 +1,37 @@
...
+        Self { cache_receiver }
     }
 
     pub async fn resolve_gateway_info(
         &self,
         address: &PublicKeyBinary,
     ) -> Result<GatewayInfo, GatewayCacheError> {
-        match self.cache.get(address).await {
+        // TODO: is borrowing sufficient here or should I clone the
+        // gateway map into the GatewayCache struct ?
```
bbalser, line 25: I believe you want the borrow here, since you only need one entry at a time.

iot_verifier/src/gateway_updater.rs
```diff
@@ -0,0 +1,95 @@
...
+pub struct GatewayStore {
+    iot_config_client: IotConfigClient,
+    refresh_interval: Duration,
+    pub receiver: MessageReceiver,
+    sender: MessageSender,
+}
+
+pub struct GatewayMap {
+    pub gateways: HashMap<PublicKeyBinary, GatewayInfo>,
+    pub timestamp: i64,
```
bbalser, line 25: Where is this used?

andymck: its not, removed

iot_verifier/src/gateway_cache.rs
```diff
@@ -1,93 +1,37 @@
...
 
 pub struct GatewayCache {
-    pub iot_config_client: IotConfigClient,
-    pub cache: Arc<Cache<PublicKeyBinary, GatewayInfo>>,
-    pub cache_monitor: JoinHandle<()>,
+    gateway_cache_receiver: MessageReceiver,
 }
 
 #[derive(Debug, thiserror::Error)]
 pub enum GatewayCacheError {
```
jeffgrunewald, line 10: do we think there are any other valid errors that can come from this cache? if not this could just be an error struct instead of an enum and i believe you can even define it in-line with the gateway cache struct so the result coming from `resolve_gateway_info` is just `Result<GatewayInfo, Self::Error>`

iot_verifier/src/gateway_cache.rs
```diff
@@ -1,93 +1,37 @@
...
     pub async fn resolve_gateway_info(
         &self,
         address: &PublicKeyBinary,
     ) -> Result<GatewayInfo, GatewayCacheError> {
-        match self.cache.get(address).await {
+        match self.gateway_cache_receiver.borrow().get(address) {
             Some(hit) => {
-                metrics::increment_counter!("oracles_iot_verifier_gateway_cache_hit");
-                Ok(hit.value().clone())
+                metrics::increment_counter!("oracles_iot_verifier_gateway_found");
```
jeffgrunewald, line 28: i know we're no longer falling back to checking directly with the config service when this call fails to return a result but i think it would still be helpful to label these metrics to highlight that we're hitting a cache, not the direct source of data

iot_verifier/src/main.rs
```diff
@@ -87,8 +87,10 @@ impl Server {
 
         let iot_config_client = IotConfigClient::from_settings(&settings.iot_config_client)?;
 
-        let gateway_cache = GatewayCache::from_settings(iot_config_client.clone());
-        _ = gateway_cache.prewarm().await;
+        let gateway_updater =
```
jeffgrunewald, line 90: the gateway_updater struct really doesn't need a copy of the `MessageReceiver` since it has no use for it; i think it's more idiomatic to have the `GatewayUpdater::from_settings()` return a tuple `(gateway_update_receiver, gateway_updater) = GatewayUpdater::from_settings(...inputs...)` and then only pass the receiver clones to the things that need them.

iot_verifier/src/tx_scaler.rs
```diff
@@ -89,10 +74,10 @@ impl Server {
...
             .await
             .map_err(sqlx::Error::from)?;
-        let mut gw_stream = self.iot_config_client.stream_gateways_info().await?;
-        while let Some(gateway_info) = gw_stream.next().await {
-            if let Some(metadata) = gateway_info.metadata {
-                if active_gateways.contains_key(&gateway_info.address.as_ref().to_vec()) {
+        for k in active_gateways.keys() {
+            let pubkey = PublicKeyBinary::from(k.clone());
+            if let Some(gateway_info) = self.gateway_cache_receiver.borrow().get(&pubkey) {
+                if let Some(metadata) = &gateway_info.metadata {
```
jeffgrunewald, line 80: the `GatewayMap` in the receiver watch channel is just a hashmap so you can save all of the pubkey cloning and serialization here by just calling `self.gateway_cache_receiver.borrow().values()` and you get an iterator over all of the gateway info entries in the map that you can then match on with `let Some(metadata) = gateway.metadata {...etc...}`. i would imagine with over 400K gateways on average for each run, avoiding all that pubkey deserialization that will significantly improve the performance of this function

andymck: I agree the vec -> pubkey binary of the pubkey here  is sub optimal, but not sure what you suggest is gonna be much better,  the gateway_cache map contains  *all* gateways ( asserted, unasserted etc and a count of ~980k ) whilst the active_gateways maps is obviously only those recently active ( <400k ).

I would need to iterative over the 980k cached gateways in the approach you suggest rather than the smaller active list.

jeffgrunewald: It’d be interesting to benchmark both solutions but since we’re not talking about that much data per gateway (these are pretty tiny records) and they’re already deserialized my guess would be the serde of the public keys is still going to be the bigger perf hit. Since the hashmap’s gateways are already deserialized and you can pattern match the items you need I’d bet the `values` iterator is going to be roughly just as if not faster.

andymck: But serde is not involved here is it ? we are going from a vec<u8> ( derived from a PublicKeyBinary.as_ref())  back to a PublicKeyBinary

jeffgrunewald: Not Serde the crate, but maybe you’re right about it going from a byte array to a pubkeybinary. I still worry about cloning 400k of those though

iot_verifier/src/gateway_cache.rs
andymck: At this point GatewayNotFound is the only error.  The enum itself is referenced in multiple modules and I think for maintainability ( should other errors be appended later ) keeping the enum is a better solution

iot_verifier/src/gateway_cache.rs
andymck: fixed here: 13fc436c2496633a4f744bd7c48299df873fc97d

iot_verifier/src/tx_scaler.rs
andymck: See below for the code used for the revised implementation:

`        for gateway in self.gateway_cache_receiver.borrow().values() {
             if active_gateways.get(&gateway.address.as_ref().to_vec()).is_some() {
                if let Some(metadata) = &gateway.metadata {
                    global_map.increment_unclipped(metadata.location)
                }
            }
        }`

iot_verifier/src/main.rs
bbalser: I actually like `from_settings` returning a tuple, would it be better if it was a free function so you called it via `gateway_update::from_settings` rather than being part of the `GatewayUpdater` impl?

andymck: I dont have a better suggestion so addressed here: 4b05c721c6c7d8f403ebf4ffeb183386973c9d2a

iot_verifier/src/gateway_updater.rs
andymck: addressed here: b8849cf44e3212e8159fd09c1135dc690073e002

bbalser (approved):

iot_verifier/src/gateway_updater.rs
```diff
@@ -0,0 +1,90 @@
...
+                _ = trigger_timer.tick() => self.handle_refresh_tick().await?,
+                _ = shutdown.clone() => return Ok(()),
+            }
+        }
+    }
+
+    async fn handle_refresh_tick(&mut self) -> Result<(), GatewayUpdaterError> {
+        tracing::info!("handling refresh tick");
+        let updated_gateway_map = refresh_gateways(&mut self.iot_config_client).await?;
+        _ = self.sender.send(updated_gateway_map);
```
jeffgrunewald, line 73: would it be better to trigger this send error here with `self.sender.send(updated_gateway_map)?;`? i believe it's another case where that should only error in the event the receiver(s) have all been dropped

iot_verifier/src/gateway_updater.rs
```diff
@@ -0,0 +1,90 @@
...
+pub struct GatewayUpdater {
+    iot_config_client: IotConfigClient,
+    refresh_interval: Duration,
+    sender: MessageSender,
+}
+
+#[derive(Debug, thiserror::Error)]
+pub enum GatewayUpdaterError {
+    #[error("gateway not found: {0}")]
+    GatewayNotFound(PublicKeyBinary),
```
jeffgrunewald, line 26: is this a valid error now since the updater is never looking for any specific gateway? perhaps this should be replaced by the SendError per the comment below?

iot_verifier/src/gateway_updater.rs
andymck: addressed here: fdabbdfa32b12cc10f890d753fa15726fd7effd8

iot_verifier/src/gateway_updater.rs
andymck: addressed here: fdabbdfa32b12cc10f890d753fa15726fd7effd8

jeffgrunewald (approved):

PR #522, opened 2023-05-26
ensure elevation and gain are default when gateway asserted

jeffgrunewald:
ensures asserted gateways found in the helius-fed database that do not have a defined gain or elevation use the default values from the old helium chain. this can happen because the gain and elevation are set by a second, separate transaction from the assert location transaction.

gateways that have not been asserted at all will still have a `None` for their hotspot info metadata field.

this change also, in the event a gateway has an asserted location that doesn't map to a lorawan region recognized by the helium network, will populate the rest of the metadata struct with the supplied data and default the region field to the recently added `Region::Unknown` variant to make it easier to identify when confirmed asserted gateways are not participating in network activities as expected.


iot_config/src/gateway_info.rs
```diff
@@ -178,8 +167,8 @@ pub(crate) mod db {
                 )
                 .map_err(|err| sqlx::Error::Decode(Box::new(err)))?,
                 location: row.get::<Option<i64>, &str>("location").map(|v| v as u64),
-                elevation: row.get::<Option<i32>, &str>("elevation"),
-                gain: row.get::<Option<i32>, &str>("gain"),
+                elevation: row.get::<Option<i32>, &str>("elevation").unwrap_or(0),
```
madninja, line 170: make const's somewhere? I'd like them to be more visible

bbalser (approved):

bbalser (approved):

PR #528, opened 2023-06-05
update oracles to replace h3ron with h3o

jeffgrunewald:
various other projects are switching away from the h3ron c-binding library for h3 index handling to the pure-rust h3o which does more error validation of indexes.

this also necessitated changing of one of the test case index u64 locations which proved to be an invalid cell address as a result of the switch.

### DIFF
--- a/iot_verifier/src/hex_density.rs
+++ b/iot_verifier/src/hex_density.rs
@@ -1,9 +1,10 @@
 use file_store::SCALING_PRECISION;
-use h3ron::{FromH3Index, H3Cell, Index};
+use h3o::{CellIndex, Resolution};
 use itertools::Itertools;
+use lazy_static::lazy_static;
 use rust_decimal::Decimal;
 use rust_decimal_macros::dec;
-use std::{cmp, collections::HashMap, ops::Range, sync::Arc};
+use std::{cmp, collections::HashMap, sync::Arc};
 use tokio::sync::RwLock;
 
 pub struct HexResConfig {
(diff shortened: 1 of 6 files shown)


iot_verifier/src/poc.rs
```diff
@@ -802,7 +808,7 @@ mod tests {
 
     const BEACONER_GAIN: u64 = 20;
     const LOC0: u64 = 631615575095659519; // malta
-    const LOC1: u64 = 631615575095699519; // malta but a lil out from LOC0
+    const LOC1: u64 = 631615576056478207; // malta but a lil out from LOC0
```
maplant, line 811: was this value just wrong before?

maplant (approved):

iot_verifier/src/poc.rs
jeffgrunewald: see the note above; this value was previously an invalid cell

maplant: ah

iot_verifier/src/hex_density.rs
```diff
@@ -22,29 +23,47 @@ impl HexResConfig {
...
-    Some(HexResConfig::new(1, 250, 800)), // 4
-    Some(HexResConfig::new(1, 100, 400)), // 5
-    Some(HexResConfig::new(1, 25, 100)),  // 6
-    Some(HexResConfig::new(2, 5, 20)),    // 7
-    Some(HexResConfig::new(2, 1, 4)),     // 8
-    Some(HexResConfig::new(2, 1, 2)),     // 9
-    Some(HexResConfig::new(2, 1, 1)),     // 10
+const MAX_RES: Resolution = Resolution::Eleven;
+const USED_RES: [Resolution; 7] = [
+    Resolution::Ten,
```
JayKickliter, line 30: huh, they went all out with the strong typing. Kinda annoying for this

jeffgrunewald: thankfully they attached a few nice features to the resolution enum but agreed; this particular use case made things a bit clumsier on our end. open to suggestions for adapting this back to a more "index-able" lookup the way it used to be indexing into a slice

iot_verifier/src/hex_density.rs
```diff
@@ -193,8 +207,8 @@ fn limit(res: u8, occupied_count: u64) -> u64 {
 pub fn compute_hex_density_map(global_map: &GlobalHexMap) -> HashMap<u64, Decimal> {
     let mut map: HashMap<u64, Decimal> = HashMap::new();
     for hex in &global_map.asserted_hexes {
-        let scale: Decimal = SCALING_RES.rev().fold(dec!(1.0), |scale, res| {
-            hex.get_parent(res).map_or(scale, |parent| {
+        let scale: Decimal = SCALING_RES.iter().fold(dec!(1.0), |scale, res| {
```
vihu, line 210: Just confirming that the previous version was going from 4..13 and now it's 13..4 (so no need to call `.rev()`)

jeffgrunewald: that's correct; these two constants used to be a range struct that incremented positively so we needed to reverse the iterator when working through it but now just creating the resolution slice in descending order

iot_verifier/src/poc.rs
```diff
@@ -671,42 +671,49 @@ fn calc_fpsl(freq: u64, distance_mtrs: u32) -> f64 {
...
+    let p1_latlng: LatLng = p1_cell.into();
+    let p2_latlng: LatLng = p2_cell.into();
+    let p1_coord: Coord = p1_latlng.into();
+    let p2_coord: Coord = p2_latlng.into();
 
     let (p1_x, p1_y) = p1_coord.x_y();
     let (p2_x, p2_y) = p2_coord.x_y();
     let p1_geo = point!(x: p1_x, y: p1_y);
     let p2_geo = point!(x: p2_x, y: p2_y);
     let distance = p1_geo.vincenty_distance(&p2_geo)?;
```
vihu, line 707: I think it's worth looking into whether https://docs.rs/h3o/latest/h3o/struct.LatLng.html#method.distance_m works for us. So instead of doing ad-hoc hex_adjustment just convert the h3 cell to latlng for p1 and p2, and call distance_m from h3o latlng. The latlng afaict is at the centroid of the h3 cell.

vihu: h3o `latlng::distance` does use haversine distance instead of vincenty distance but I don't think the difference is big enough to not do the simpler thing? Furthermore, dependent applications won't _really_ be able to reproduce the distance calculation unless they also do the hex adjustment stuff.

cc: @madninja

vihu: For bookkeeping, we will investigate this further and check if rewards, verifications align with the proposed change. It doesn't need to happen in this particular PR.

andymck (approved):

bbalser (approved):

PR #529, opened 2023-06-06
dont send empty gateway list to consumers

andymck:
Whilst running the IOT verifier locally over the previous epoch 24 hrs worth of history, encountered an issue whereby the refreshed gateway map received from iot config service was zero.  This was then sent to the consumers ( hex density scaler and the cache ) resulting in all poc reports failing verification.

`
2023-06-06T11:32:28.012131Z  INFO iot_verifier::gateway_updater: completed refreshing gateways, total gateways: 0

2023-06-06T11:32:28.149578Z  INFO iot_verifier::tx_scaler: density_scaler: generating hex scaling map, starting at 2023-06-04T11:32:28.149576Z

2023-06-06T11:32:30.969204Z  INFO iot_verifier::tx_scaler: density_scaler: scaling factor map entries: 0
`

This adds a guard so that in this scenario the refreshed map is dropped, and not sent to consumer.

A more robust solution may be to compare the updated map size against the previous map side and if non zero but significantly smaller then to also drop...would need to consider any side effects of such a solution

### DIFF
--- a/iot_verifier/src/gateway_updater.rs
+++ b/iot_verifier/src/gateway_updater.rs
@@ -70,7 +70,13 @@ impl GatewayUpdater {
     async fn handle_refresh_tick(&mut self) -> Result<(), GatewayUpdaterError> {
         tracing::info!("handling refresh tick");
         let updated_gateway_map = refresh_gateways(&mut self.iot_config_client).await?;
-        self.sender.send(updated_gateway_map)?;
+        let gateway_count = updated_gateway_map.len();
+        if gateway_count > 0 {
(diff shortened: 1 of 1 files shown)


bbalser (approved):
Nice catch!

andymck:
> This is a good catch on the verifier side but do we know why the map was updated with 0 records? Was this just an unexpected side effect of running the service locally against a non-production config service or is there a potential for this in production we should be tracking down and mitigating as well?

I suspect the root cause was that my local postgres instance was under heavy load.  I was seeing DB timeouts from some of the verifier requests to the DB.   I agree that we should consider additional guards on the iot config side

PR #530, opened 2023-06-06
subscriber location rewards

andymck:


mobile_packet_verifier/src/accumulate.rs
```diff
@@ -41,8 +41,8 @@ pub async fn accumulate_sessions(
         }
         sqlx::query(
             r#"
-            INSERT INTO data_transfer_sessions (pub_key, payer, uploaded_bytes, downloaded_bytes, first_timestamp, last_timestamp)
-            VALUES ($1, $2, $3, $4, $5, $5)
+            INSERT INTO data_transfer_sessions (pub_key, payer, uploaded_bytes, downloaded_bytes, subscriber_id, first_timestamp, last_timestamp)
```
bbalser, line 44: Why do we have a `first_timestamp` and `last_timestamp` if they are always set to the same value?

jeffgrunewald: a typo?

mobile_verifier/src/subsciber_location.rs
```diff
@@ -0,0 +1,179 @@
...
+    pub async fn save(
+        &self,
+        subscriber_id: Vec<u8>,
+        timestamp: DateTime<Utc>,
+        db: &mut Transaction<'_, Postgres>,
+    ) -> Result<(), sqlx::Error> {
+        sqlx::query(
+            r#"
+                INSERT INTO subscriber_loc (subscriber_id, date_bucket, hour_bucket, reward_timestamp)
+                VALUES ($1, DATE_TRUNC('hour', $2), DATE_PART('hour', $2), $2)
```
jeffgrunewald, line 110: i think this could be simplified to just 2-3 columns: the subscriber ID, and the timestamp, or a separate "processing/ingest" timestamp from the reward timestamp if there's a meaningful distinction to be had there.
my understanding of the `date_trunc` and the `date_part` functions is that they are used to retrieve sub-pieces of the datetime timestamp field, but they still all operate on a timestamptz that's stored in the db. so you would just insert the timestamptz in these save functions but then you could retrieve different components (hour component) in the select query when you're retrieving records to do reward eligibility calculation

mobile_verifier/src/subsciber_location.rs
```diff
@@ -0,0 +1,179 @@
...
+    )
+    .bind(reward_period.start)
+    .bind(reward_period.end)
+    .fetch(db);
+    let mut location_shares = LocationSharingMap::new();
+    while let Some(share) = rows.try_next().await? {
+        location_shares
+            .entry(share.subscriber_id)
+            .or_insert_with(Vec::new)
+            .push(share.hour_bucket)
```
jeffgrunewald, line 165: are these guaranteed to be in order based on ingest? is there any chance we'd need to order this vec right before you call `check_subscriber_location_shares_rewardable`?

mobile_verifier/src/cli/server.rs
```diff
@@ -42,6 +45,7 @@ impl Cmd {
         let store_base_path = std::path::Path::new(&settings.cache);
 
         let ingest = FileStore::from_settings(&settings.ingest).await?;
+        let data_transfer_ingest = FileStore::from_settings(&settings.data_transfer_ingest).await?;
```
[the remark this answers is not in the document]
andymck: addressed here: 59b78f66d1bcb920c9394f2fc6e2eb0f3b2fe156

reward_index/src/indexer.rs
```diff
@@ -139,7 +141,20 @@ impl Indexer {
                         },
                         r.dc_transfer_reward,
                     )),
-                    _ => bail!("got an invalid mobile reward share"),
+                    Some(MobileReward::SubscriberReward(r)) => {
+                        let key = match str::from_utf8(&r.subscriber_id) {
```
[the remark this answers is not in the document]
andymck: I will hold judgement until I know more about the actual implementation of these IDs

mobile_packet_verifier/src/accumulate.rs
bbalser: yup missed it changing in the on conflict.

mobile_verifier/src/rewarder.rs
```diff
@@ -198,6 +196,36 @@ impl Rewarder {
                 .await??;
         }
 
+        // *
+        // Subscriber rewards
+        // currently only includes rewards for subscriber location sharing
+        // but others will follow
+        // to be eligable for location sharing rewards a subscriber must have shared its location
+        // across a minimum of 3 hours in any 8 hour window within the epoch ( 24 hrs) AND
+        // also have transferred data within the epoch ( 24 hrs )
```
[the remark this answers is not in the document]
andymck: addressed here: 8c63c825c93d7bd3ca7a45b7f05e65069c7fb2ee

mobile_verifier/src/subsciber_location.rs
andymck: sorting ensured here: afb0bf72752c1d9c1e2ce2c096831dd2f07e652d

mobile_verifier/src/data_session.rs
```diff
@@ -0,0 +1,243 @@
...
+            .map(anyhow::Ok)
+            .try_fold(transaction, |mut transaction, report| async move {
+                // if we have a subscriber id, then we know this is
+                // a data transfer event from a mobile subscriber
+                match report.subscriber_id.clone() {
+                    Some(subscriber_id) => {
+                        let data_session = SubscriberDataSession::from_valid_data_session(
+                            report,
+                            subscriber_id,
+                            file_ts,
```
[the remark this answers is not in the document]
andymck: addressed here: afb0bf72752c1d9c1e2ce2c096831dd2f07e652d

mobile_verifier/src/data_session.rs
```diff
@@ -0,0 +1,243 @@
...
+
+pub async fn aggregate_subscriber_data_sessions_to_bytes<'a>(
+    exec: impl sqlx::PgExecutor<'a> + Copy + 'a,
+    epoch: &Range<DateTime<Utc>>,
+) -> Result<SubscriberMap, sqlx::Error> {
+    let stream = sqlx::query_as::<_, SubscriberDataSession>(
+        r#"
+            SELECT *
+            FROM subscriber_data_transfer_sessions
+            WHERE download_bytes > 0
```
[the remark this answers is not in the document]
jeffgrunewald: should it be `download_bytes > 0 || upload_bytes > 0`? if we're rewarding for usage of the network aren't uploads just as good as downloads?

mobile_verifier/src/reward_shares.rs
```diff
@@ -25,6 +29,20 @@ const DC_USD_PRICE: Decimal = dec!(0.00001);
 /// Default precision used for rounding
 const DEFAULT_PREC: u32 = 15;
 
+// Percent of total emissions allocated for mapper rewards
+const _MAPPERS_REWARDS_PERCENT: Decimal = dec!(0.2);
```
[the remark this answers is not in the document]
jeffgrunewald: forgot about silencing unused code warnings; didn't even think about that for a function name.

mobile_verifier/src/cli/server.rs
```diff
@@ -143,22 +166,40 @@ impl Cmd {
...
+        // TODO: retrieve initial carrier keys from config service
+        let subscriber_location_ingestor = SubscriberLocationIngestor::from_settings(
+            settings,
+            pool.clone(),
+            config_client.clone(),
+            Vec::new(),
+            subscriber_location_ingest,
+        );
+
+        let data_session_ingestor = DataSessionIngestor { pool: pool.clone() };
```
jeffgrunewald, line 180: this should probably be created from a `new` or `from_settings` function as well (new seems more appropriate here since it's just the one field)

mobile_verifier/src/reward_shares.rs
```diff
@@ -134,6 +141,61 @@ impl TransferRewards {
...
+    ) -> Self {
+        let mut discover_location_shares: Vec<Vec<u8>> = Vec::new();
+        active_subscribers
+            .into_iter()
+            .for_each(|(subscriber_id, data_transferred)| {
+                if subscriber_shares_rewardable(&location_shares, &subscriber_id, data_transferred)
+                {
+                    discover_location_shares.push(subscriber_id);
+                }
+            });
```
jeffgrunewald, line 162:
```suggestion
        let discover_location_shares: Vec<Vec<u8>> =
            active_subscribers
                .into_iter()
                .filter(|(subscriber_id, data_transferred)| {
                    subscriber_shares_rewardable(&location_shares, subscriber_id, *data_transferred)
                })
                .map(|(subscriber_id, _)| subscriber_id)
                .collect();
```

jeffgrunewald: tested

andymck: applied manually here: f6d4e3e8a44c9930b16e929d372cd03815078998

mobile_verifier/src/rewarder.rs
```diff
@@ -198,6 +196,42 @@ impl Rewarder {
...
 
+        // *
+        // Mapper rewards
+        // currently only includes rewards for discovery mapping
+        // but others will follow
+        // to be eligable for discovery mapping rewards a subscriber must have shared its location
+        // across a minimum of 3 hours in any 8 hour window within the epoch ( 24 hrs) AND
+        // also have consumed data within the epoch ( 24 hrs )
+        // consumed data applies to uploaded or downloaded data
+        // *
```
maplant, line 207:
```suggestion
        // Mapper rewards currently only includes rewards for discovery mapping.
        // To be eligible for discovery mapping rewards a subscriber must have 
        // shared its location across a minimum of 3 hours in any 8 hour window 
        // within the reward period (24 hours) AND also have consumed data within
        // the reward period. Consumed data includes uploaded and downloaded
        // data.
```

mobile_verifier/src/cli/server.rs
andymck: addressed here: 97084c89f45b29e241c4d7c6c2bc2c25c694197f

mobile_verifier/src/subscriber_location.rs
```diff
@@ -0,0 +1,179 @@
...
+
+    //TODO: reinstate check when keys are available
+    fn verify_known_carrier_key(&self, _public_key: &PublicKeyBinary) -> bool {
+        // self.carrier_keys.contains(public_key)
+        true
+    }
+
+    // TODO: update when actual API is available
+    async fn refresh_carrier_keys(&mut self) -> anyhow::Result<()> {
+        // match self
```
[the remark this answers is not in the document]
andymck: aha, nice.  added here: 6925d9940cbd0e9cc8d29b84c94463d2c15632e3

mobile_verifier/src/data_session.rs
```diff
@@ -0,0 +1,243 @@
...
+                        break;
+                    }
+                    Some(file) = receiver.recv() => self.process_file(file).await?,
+                }
+            }
+
+            Ok(())
+        })
+        .map_err(anyhow::Error::from)
+        .and_then(|result| async move { result })
```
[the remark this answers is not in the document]
jeffgrunewald: it just looks a little odd to flat_map the nested result when you can just
```
do_thing
.map_err(anyhow::Error::from)
.await?
```
and unwrap the outer result

mobile_verifier/src/data_session.rs
```diff
@@ -0,0 +1,243 @@
...
+    pub num_dcs: i64,
+    pub reward_timestamp: DateTime<Utc>,
+}
+
+#[derive(thiserror::Error, Debug)]
+#[error("data session error: {0}")]
+pub struct DataSessionError(#[from] sqlx::Error);
+
+impl SubscriberDataSession {
+    pub async fn save(self, db: &mut Transaction<'_, Postgres>) -> Result<(), DataSessionError> {
```
jeffgrunewald, line 113: in this case i'd skip the `DataSessionError` and just have the function return `Result<(), sqlx::Error>` since that's the only error this particular function can throw. the function calling this is returning an `anyhow::Result` which can safely handle receiving and throwing either the DataSessionError or the underlying sqlx::Error

mobile_verifier/src/data_session.rs
```diff
@@ -0,0 +1,243 @@
...
+        WHERE reward_timestamp >= $1 and reward_timestamp < $2
+        "#,
+    )
+    .bind(epoch.start)
+    .bind(epoch.end)
+    .fetch(exec);
+    data_sessions_to_dc(stream).await
+}
+
+pub async fn data_sessions_to_dc<'a>(
```
jeffgrunewald, line 193: since `data_sessions_to_dc` is only callable by `aggregate_hotspot_data_sessions_to_dc` and `data_sessions_to_bytes` is only callable by `aggregate_subscriber_data_sessions_to_bytes` and they both return the same types can they just be collapsed from four functions into two and avoid the stream pinning?

if it's for the purposes of testing the underlying function without needing to query a database i would argue the function is really more testing the stream iteration than the aggregation of dcs/bytes since the operation it's performing is relatively simple

mobile_verifier/src/data_session.rs
```diff
@@ -0,0 +1,243 @@
...
+    tokio::pin!(stream);
+    let mut map = SubscriberMap::new();
+    while let Some(session) = stream.try_next().await? {
+        *map.entry(session.subscriber_id).or_default() +=
+            (session.download_bytes + session.upload_bytes) as u64
+    }
+    Ok(map)
+}
+
+pub async fn clear_subscriber_data_sessions(
```
jeffgrunewald, line 234: i don't see us clearing hotspot data sessions? do we need to wipe those from the database once the rewards for the covered period have been generated as well?

mobile_verifier/src/data_session.rs
andymck: These functions were split out to aid testing and to allow the tests to follow real world usage as close as possible. Agree, the aggregation is simple but the value is in how they allow the tests to be constructed

mobile_verifier/src/reward_shares.rs
```diff
@@ -134,6 +141,65 @@ impl TransferRewards {
...
+    pub fn into_subscriber_rewards(
+        self,
+        reward_period: &'_ Range<DateTime<Utc>>,
+        reward_per_share: Decimal,
+    ) -> impl Iterator<Item = proto::MobileRewardShare> + '_ {
+        self.discovery_mapping_shares
+            .into_iter()
+            .map(move |subscriber_id| proto::SubscriberReward {
+                subscriber_id,
+                discovery_location_amount: (reward_per_share * DISCOVERY_MAPPING_POINTS_PER_SHARE)
```
jeffgrunewald, line 189: perhaps this is just a naming issue but multiplying these two variables seems to imply there's a missing component to the reward calculation to get the mobile bones in the `SubscriberReward`. i would expect to multiply two `_per_share` variables by a third `num_shares` or else `num_points * points_per_share` to get the number of shares and then multiply that by `reward_per_share` to get the reward bones

reward_index/src/indexer.rs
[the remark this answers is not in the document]
jeffgrunewald: that reference looks like it points to earlier in the file adding the `new` method to the data session struct. for this one, per a conversation with the foundation
> that's dumped directly from on-chain so it should be just bytes.
so i _believe_ we should be adding rewards to the rewards index in the same "just bytes" format of the subscriber id we've been passing around throughout the ingest/verification process so it's in the same format in the reward index DB when the user wallet goes looking for it to cash out accumulated rewards from their compatible wallet.

mobile_verifier/src/subscriber_location.rs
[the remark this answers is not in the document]
andymck: :( todo is nice but it blocks the code from running...

mobile_verifier/src/subscriber_location.rs
```diff
@@ -0,0 +1,200 @@
...
+
+    pub async fn save(
+        &self,
+        loc_ingest_report: &SubscriberLocationIngestReport,
+        db: &mut Transaction<'_, Postgres>,
+    ) -> Result<(), sqlx::Error> {
+        sqlx::query(
+            r#"
+                INSERT INTO subscriber_loc (subscriber_id, received_timestamp)
+                VALUES ($1, $2, $3)
```
bbalser, line 123: remove $3

bbalser (approved):

mobile_verifier/src/data_session.rs
```diff
@@ -0,0 +1,155 @@
...
+        .and_then(|result| async move { result })
+        .await
+    }
+
+    async fn process_file(
+        &self,
+        file_info_stream: FileInfoStream<ValidDataTransferSession>,
+    ) -> anyhow::Result<()> {
+        tracing::info!(
+            "handling valid data transfer file {:?}",
```
maplant, line 53:
```suggestion
            "handling valid data transfer file {}",
```

It's a string, no reason to use debug fmt here

mobile_verifier/src/data_session.rs
```diff
@@ -0,0 +1,155 @@
+use chrono::{DateTime, Utc};
+use file_store::{file_info_poller::FileInfoStream, mobile_transfer::ValidDataTransferSession};
+use futures::{
+    stream::{Stream, StreamExt, TryStreamExt},
+    TryFutureExt,
+};
+use helium_crypto::PublicKeyBinary;
+use sqlx::{PgPool, Postgres, Transaction};
+use std::collections::HashMap;
+use std::ops::Range;
```
jeffgrunewald, line 10:
```suggestion
use std::{
    collections::HashMap,
    ops::Range,
};
```

mobile_verifier/src/data_session.rs
jeffgrunewald: better still
```
tracing::info!(file = file_info_stream.file_info.key, "handling valid data transfer file");
```

mobile_verifier/src/reward_shares.rs
```diff
@@ -134,6 +133,66 @@ impl TransferRewards {
     }
 }
 
+#[derive(Default)]
+pub struct MapperShares {
+    pub discovery_mapping_shares: Vec<Vec<u8>>,
```
jeffgrunewald, line 138:
```suggestion
    pub discovery_mapping_shares: SubscriberLocationMap,
```
reuse the same "type" everywhere, even if it's just an alias. also we should think about renaming this type/alias name since it's no longer a map. `SubscriberValidatedLocations`?

mobile_verifier/src/rewarder.rs
```diff
@@ -198,6 +200,43 @@ impl Rewarder {
...
+            mapping_shares.into_subscriber_rewards(reward_period, rewards_per_share)
+        {
+            if self.disable_discovery_loc_rewards_to_s3 {
+                tracing::info!(
+                    "discovery location rewards output to s3 is disabled, outputting to logs only"
+                );
+                if let Some(ProtoReward::SubscriberReward(reward)) = mapping_share.reward {
+                    tracing::info!(
+                        "subscriber_id: {:?}, discovery_location_amount: {}",
+                        reward.subscriber_id,
```
jeffgrunewald, line 227: since this is just a byte array, will the debug output of the `subscriber_id` be valuable logging or will it just be visual noise? should we truncate it in case it outputs with weird formatting or base64 encode it or something?

file_store/src/cli/dump.rs
```diff
@@ -216,6 +221,21 @@ impl Cmd {
...
+                        "carrier_pub_key": report.report.carrier_pub_key,
+                        "recv timestamp": report.received_timestamp}))?;
+                }
+                FileType::VerifiedSubscriberLocationIngestReport => {
+                    let report = VerifiedSubscriberLocationIngestReport::decode(msg)?;
+                    print_json(&json!({
+                        "subscriber_id": report.report.report.subscriber_id,
+                        "carrier_pub_key": report.report.report.carrier_pub_key,
+                        "status": report.status,
+                        "recv timestamp": report.report.received_timestamp}))?;
```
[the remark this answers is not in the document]
andymck: the received ts here is meant to be that of the time the report hit the ingestor, which is as it is atm ie report.report.received_timestamp

file_store/src/mobile_subscriber.rs
```diff
@@ -30,23 +39,69 @@ impl MsgDecode for SubscriberLocationIngestReport {
...
         Ok(Self {
             subscriber_id: v.subscriber_id,
             timestamp: v.timestamp.to_timestamp()?,
-            pubkey: v.carrier_pub_key.into(),
+            carrier_pub_key: v.carrier_pub_key.into(),
         })
     }
 }
 
+impl From<SubscriberLocationReq> for SubscriberLocationReqV1 {
```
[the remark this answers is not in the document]
andymck: This is required, when verifying the subscriber location report in the mobile verifier we write out a paper trail to S3 with a status of whether the report is valid or invalid and at that point we go from `SubscriberLocationReq` to `SubscriberLocationReqV1`

mobile_verifier/src/settings.rs
```diff
@@ -22,9 +22,17 @@ pub struct Settings {
     pub output: file_store::Settings,
     pub metrics: poc_metrics::Settings,
     pub price_tracker: price::price_tracker::Settings,
-    pub config_client_settings: mobile_config::ClientSettings,
+    pub config_client: mobile_config::ClientSettings,
+    pub authorization_client: mobile_config::ClientSettings,
+    pub entity_client: mobile_config::ClientSettings,
```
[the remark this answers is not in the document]
andymck: yeah i think these can be the same, when writing i had it in my head we might want to use diff signing keys or something but now i think about it thats is unlikely

mobile_verifier/src/rewarder.rs
andymck: i think its fine as is, this is temp logging to show the discovery location rewards are being calculated in advance of them going live.  Once they are live this code should be removed

mobile_verifier/src/data_session.rs
andymck: fixed here: d4ab9d1ee85f20f2779672ba4a806b77614ec2e9

mobile_verifier/migrations/14_subscriber_location.sql
```diff
@@ -0,0 +1,5 @@
+create table subscriber_loc (
```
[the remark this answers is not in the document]
andymck: fixed here: d4ab9d1ee85f20f2779672ba4a806b77614ec2e9

mobile_verifier/src/data_session.rs
andymck: addressed here: d4ab9d1ee85f20f2779672ba4a806b77614ec2e9

file_store/src/cli/dump.rs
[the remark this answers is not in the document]
jeffgrunewald: I’m pointing out you have json keys with spaces in them which I’m not sure they parse correctly across all languages and tools that might be reading this json output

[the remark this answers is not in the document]
andymck: aha, missed that completely. addressed here: 5066e377cfbfebead31fdacf76cfbe8174baab56

jeffgrunewald (approved):

PR #533, opened 2023-06-07
Fix speedtest migration. Don't drop data; migrate

maplant:


bbalser (approved):
Ran it a few times locally starting from the prod pg_dump and it all looks correct to me

jeffgrunewald (approved):

andymck (approved):

PR #540, opened 2023-06-12
emit metric gauge whenever reward indexer processes rewards

jeffgrunewald:
emit a trackable gauge metric whenever the reward indexer processes a reward file to allow alerting when a defined period of time has passed without updating the value as expected


jeffgrunewald:
> I thought the plan was to emit this from the verifier at the point of reward being committed to s3...ie testing that the rewards were generated

it was in both; that gauge has already been merged into the verifier but @bbalser wanted to have one in the reward indexer as well to make sure if something fell down there we didn't miss it just b/c everything was happy in the verifier

bbalser (approved):

PR #542, opened 2023-06-13
Fixed authorization in RouteService::update

ederuiter:
This fixes a authorization bypass vulnerability which would allow users to update routes of other oui's by specifying the route id of another oui together with their own oui. For example: `{route: {id: <not my route id>, oui: <my oui id>, signer: <my oui public key>, signature: <signature based on my oui's private key>}}` The impact is mimimal as the route id's are uuid's (although mac/time based ones).

### DIFF
--- a/iot_config/src/route_service.rs
+++ b/iot_config/src/route_service.rs
@@ -296,7 +296,7 @@ impl iot_config::Route for RouteService {
         );
 
         let signer = verify_public_key(&request.signer)?;
-        self.verify_request_signature(&signer, &request, OrgId::Oui(route.oui))
+        self.verify_request_signature(&signer, &request, OrgId::RouteId(route.id))
             .await?;
 
         let updated_route = route::update_route(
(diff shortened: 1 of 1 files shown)


iot_config/src/route_service.rs
```diff
@@ -296,7 +296,7 @@ impl iot_config::Route for RouteService {
         );
 
         let signer = verify_public_key(&request.signer)?;
-        self.verify_request_signature(&signer, &request, OrgId::Oui(route.oui))
+        self.verify_request_signature(&signer, &request, OrgId::RouteId(route.id))
```
jeffgrunewald, line 299:
```suggestion
        self.verify_request_signature(&signer, &request, OrgId::RouteId(&route.id))
```

jeffgrunewald (approved):

bbalser (approved):

jeffgrunewald:
Thanks for the catch, @ederuiter ; approved!

PR #546, opened 2023-06-15
Mobile config service supports subscriber rewards

jeffgrunewald:
- Restructures the API around verifying authorized mobile network signing keys are registered to the config service (carrier or mobile router keys)
- Adds verification for generic rewardable entities found on the blockchain or not by binary entity_id
- Adds clients for both
- Restructures the existing mobile gateway client code to align with the additional mobile clients


mobile_config/src/client/gateway_client.rs
```diff
@@ -0,0 +1,110 @@
...
+        &mut self,
+    ) -> Result<gateway_info::GatewayInfoStream, Self::Error> {
+        let mut req = mobile_config::GatewayInfoStreamReqV1 {
+            batch_size: self.batch_size,
+            signer: self.signing_key.public_key().into(),
+            signature: vec![],
+        };
+        req.signature = self.signing_key.sign(&req.encode_to_vec())?;
+        tracing::debug!("fetching gateway info stream");
+        let pubkey = Arc::new(self.config_pubkey.clone());
```
andymck, line 90: the Arc'ing here seems redundant

iot_packet_verifier/src/daemon.rs
```diff
@@ -155,7 +156,9 @@ impl Cmd {
         .create()
         .await?;
 
-        let org_client = settings.connect_org();
+        let org_client = Arc::new(Mutex::new(OrgClient::from_settings(
```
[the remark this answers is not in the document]
jeffgrunewald: The arc isn’t necessary directly on the org client b/c it implements clone so normally you’d just clone the client but you _have_ to wrap a mutex in an arc to share it (which we do in this context) and in this case I was pretty sure I recalled @maplant had used the arc/mutex to ensure the task monitoring funds and the verifier task weren’t modifying the same orgs simultaneously

mobile_config/src/client/gateway_client.rs
jeffgrunewald: In this case we need to b/c we’re moving the cloned key into the async closure within the stream to verify the signature on each batch of gateways returned. We have to clone it the once to copy it out of the client itself but then we move the it arc-wrapped into the closure

andymck: I tried it locally and clippy doesnt highlight any probs in the closure with the arc removed

mobile_verifier/src/cli/server.rs
```diff
@@ -116,7 +116,7 @@ impl Cmd {
         .await?;
 
         let reward_period_hours = settings.rewards;
-        let config_client = Client::from_settings(&settings.config_client)?;
+        let config_client = GatewayClient::from_settings(&settings.config_client)?;
```
andymck, line 119:
```suggestion
        let gateway_client = GatewayClient::from_settings(&settings.config_client)?;
```

mobile_verifier/src/heartbeats.rs
```diff
@@ -44,15 +44,15 @@ impl From<HeartbeatKey> for HeartbeatReward {
...
+    config_client: GatewayClient,
     heartbeats: Receiver<FileInfoStream<CellHeartbeatIngestReport>>,
     file_sink: FileSinkClient,
 }
 
 impl HeartbeatDaemon {
     pub fn new(
         pool: sqlx::Pool<sqlx::Postgres>,
-        config_client: Client,
+        config_client: GatewayClient,
```
andymck, line 55: Similar to above suggestion be good to have all the internal instances of `config_client` renamed to `gateway_client`

bbalser (approved):

andymck (approved):

maplant (approved):
Looks fine on the iot packet verifier side. glad to have this stuff moved into the client

PR #548, opened 2023-06-17
Mobile Config Cli

jeffgrunewald:
Provide a cli for interacting with the mobile config service:
- query the gateway service by public key to verify the gateway is on-chain and retrieve any registered metadata
- add/remove administrative signing and verification public keys
- verify carrier and router public keys are registered to the service for validating ingested reports
- verify rewardable entities are found on-chain (submit the base64 encoded string of the binary entity_id)

Also provides a dockerfile/compose definition for running the mobile config server


mobile_config_cli/src/cmds/mod.rs
```diff
@@ -0,0 +1,216 @@
...
+        env = ENV_CONFIG_HOST,
+        default_value = "https://config.mobile.mainnet.helium.io:6080"
+    )]
+    pub config_host: String,
+
+    #[arg(
+        global = true,
+        long,
+        env = ENV_CONFIG_PUBKEY,
+        default_value = "137oJzq1qZpSbzHawaysTGGsRCYTXG1MiTMQNxYSsQJp4YMDdN8"
```
jeffgrunewald, line 36: If someone has access to the server’s signing key (@bbalser) it’d be nice to pull this value from it and update the PR with the correct value

maplant (approved):

bbalser (approved):

PR #559, opened 2023-06-28
Modeled Coverage

maplant:
This is mostly complete, there are two remaining tasks that should be relatively quick to complete:

- [ ] Convert unit tests in the reward_share module over
- [ ] Add a cache to the heartbeat deamon to fetch coverage objects so we can invalidate heartbeats that use incorrect coverage objects

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -97,7 +97,7 @@ itertools = "*"
 data-credits = {git = "https://github.com/helium/helium-program-library.git", tag = "v0.1.0"}
 helium-sub-daos = {git = "https://github.com/helium/helium-program-library.git", tag = "v0.1.0"}
 price-oracle = {git = "https://github.com/helium/helium-program-library.git", tag = "v0.1.0"}
-uuid = {version = "1.3.4", features = ["serde"]}
+uuid = {version = "1.3.4", features = ["v4", "serde"]}
 
 [patch.crates-io]
 sqlx = { git = "https://github.com/helium/sqlx.git", rev = "92a2268f02e0cac6fccb34d3e926347071dbb88d" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -57,7 +57,7 @@ sqlx = {version = "0", features = [
 ]}
 
 helium-crypto = {version = "0.6.8", features=["sqlx-postgres", "multisig"]}
-helium-proto = {git = "https://github.com/helium/proto", branch = "master", features = ["services"]}
+helium-proto = {git = "https://github.com/helium/proto", branch = "map/modeled-coverage", features = ["services"]}
 hextree = "*"
 solana-client = "1.14"
 solana-sdk = "1.14"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -83,8 +83,8 @@ prost = "*"
 once_cell = "1"
 lazy_static = "1"
 config = {version="0", default-features=false, features=["toml"]}
-h3o = "0"
-xorf = {version = "0", features = ["serde"] }
+h3o = {version = "0", features = ["serde"]}
+xorf = {version = "0", features = ["serde"]}
 bytes = "*"
 structopt = "0"
 bincode = "1"
@@ -97,6 +97,7 @@ itertools = "*"
 data-credits = {git = "https://github.com/helium/helium-program-library.git", tag = "v0.1.0"}
 helium-sub-daos = {git = "https://github.com/helium/helium-program-library.git", tag = "v0.1.0"}
 price-oracle = {git = "https://github.com/helium/helium-program-library.git", tag = "v0.1.0"}
+uuid = {version = "1.3.4", features = ["serde"]}
 
 [patch.crates-io]
 sqlx = { git = "https://github.com/helium/sqlx.git", rev = "92a2268f02e0cac6fccb34d3e926347071dbb88d" }
--- a/file_store/src/coverage.rs
+++ b/file_store/src/coverage.rs
@@ -1,0 +1,78 @@
+use crate::{
+    error::DecodeError,
+    traits::{MsgDecode, TimestampDecode},
+    Error, Result,
+};
+use chrono::{DateTime, Utc};
+use helium_crypto::PublicKeyBinary;
+use helium_proto::services::poc_mobile::{
+    CoverageObjectIngestReportV1, CoverageObjectReqV1,
+    RadioHexSignalLevel as RadioHexSignalLevelProto, SignalLevel,
+};
+use serde::{Deserialize, Serialize};
+use uuid::Uuid;
+
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct RadioHexSignalLevel {
+    pub location: h3o::CellIndex,
+    pub signal_level: SignalLevel,
+    pub signal_power: f64,
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct CoverageObject {
+    pub pub_key: PublicKeyBinary,
+    pub uuid: Uuid,
+    pub cbsd_id: String,
+    pub coverage_claim_time: DateTime<Utc>,
+    pub indoor: bool,
+    pub coverage: Vec<RadioHexSignalLevel>,
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct CoverageObjectIngestReport {
+    pub received_timestamp: DateTime<Utc>,
+    pub report: CoverageObject,
+}
+
+impl MsgDecode for CoverageObject {
+    type Msg = CoverageObjectReqV1;
+}
+
+impl MsgDecode for CoverageObjectIngestReport {
+    type Msg = CoverageObjectIngestReportV1;
+}
+
+impl TryFrom<CoverageObjectReqV1> for CoverageObject {
+    type Error = Error;
+
+    fn try_from(v: CoverageObjectReqV1) -> Result<Self> {
+        let coverage: Result<Vec<RadioHexSignalLevel>> = v
+            .coverage
+            .into_iter()
+            .map(RadioHexSignalLevel::try_from)
+            .collect();
+        Ok(Self {
+            pub_key: v.pub_key.into(),
+            uuid: Uuid::from_slice(&v.uuid).map_err(DecodeError::from)?,
+            cbsd_id: v.cbsd_id,
+            coverage_claim_time: v.coverage_claim_time.to_timestamp()?,
+            indoor: v.indoor,
+            coverage: coverage?,
+        })
+    }
+}
+
+impl TryFrom<RadioHexSignalLevelProto> for RadioHexSignalLevel {
+    type Error = Error;
+
+    fn try_from(v: RadioHexSignalLevelProto) -> Result<Self> {
+        Ok(Self {
+            signal_level: SignalLevel::from_i32(v.signal_level).ok_or_else(|| {
+                DecodeError::unsupported_signal_level("coverage_object_req_v1", v.signal_level)
+            })?,
+            signal_power: v.signal_power,
+            location: v.location.parse().map_err(DecodeError::from)?,
+        })
+    }
+}
--- a/file_store/src/coverage.rs
+++ b/file_store/src/coverage.rs
@@ -43,6 +43,20 @@ impl MsgDecode for CoverageObjectIngestReport {
     type Msg = CoverageObjectIngestReportV1;
 }
 
+impl TryFrom<CoverageObjectIngestReportV1> for CoverageObjectIngestReport {
+    type Error = Error;
+
+    fn try_from(v: CoverageObjectIngestReportV1) -> Result<Self> {
+        Ok(Self {
+            received_timestamp: v.received_timestamp.to_timestamp_millis()?,
+            report: v
+                .report
+                .ok_or_else(|| Error::not_found("ingest coverage object report"))?
+                .try_into()?,
+        })
+    }
+}
+
 impl TryFrom<CoverageObjectReqV1> for CoverageObject {
     type Error = Error;
 
@@ -74,5 +88,15 @@ impl TryFrom<RadioHexSignalLevelProto> for RadioHexSignalLevel {
             signal_power: v.signal_power,
             location: v.location.parse().map_err(DecodeError::from)?,
         })
+    }
+}
+
+impl Into<RadioHexSignalLevelProto> for RadioHexSignalLevel {
+    fn into(self) -> RadioHexSignalLevelProto {
+        RadioHexSignalLevelProto {
+            signal_level: self.signal_level as i32,
+            signal_power: self.signal_power,
+            location: self.location.to_string(),
+        }
     }
 }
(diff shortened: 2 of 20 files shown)


file_store/src/coverage.rs
```diff
@@ -0,0 +1,102 @@
...
+impl MsgDecode for CoverageObjectIngestReport {
+    type Msg = CoverageObjectIngestReportV1;
+}
+
+impl TryFrom<CoverageObjectIngestReportV1> for CoverageObjectIngestReport {
+    type Error = Error;
+
+    fn try_from(v: CoverageObjectIngestReportV1) -> Result<Self> {
+        Ok(Self {
+            received_timestamp: v.received_timestamp.to_timestamp_millis()?,
```
andymck, line 51: any reason not to implement the MsgTimestamp trait to handle these timestamps here and in subsequent msgs below ?

Cargo.toml
```diff
@@ -57,7 +57,7 @@ sqlx = {version = "0", features = [
 ]}
 
 helium-crypto = {version = "0.6.8", features=["sqlx-postgres", "multisig"]}
-helium-proto = {git = "https://github.com/helium/proto", branch = "master", features = ["services"]}
+helium-proto = {git = "https://github.com/helium/proto", branch = "map/modeled-coverage", features = ["services"]}
```
andymck, line 60: just a note to update the `beacon` branch of proto at same time otherwise compile fails due to multiple version of proto in use

mobile_verifier/migrations/13_modeled_coverage.sql
```diff
@@ -0,0 +1,16 @@
+CREATE TYPE signal_level as ENUM (
+       'no',
```
andymck, line 2: `no` sounds odd here, does `none` or `zero` work better?

mobile_verifier/src/coverage.rs
```diff
@@ -0,0 +1,466 @@
...
+            let CoverageObject {
+                pub_key,
+                uuid,
+                cbsd_id,
+                coverage_claim_time,
+                indoor,
+                coverage,
+            } = report;
+
+            // TODO: Validate pub key is authorized
```
andymck, line 113: todo need removed ?

mobile_verifier/src/coverage.rs
```diff
@@ -0,0 +1,466 @@
...
+                .await?
+            {
+                CoverageObjectValidity::Valid
+            } else {
+                CoverageObjectValidity::InvalidPubKey
+            };
+
+            self.file_sink
+                .write(
+                    proto::CoverageObject {
```
andymck, line 126: Should we be creating proto struct directly ?  i understood that the convention was to create an internal struct and then perform an into/try_into to get to the proto.  

If we have properly defined filestore conversions it ensures there are no mishaps with 
creating proto struct manually ( ie getting a timestamp precision incorrect )

We should clarify amongst ourselves if this convention is valid, as i see both approaches in use in different parts of the code.

file_store/src/coverage.rs
```diff
@@ -0,0 +1,102 @@
...
+            signal_level: SignalLevel::from_i32(v.signal_level).ok_or_else(|| {
+                DecodeError::unsupported_signal_level("coverage_object_req_v1", v.signal_level)
+            })?,
+            signal_power: v.signal_power,
+            location: v.location.parse().map_err(DecodeError::from)?,
+        })
+    }
+}
+
+impl Into<RadioHexSignalLevelProto> for RadioHexSignalLevel {
```
andymck, line 94: why not implement the From instead of the Into, get the into for free ?

mobile_verifier/src/coverage.rs
```diff
@@ -0,0 +1,466 @@
...
+                    anyhow::Ok(default_hex_coverage("3", SignalLevel::Medium)),
+                    anyhow::Ok(default_hex_coverage("4", SignalLevel::High)),
+                    anyhow::Ok(default_hex_coverage("5", SignalLevel::Medium)),
+                    anyhow::Ok(default_hex_coverage("6", SignalLevel::Low)),
+                    anyhow::Ok(default_hex_coverage("7", SignalLevel::No)),
+                ]),
+            )
+            .await
+            .unwrap();
+        let rewards: Vec<_> = covered_hexes.into_iter().collect();
```
andymck, line 334: Is my understanding correct that this only returns a single radio and not the top 5 as the requirement is to return up to 5 radios only of the *same* strength and as there is only a single radio with High only that radio is returned ? If 2 radios had high then 2 would have been returned here ?

mobile_verifier/src/reward_shares.rs
```diff
@@ -141,60 +142,96 @@ pub fn dc_to_mobile_bones(dc_amount: Decimal, mobile_bone_price: Decimal) -> Dec
...
+        for CoverageReward {
+            cbsd_id,
+            points,
+            hotspot,
+        } in covered_hexes.into_iter()
+        {
+            // Guaranteed that points contains the given hotspot.
+            *coverage_points
+                .get_mut(&hotspot)
+                .unwrap()
```
andymck, line 213: the olde unwrap usage

mobile_verifier/src/coverage.rs
maplant: Yup, requirement is oldest five of the top strength

andymck (commented):
this looks to be doing the right thing, will give it another pass when conflicts and tests are fixed

mobile_verifier/src/heartbeats.rs
```diff
@@ -231,10 +252,11 @@ impl Heartbeat {
         Ok(
             sqlx::query_as::<_, HeartbeatSaveResult>(
                 r#"
-                INSERT INTO heartbeats (cbsd_id, hotspot_key, cell_type, latest_timestamp, truncated_timestamp)
-                VALUES ($1, $2, $3, $4, $5)
+                INSERT INTO heartbeats (cbsd_id, hotspot_key, cell_type, latest_timestamp, truncated_timestamp, coverage_object)
```
jeffgrunewald, line 255: did i miss a migration to add the coverage_object uuid to the heartbeats table?

maplant: yup!

maplant: I mean - nope! I forgot to add it! my b lol

mobile_verifier/src/coverage.rs
```diff
@@ -291,6 +334,72 @@ impl CoveredHexes {
...
+}
+
+#[derive(Clone)]
+pub struct CachedCoverage {
+    pub cbsd_id: String,
+    coverage: Vec<CellIndex>,
+}
+
+impl CachedCoverage {
+    pub fn max_distance_km(&self, latlng: LatLng) -> f64 {
```
jeffgrunewald, line 395: would it be more accurate or more efficient to do this distance calculation with the native h3 grid_distance functions (`grid_disk_distances_fast/safe()`)? could we confidently say due to RF propagation that coverage has to be within N range of the k-ring from the "source", get the `(CellIndex, u32)` iterator and return the furthest of those that's also a member of the `coverage` vec?

mobile_verifier/src/coverage.rs
```diff
@@ -0,0 +1,464 @@
...
+    hexes: HashMap<CellIndex, BTreeMap<SignalLevel, BinaryHeap<CoverageLevel>>>,
+}
+
+pub fn covered_hex_stream<'a>(
+    pool: &'a Pool<Postgres>,
+    cbsd_id: &str,
+    coverage_obj: &Uuid,
+) -> BoxStream<'a, Result<HexCoverage, sqlx::Error>> {
+    sqlx::query_as("SELECT * FROM coverage WHERE cbsd_id = $1 AND uuid = $2")
+        .bind(cbsd_id.to_string())
```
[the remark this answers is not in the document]
maplant: it does

mobile_verifier/migrations/13_modeled_coverage.sql
```diff
@@ -0,0 +1,16 @@
...
+       'low',
+       'medium',
+       'high'
+);
+
+CREATE TABLE hex_coverage (
+       uuid UUID NOT NULL,
+       hex INTEGER NOT NULL,
```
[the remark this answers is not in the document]
jeffgrunewald:
```suggestion
       hex BIGINT NOT NULL,
```

ingest/src/settings.rs
```diff
@@ -1,5 +1,5 @@
 use config::{Config, Environment, File};
-use helium_crypto::Network;
+use helium_crypto::{Network, PublicKey};
```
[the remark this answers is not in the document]
jeffgrunewald:
```suggestion
use helium_crypto::Network;
```

mobile_verifier/migrations/13_modeled_coverage.sql
```diff
@@ -0,0 +1,30 @@
...
+       'high'
+);
+
+CREATE TABLE hex_coverage (
+       uuid UUID NOT NULL,
+       hex BIGINT NOT NULL,
+       indoor BOOLEAN NOT NULL,
+       cbsd_id TEXT NOT NULL,
+       signal_level signal_level NOT NULL,
+       coverage_claim_time TIMESTAMPTZ NOT NULL
```
bbalser, line 14: need a `,` at the end of this line and I think this migration needs to be renamed to `15_modeled_coverage.sql`

mobile_verifier/src/coverage.rs
```diff
@@ -0,0 +1,584 @@
...
+                .auth_client
+                .verify_authorized_key(&pub_key, NetworkKeyRole::MobilePcs)
+                .await?
+            {
+                CoverageObjectValidity::Valid
+            } else {
+                CoverageObjectValidity::InvalidPubKey
+            };
+
+            self.file_sink
```
bbalser, line 125: You need to call commit on this file_sink

mobile_verifier/src/coverage.rs
```diff
@@ -0,0 +1,584 @@
...
+
+            if !matches!(validity, CoverageObjectValidity::Valid) {
+                continue;
+            }
+
+            for hex in coverage {
+                let location: u64 = hex.location.into();
+                sqlx::query(
+                    r#"
+                    INSERT INTO coverage_objects
```
bbalser, line 153: I believe the table name should be `hex_coverage` in this query

file_store/src/coverage.rs
maplant: Eh, just didn't see a reason to implement it to only be used one other place

mobile_verifier/src/coverage.rs
```diff
@@ -0,0 +1,466 @@
...
+                .write(
+                    proto::CoverageObject {
+                        pub_key: pub_key.into(),
+                        uuid: Vec::from(uuid.into_bytes()),
+                        cbsd_id: cbsd_id.clone(),
+                        coverage_claim_time: coverage_claim_time.encode_timestamp(),
+                        indoor,
+                        // It's pretty weird that we have to convert these back and forth, but it
+                        // shouldn't be too inefficient
+                        coverage: coverage.clone().into_iter().map(Into::into).collect(),
```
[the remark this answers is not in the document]
andymck: my take is that we should have an internal type and especially where we are deriving the source data from the DB and then subsequently manipulating that data.  I would then argue if we then have to convert the internal type to the proto type that should be handled in the filestore where we can have conversions centralised and have a standard set of conventions ( such as timestamp traits ).  Generating protos on the fly leaves more opportunity for mishaps such as mis-setting a timestamp ( secs insteads of millis for example )

mobile_verifier/migrations/15_modeled_coverage.sql
```diff
@@ -0,0 +1,25 @@
...
+       uuid UUID NOT NULL,
+       hex BIGINT NOT NULL,
+       indoor BOOLEAN NOT NULL,
+       cbsd_id TEXT NOT NULL,
+       signal_level signal_level NOT NULL,
+       coverage_claim_time TIMESTAMPTZ NOT NULL,
+       PRIMARY KEY (uuid, hex)
+);
+
+CREATE TABLE coverage_claim_time (
```
bbalser, line 18: I don't really like the naming of this table, it is really tracking the seniority of the cbsd_id not the coverage_claim_time, sometimes the seniority is set from coverage_claim_time and sometimes it is set from latest_heartbeat.

mobile_verifier/src/heartbeats.rs
```diff
@@ -136,11 +187,31 @@ impl HeartbeatReward {
     ) -> impl Stream<Item = Result<HeartbeatReward, sqlx::Error>> + 'a {
         sqlx::query_as::<_, HeartbeatKey>(
             r#"
-            SELECT hotspot_key, cbsd_id, cell_type
+            WITH coverage_objs AS (
```
bbalser, line 190: Is it possible for this query to pull a new coverage_object that is attached to a heartbeat that came after the rewardable period?

mobile_verifier/src/reward_shares.rs
```diff
@@ -265,43 +327,65 @@ impl PocShares {
...
-            let start_period = epoch.start.encode_timestamp();
-            let end_period = epoch.end.encode_timestamp();
-            Some(
-                self.hotspot_shares
+        available_poc_rewards
+            .checked_div(total_shares)
+            .map(|poc_rewards_per_share| {
+                let start_period = epoch.start.encode_timestamp();
+                let end_period = epoch.end.encode_timestamp();
+                self.coverage_points
```
bbalser, line 335: There has to be a way to make this more readable

mobile_verifier/src/coverage.rs
```diff
@@ -0,0 +1,719 @@
...
+pub const MAX_RADIOS_PER_HEX: usize = 5;
+
+#[async_trait::async_trait]
+pub trait CoveredHexStream {
+    async fn covered_hex_stream<'a>(
+        &'a self,
+        cbsd_id: &'a str,
+        coverage_obj: &'a Uuid,
+        period_end: DateTime<Utc>,
+    ) -> Result<BoxStream<'a, Result<HexCoverage, sqlx::Error>>, sqlx::Error>;
```
jeffgrunewald, line 290: are these errors overly prescriptive for a trait? why have the trait at all if it always assumes a postgres/sql source?

mobile_verifier/src/coverage.rs
```diff
@@ -0,0 +1,719 @@
...
+            WHERE
+              cbsd_id = $1 AND
+              inserted_at <= $2
+            ORDER BY inserted_at DESC
+            LIMIT 1
+            "#,
+        )
+        .bind(cbsd_id)
+        .bind(period_end)
+        .fetch_one(self)
```
jeffgrunewald, line 322: is always guaranteed to be one? what happens when we first activate the feature and it's populating the new record types?

jeffgrunewald (approved):

mobile_verifier/migrations/15_modeled_coverage.sql
```diff
@@ -21,6 +21,7 @@ CREATE TABLE seniority (
        seniority_ts TIMESTAMPTZ NOT NULL,
        last_heartbeat TIMESTAMPTZ NOT NULL,
        uuid UUID NOT NULL,
+       update_reason INT NOT NULL,
```
bbalser, line 24: why is update_reason an int and not an enum?

PR #565, opened 2023-07-07
fail region params lookup on db error

jeffgrunewald:
when gateway isn't found on the solana chain (the query succeeds but no record is returned) continue to return the "fallback" response of the region passed in the incoming request and a 0 gain but now in the event of a failure to lookup the requested gateway return a failure to the gateway requesting the region params

### DIFF
--- a/iot_config/src/gateway_service.rs
+++ b/iot_config/src/gateway_service.rs
@@ -184,13 +184,22 @@ impl iot_config::Gateway for GatewayService {
         })?;
 
         let (region, gain) = match self.resolve_gateway_info(address).await {
-            Err(_) => {
+            Err(status) if status.code() == tonic::Code::NotFound => {
                 tracing::debug!(
                     pubkey = %address,
                     %default_region,
                     "unable to retrieve gateway from chain"
                 );
                 (default_region, 0)
+            }
+            Err(status) => {
+                tracing::error!(
+                    pubkey = %address,
+                    %default_region,
+                    "gateway lookup failed"
+                );
+                telemetry::count_region_lookup(default_region, Region::Unknown);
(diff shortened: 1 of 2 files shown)


iot_config/src/gateway_service.rs
```diff
@@ -184,14 +184,23 @@ impl iot_config::Gateway for GatewayService {
...
                 );
                 (default_region, 0)
             }
+            Err(status) => {
+                tracing::error!(
+                    pubkey = %address,
+                    %default_region,
+                    "gateway lookup failed"
+                );
+                telemetry::count_region_lookup(default_region, Region::Unknown);
```
madninja, line 201: why count this as a region unknown?

jeffgrunewald: i didn't want to not count the request in the event we had unexpected errors and the failure count suddenly spiked and it seemed technically correct that we couldn't in fact verify the region. also i didn't want to make the metric count take an option for this probably unlikely scenario but i could do that as well

madninja: would it then not bucket the failed lookups with the unknown region lookups? i.e. it'd miscount unasserted DO hotspots?

jeffgrunewald: Yeah, thinking more about it that second argument should be an option

jeffgrunewald: fixed [d93d2b6](https://github.com/helium/oracles/pull/565/commits/d93d2b64a7f2bad7e20f19d220e8adf5c07b2af4)

madninja (approved):

maplant (approved):

PR #566, opened 2023-07-08
allow mobile config service to register pcs signing keys

jeffgrunewald:
depends on https://github.com/helium/proto/pull/352

allow the mobile config service to register pcs signing keys so the mobile verifier is able to validate subscriber location and data usage threshold reports submitted by a carrier's propagation calculation service

cleans up duplication of the KeyRole type shared by the mobile config service and its CLI

- temporarily adjusts mobile ingest service and cell heartbeat req messages to allow compilation to succeed with upstream changes already in the proto repo

### DIFF
--- a/mobile_config/migrations/20230708171204_add_pcs_key_role.sql
+++ b/mobile_config/migrations/20230708171204_add_pcs_key_role.sql
@@ -1,1 +1,1 @@
-ALTER TYPE key_role ADD VALUE IF NOT EXISTS 'pcs';-- Add migration script here
+ALTER TYPE key_role ADD VALUE IF NOT EXISTS 'pcs';
--- a/mobile_config/migrations/20230708171204_add_pcs_key_role.sql
+++ b/mobile_config/migrations/20230708171204_add_pcs_key_role.sql
@@ -1,0 +1,1 @@
+ALTER TYPE key_role ADD VALUE IF NOT EXISTS 'pcs';-- Add migration script here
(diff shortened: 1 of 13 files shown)


mobile_config/migrations/20230708171204_add_pcs_key_role.sql
```diff
@@ -0,0 +1 @@
+ALTER TYPE key_role ADD VALUE IF NOT EXISTS 'pcs';-- Add migration script here
```
madninja, line 1: “Add migration script here”?

mobile_config/src/lib.rs
```diff
@@ -20,3 +22,70 @@ pub type GrpcStreamResult<T> = ReceiverStream<Result<T, Status>>;
...
+            ProtoKeyRole::Router => Self::Router,
+            ProtoKeyRole::Pcs => Self::Pcs,
+        }
+    }
+}
+
+impl std::fmt::Display for KeyRole {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        let s = match self {
+            Self::Administrator => "Administrator",
```
madninja, line 83: Match on ignore case? Because this ain’t snake case

mobile_config_cli/src/lib.rs
```diff
@@ -137,6 +104,7 @@ impl Display for NetworkKeyRole {
         match self {
             NetworkKeyRole::MobileCarrier => write!(f, "Carrier"),
             NetworkKeyRole::MobileRouter => write!(f, "Router"),
+            NetworkKeyRole::MobilePcs => write!(f, "PCS"),
```
madninja, line 107: Not snake case here either

madninja (commented):
Some minor comments. And the proto branch is merged

mobile_config/migrations/20230708171204_add_pcs_key_role.sql
jeffgrunewald: ha. that's the default comment the migration create script leaves; thought i'd highlighted and typed over it.

maplant (approved):

madninja (approved):

PR #567, opened 2023-07-11
IoT Config locked route updates and shutdown listeners

jeffgrunewald:
Ensures that locked/disabled routes are still broadcast to the HPR servers so they are always available. HPRs check for a route's status (active and not disabled) anyway so they are _only_ removed from the HPR when they are purposely deleted by an admin or operator.

Also adds more shutdown listeners to the long-running grpc streaming handlers to try and force the config service to respect and properly respond to sigterm messages sent from systemd when the service is being shut down (such as for update deployments).

### DIFF
--- a/iot_config/src/gateway_service.rs
+++ b/iot_config/src/gateway_service.rs
@@ -35,6 +35,7 @@ pub struct GatewayService {
     region_map: RegionMapReader,
     signing_key: Arc<Keypair>,
     delegate_cache: watch::Receiver<org::DelegateCache>,
+    shutdown: triggered::Listener,
 }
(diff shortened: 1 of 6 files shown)


macpie (approved):

jeffgrunewald:
After feedback from @maplant i modified the long-running streaming results responses to switch off a `tokio::select!`. He pointed out the previous versions of these shutdown checks would have been polling unnecessarily on each iteration of a loop instead of waiting on the select which will preempt the longer-running function if a shutdown is initiated.

bbalser (approved):

PR #576, opened 2023-07-20
Add a taskmanager to manage start stop order of processes

andymck:
This takes bbalser's original work on the taskmanager, updates it and extends it to all oracles.  

The intent to to implement a controlled start / stop order for tasks/processes.

Some considerations I still need to circle back / conclude on:

- [x] The gateway and region caches, prior to this change, were passed by reference to the run function of a process ( for example the runner in the iot verifier ).  With the changes here that is no longer possible as the taskman API changes the run API to only take the shutdown listener and moves any previous params into the struct.  To accommodate this caches are now cloned into the struct and then derived from there.  The caches at this point are watch receivers...I am still unsure whether these clones are pointers to the underlying value ( in this case a retainer cache ) or it its cloning the actual retainer cache.
- [ ] In the iot config I removed some internal shutdown listener handler ( such as in the gateway_service and the route_service ).  Maybe there was good reason they were there ?  If so I will need to rejig a few things to get the shutdown listener available to those services
- [x] the pricer's `price_tracker` task does not yet implement the new taskman api.  It has an unusual startup pattern which I will need to discuss with brian as to the original intent


task_manager/src/lib.rs
```diff
@@ -0,0 +1,280 @@
...
+
+fn create_triggers(n: usize) -> Vec<(triggered::Trigger, triggered::Listener)> {
+    let (shutdown_trigger, shutdown_listener) = triggered::trigger();
+    (0..n).fold(Vec::new(), |mut vec, _| {
+        vec.push((shutdown_trigger.clone(), shutdown_listener.clone()));
+        vec
+    })
+}
+
+// #[cfg(test)]
```
maplant, line 155: These tests should be removed or made working; preferably the latter

andymck:
superseded by #606

PR #577, opened 2023-07-24
rewards halvening

andymck:
** DO NOT MERGE ***

This updates the iot and mobile verifier to account for the rewards halvening to be applied in August

### DIFF
--- a/iot_verifier/src/reward_share.rs
+++ b/iot_verifier/src/reward_share.rs
@@ -18,7 +18,7 @@ const DEFAULT_PREC: u32 = 15;
 lazy_static! {
     // TODO: year 1 emissions allocate 30% of total to PoC with 6% to beacons and 24% to witnesses but subsequent years back
     // total PoC percentage off 1.5% each year; determine how beacons and witnesses will split the subsequent years' allocations
-    static ref REWARDS_PER_DAY: Decimal = (Decimal::from(32_500_000_000_u64) / Decimal::from(365)) * Decimal::from(1_000_000); //  89_041_095_890_410.5
+    static ref REWARDS_PER_DAY: Decimal = (Decimal::from(32_500_000_000_u64) / Decimal::from(366)) * Decimal::from(1_000_000); //  88_797_814_207_650.273224043715847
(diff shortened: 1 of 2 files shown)


jeffgrunewald (commented):
Would we not like to gate this feature by date comparison since these rewards are usually calculated in the context of a DateTime range “epoch” period and go ahead and merge this well ahead of the event?

bbalser (approved):

andymck:
> Would we not like to gate this feature by date comparison since these rewards are usually calculated in the context of a DateTime range “epoch” period and go ahead and merge this well ahead of the event?

I mean we could but it does add additional risk and will make it that bit more difficult to test and verify before going live.  I think merging on the day is likely the best scenario here.

The plan is to implement the reward curve into the code post this halving

maplant:
Would it make more sense to define some constant that is halving? So instead of having to manually halve all of the constants in the tests, we can just divide them by the constant?

iot_verifier/src/reward_share.rs
```diff
@@ -18,7 +18,7 @@ const DEFAULT_PREC: u32 = 15;
 lazy_static! {
     // TODO: year 1 emissions allocate 30% of total to PoC with 6% to beacons and 24% to witnesses but subsequent years back
     // total PoC percentage off 1.5% each year; determine how beacons and witnesses will split the subsequent years' allocations
-    static ref REWARDS_PER_DAY: Decimal = (Decimal::from(65_000_000_000_u64) / Decimal::from(365)) * Decimal::from(1_000_000); // 178_082_191_780_822
+    static ref REWARDS_PER_DAY: Decimal = (Decimal::from(32_500_000_000_u64) / Decimal::from(366)) * Decimal::from(1_000_000); //  88_797_814_207_650.273224043715847
```
maplant, line 21: Is the 366 because 2024 is a leap year? Are rewards not allocated per 365 days?

jeffgrunewald (approved):

iot_verifier/src/reward_share.rs
andymck: correct its due to 2024 being a leap year which is accounted for in the number of days

PR #578, opened 2023-07-25
Modeled Coverage - Phase 1

maplant:


mobile_verifier/src/heartbeats.rs
```diff
@@ -285,5 +439,26 @@ async fn validate_heartbeat(
         return Ok((cell_type, proto::HeartbeatValidity::GatewayOwnerNotFound));
     }
 
+    let Some(coverage_object) = heartbeat.report.coverage_object() else {
```
bbalser, line 442: Should we be invalidating heartbeats based on coverage objects in Phase 1??

mobile_verifier/src/coverage.rs
```diff
@@ -0,0 +1,725 @@
...
+        coverage_object: &Option<Uuid>,
+        exec: &mut Transaction<'_, Postgres>,
+    ) -> Result<DateTime<Utc>, sqlx::Error> {
+        let key = (cbsd_id.to_string(), *coverage_object);
+        if let Some(coverage_claim_time) = self.cache.get(&key).await {
+            Ok(*coverage_claim_time)
+        } else {
+            let coverage_claim_time: DateTime<Utc> = sqlx::query_scalar(
+                r#"
+                SELECT coverage_claim_time FROM hex_coverage WHERE cbsd_id = $1 AND uuid = $2
```
bbalser, line 465: should you have a LIMIT 1 on this query?

mobile_verifier/src/heartbeats.rs
```diff
@@ -212,18 +267,113 @@ impl Heartbeat {
...
-    ) -> Result<bool, SaveHeartbeatError> {
+    ) -> anyhow::Result<()> {
+        enum InsertOrUpdate {
+            Insert(proto::SeniorityUpdateReason),
+            Update(DateTime<Utc>),
+        }
+
+        let (seniority_ts, update_reason) = if let Some(prev_seniority) =
+            sqlx::query_as::<_, crate::coverage::Seniority>(
+                "SELECT * FROM seniority WHERE cbsd_id = $1 ORDER BY last_heartbeat DESC",
```
bbalser, line 296: Should this query have a LIMIT 1 on it?

mobile_verifier/src/cli/server.rs
```diff
@@ -82,11 +82,26 @@ impl Cmd {
...
+        // Seniority updates
+        let (seniority_updates, mut seniority_updates_server) = file_sink::FileSinkBuilder::new(
+            FileType::SeniorityUpdate,
+            store_base_path,
+            concat!(env!("CARGO_PKG_NAME"), "_seniority_update"),
+            shutdown_listener.clone(),
+        )
+        .deposits(Some(file_upload_tx.clone()))
+        .auto_commit(false)
+        .roll_time(Duration::minutes(15))
```
andymck, line 94: Is 15 mins a reasonable duration taking into account the volume ? Seems high but I have no idea re expected volumes

mobile_verifier/src/coverage.rs
```diff
@@ -0,0 +1,725 @@
...
+        }
+
+        self.file_sink.commit().await?;
+        transaction.commit().await?;
+
+        Ok(())
+    }
+}
+
+pub struct CoverageObject {
```
andymck, line 121: A CoverageObject is also defined in filestore ( without the validity field ).  Why repeat the whole thing here ? Instead why not something like:

pub struct CoverageObject {
object: Filestore::coverage:CoverageObject,
validity: CoverageObjectValidity,
}

I guess you have the second CoverageObject instance here to allow for easier `impl`s but duplicating their entire structure would not be an ideal pattern

I recently just noticed the same pattern in my speedtest PR and am looking to address that in a similar means to the above

mobile_verifier/src/heartbeats.rs
```diff
@@ -237,23 +387,25 @@ impl Heartbeat {
 
         let truncated_timestamp = self.truncated_timestamp()?;
         Ok(
-            sqlx::query_as::<_, HeartbeatSaveResult>(
+            sqlx::query_scalar(
                 r#"
-                INSERT INTO heartbeats (cbsd_id, hotspot_key, cell_type, latest_timestamp, truncated_timestamp)
-                VALUES ($1, $2, $3, $4, $5)
+                INSERT INTO heartbeats (cbsd_id, hotspot_key, cell_type, first_timestamp, latest_timestamp, truncated_timestamp, coverage_object)
+                VALUES ($1, $2, $3, $4, $5, $6)
```
andymck, line 393:
```suggestion
                VALUES ($1, $2, $3, $4, $5, $6, $7)
```

mobile_verifier/src/cli/server.rs
bbalser: the volume of seniority_updates should actually be relatively low, we should only output this when the seniority_ts changes, which can be caused by a new coverage_object or going 72 hours with no heartbeats.

mobile_verifier/src/coverage.rs
maplant: Good point. Addressed

mobile_verifier/src/coverage.rs
```diff
@@ -0,0 +1,725 @@
...
+            LIMIT 1
+            "#,
+        )
+        .bind(cbsd_id)
+        .bind(period_end)
+        .fetch_one(self)
+        .await?;
+
+        // We can safely delete any seniority objects that appear before the latest in the reward period
+        sqlx::query("DELETE FROM seniority WHERE inserted_at < $1")
```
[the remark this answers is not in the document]
andymck: yeah feels like it all needs wrapped in a txn.  And/Or would using foreign keys and delete cascade make this all a lot less verbose?

mobile_verifier/src/coverage.rs
```diff
@@ -0,0 +1,715 @@
...
+                    coverage_claim_time: seniority.seniority_ts,
+                    ..hc
+                })
+                .boxed(),
+        )
+    }
+}
+
+#[derive(Default)]
+#[allow(dead_code)]
```
andymck, line 359: why all the dead code, is this in prep for phase 3?

mobile_verifier/src/heartbeats.rs
```diff
@@ -158,45 +212,47 @@ pub struct Heartbeat {
     pub cell_type: Option<CellType>,
```
andymck, line 212: Similar to the comment on speedtests.rs, this is duplicating an existing filestore struct `CellHeartbeat` and adding a validity field

mobile_verifier/src/heartbeats.rs
```diff
@@ -158,45 +212,47 @@ pub struct Heartbeat {
...
     pub hotspot_key: PublicKeyBinary,
     pub timestamp: DateTime<Utc>,
+    pub coverage_object: Option<Uuid>,
+    pub lat: f64,
```
andymck, line 216: Werent we avoiding the use of floats, especially when it came to the protos ?  That said i see these were previously defined as double in the proto..

maplant: We might have been, but the proto has used floats for lat and lon in cellheartbeat message since the beginning

mobile_verifier/src/coverage.rs
maplant: yes

file_store/src/coverage.rs
```diff
@@ -0,0 +1,104 @@
...
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct RadioHexSignalLevel {
+    pub location: h3o::CellIndex,
+    pub signal_level: SignalLevel,
+    pub signal_power: i32,
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct CoverageObject {
+    pub pub_key: PublicKeyBinary,
```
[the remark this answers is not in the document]
maplant: I think this _is_ the pub key of the hotspot

mobile_verifier/migrations/15_modeled_coverage.sql
```diff
@@ -0,0 +1,28 @@
...
+);
+
+CREATE TABLE hex_coverage (
+       uuid UUID NOT NULL,
+       hex BIGINT NOT NULL,
+       indoor BOOLEAN NOT NULL,
+       cbsd_id TEXT NOT NULL,
+       signal_level signal_level NOT NULL,
+       coverage_claim_time TIMESTAMPTZ NOT NULL,
+       inserted_at TIMESTAMPTZ NOT NULL,
```
andymck, line 15:
```suggestion
       inserted_at TIMESTAMPTZ NOT NULL DEFAULT NOW(),
```

mobile_verifier/src/coverage.rs
```diff
@@ -0,0 +1,721 @@
...
+                  ($1, $2, $3, $4, $5, $6, $7)
+                "#,
+            )
+            .bind(self.coverage_obj.uuid)
+            .bind(location as i64)
+            .bind(self.coverage_obj.indoor)
+            .bind(&self.coverage_obj.cbsd_id)
+            .bind(SignalLevel::from(hex.signal_level))
+            .bind(self.coverage_obj.coverage_claim_time)
+            .bind(Utc::now())
```
andymck, line 190: As per suggestion in the migration, if `inserted_at` uses default now(), you can eliminate this binding

andymck (approved):

mobile_verifier/src/heartbeats.rs
maplant: Nope! Removed

mobile_verifier/src/heartbeats.rs
```diff
@@ -13,7 +13,7 @@ use futures::{
     stream::{Stream, StreamExt, TryStreamExt},
     TryFutureExt,
 };
-use h3o::LatLng;
+// use h3o::LatLng;
```
jeffgrunewald, line 16:
```suggestion
```
not needed or will this come into play in phase 3?

mobile_verifier/src/heartbeats.rs
```diff
@@ -459,6 +465,7 @@ async fn validate_heartbeat(
     if coverage.max_distance_km(latlng) > max_distance {
         return Ok((cell_type, proto::HeartbeatValidity::TooFarFromCoverage));
     }
+    */
```
jeffgrunewald, line 468: you said "removed" did you mean more specifically that we _should_ be invalidating these in phase 3 so this is only temporarily commented out or was this intended to be deleted once it successfully compiled or something?

jeffgrunewald (approved):
only questions left are about comments still in the code; assuming that's stuff intended to be uncommented in phase 3 please ignore/resolve

andymck (approved):

PR #580, opened 2023-07-26
Cache org client and remove solana cache invalidation for invalids

maplant:
This should dramatically speed up the iot packet verifier in the pathological case where we've disabled and org, but it continues to hammer the packet verifier. This is basically the worst case; insufficient funds means we must check balance sufficiency every time a packet comes in. We also cache the org disabling so that we don't send outbound requests to the org client every time this happens. 

TODO:
- [ ] Fix remaining integration test

### DIFF
--- a/iot_packet_verifier/src/balances.rs
+++ b/iot_packet_verifier/src/balances.rs
@@ -14,11 +14,11 @@ use tokio::sync::Mutex;
 /// Caches balances fetched from the solana chain and debits made by the
 /// packet verifier.
 pub struct BalanceCache<S> {
-    balances: BalanceStore,
+    payer_accounts: BalanceStore,
     solana: S,
 }
 
-pub type BalanceStore = Arc<Mutex<HashMap<PublicKeyBinary, Balance>>>;
+pub type BalanceStore = Arc<Mutex<HashMap<PublicKeyBinary, PayerAccount>>>;
 
 impl<S> BalanceCache<S>
 where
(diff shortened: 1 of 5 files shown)


jeffgrunewald (approved):

andymck (commented):
Is my understanding correct that should an org be disabled and subsequently cached it will remain disabled until whenever next tick of `monitor_funds` shows it is back in credit again and thus packets can potentially be declared invalid if the owner has credited the account but that hasnt yet been picked up by monitor funds ?

Also I see a bunch of unwraps in this code.   Are we dropping the hard rule of no unwraps?  In the past even if the code was infallible there were to be avoided

maplant:
> Is my understanding correct that should an org be disabled and subsequently cached it will remain disabled until whenever next tick of `monitor_funds` shows it is back in credit again and thus packets can potentially be declared invalid if the owner has credited the account but that hasnt yet been picked up by monitor funds ?
> 

Yes, exactly. Effectively this is to prevent an account from continuing to send invalid packets despite us disabling them. We'll increase how often we check for updated funds to compensate 

> 
> 
> Also I see a bunch of unwraps in this code.   Are we dropping the hard rule of no unwraps?  In the past even if the code was infallible there were to be avoided

I've always maintained that for hashmaps using unwraps is fine if you know the entry is populated. It's not worth throwing an error that can never occur

andymck (approved):

bbalser (approved):

PR #581, opened 2023-07-26
refactor speedtests handling, ensure rewards only include tests up until epoch end

andymck:
Refactor the speedtests to prevent scenario whereby the rewards can use reports from after the epoch end.  To accommodate this I had to explode the speedtests reports in the DB from 1 row containing an array of up to 6 reports out into 1 row per report.

At the same time i have split the averaging code out from the speedtest module into its own module for better readability and maintainability.

- [x] Add migration to explode speedtests out to 1 test per db row
- [x] delete speedtests after rewarding ( ensure we keep 48hrs worth of tests )
- [x] handle todo's
- [x] test locally
- [x] tidy up
- [x] add indexes to speedtest table

### DIFF
--- a/mobile_verifier/migrations/15_speedtests_one_to_one.sql
+++ b/mobile_verifier/migrations/15_speedtests_one_to_one.sql
@@ -4,16 +4,16 @@ CREATE TABLE speedtests_migration (
        upload_speed bigint,
        download_speed bigint,
        latency integer,
-       "serial" text,
+       serial_num text,
        timestamp timestamptz NOT NULL,
        inserted_at timestamptz default now(),
        PRIMARY KEY(pubkey, timestamp)
 );
-CREATE INDEX idx_speedtests2_pubkey on speedtests_migration (pubkey);
+CREATE INDEX idx_speedtests_pubkey on speedtests_migration (pubkey);
 
-INSERT INTO speedtests_migration (pubkey, upload_speed, download_speed, latency, serial, timestamp)
+INSERT INTO speedtests_migration (pubkey, upload_speed, download_speed, latency, serial_num, timestamp)
 SELECT id, (st).upload_speed, (st).download_speed, (st).latency, '', (st).timestamp
-FROM (select id, unnest(speedtests) as st from speedtests_old) as tmp
+FROM (select id, unnest(speedtests) as st from speedtests) as tmp
 ON CONFLICT DO NOTHING;
 
 ALTER TABLE speedtests RENAME TO speedtests_old;
--- a/mobile_verifier/migrations/15_speedtests_one_to_one.sql
+++ b/mobile_verifier/migrations/15_speedtests_one_to_one.sql
@@ -1,0 +1,22 @@
+
+CREATE TABLE speedtests_migration (
+       pubkey text NOT NULL,
+       upload_speed bigint,
+       download_speed bigint,
+       latency integer,
+       serial text,
+       timestamp timestamptz NOT NULL,
+       inserted_at timestamptz default now(),
+       PRIMARY KEY(pubkey, timestamp)
+);
+CREATE INDEX idx_speedtests_pubkey on speedtests_migration (pubkey);
+
+INSERT INTO speedtests_migration (pubkey, upload_speed, download_speed, latency, serial, timestamp)
+SELECT id, (st).upload_speed, (st).download_speed, (st).latency, '', (st).timestamp
+FROM (select id, unnest(speedtests) as st from speedtests) as tmp
+ON CONFLICT DO NOTHING;
+
+ALTER TABLE speedtests RENAME TO speedtests_old;
+ALTER TABLE speedtests_migration RENAME TO speedtests;
+
+
--- a/mobile_verifier/migrations/15_speedtests_one_to_one.sql
+++ b/mobile_verifier/migrations/15_speedtests_one_to_one.sql
@@ -4,16 +4,16 @@ CREATE TABLE speedtests_migration (
        upload_speed bigint,
        download_speed bigint,
        latency integer,
-       serial text,
+       "serial" text,
        timestamp timestamptz NOT NULL,
        inserted_at timestamptz default now(),
        PRIMARY KEY(pubkey, timestamp)
 );
-CREATE INDEX idx_speedtests_pubkey on speedtests_migration (pubkey);
+CREATE INDEX idx_speedtests2_pubkey on speedtests_migration (pubkey);
 
 INSERT INTO speedtests_migration (pubkey, upload_speed, download_speed, latency, serial, timestamp)
 SELECT id, (st).upload_speed, (st).download_speed, (st).latency, '', (st).timestamp
-FROM (select id, unnest(speedtests) as st from speedtests) as tmp
+FROM (select id, unnest(speedtests) as st from speedtests_old) as tmp
 ON CONFLICT DO NOTHING;
 
 ALTER TABLE speedtests RENAME TO speedtests_old;
--- a/mobile_verifier/src/heartbeats.rs
+++ b/mobile_verifier/src/heartbeats.rs
@@ -289,3 +289,14 @@ async fn validate_heartbeat(
 
     Ok((cell_type, proto::HeartbeatValidity::Valid))
 }
+
+pub async fn clear_heartbeats(
+    tx: &mut sqlx::Transaction<'_, sqlx::Postgres>,
+    timestamp: &DateTime<Utc>,
+) -> Result<(), sqlx::Error> {
+    sqlx::query("DELETE FROM heartbeats WHERE truncated_timestamp < $1")
+        .bind(timestamp)
+        .execute(&mut *tx)
+        .await?;
+    Ok(())
+}
--- a/mobile_verifier/src/reward_shares.rs
+++ b/mobile_verifier/src/reward_shares.rs
@@ -1,7 +1,6 @@
+use crate::speedtests_average::SpeedtestAverages;
 use crate::{
-    data_session::HotspotMap,
-    heartbeats::HeartbeatReward,
-    speedtests::{Average, SpeedtestAverages},
+    data_session::HotspotMap, heartbeats::HeartbeatReward, speedtests_average::SpeedtestAverage,
     subscriber_location::SubscriberValidatedLocations,
 };
 
@@ -221,15 +220,15 @@ pub struct PocShares {
 impl PocShares {
     pub async fn aggregate(
         heartbeats: impl Stream<Item = Result<HeartbeatReward, sqlx::Error>>,
-        speedtests: SpeedtestAverages,
+        speedtest_averages: &SpeedtestAverages,
     ) -> Result<Self, sqlx::Error> {
         let mut poc_shares = Self::default();
         let mut heartbeats = std::pin::pin!(heartbeats);
         while let Some(heartbeat) = heartbeats.next().await.transpose()? {
-            let speedmultiplier = speedtests
+            let speedmultiplier = speedtest_averages
                 .get_average(&heartbeat.hotspot_key)
                 .as_ref()
-                .map_or(Decimal::ZERO, Average::reward_multiplier);
+                .map_or(Decimal::ZERO, SpeedtestAverage::reward_multiplier);
             *poc_shares
                 .hotspot_shares
                 .entry(heartbeat.hotspot_key)
@@ -326,14 +325,16 @@ mod test {
         data_session,
         data_session::HotspotDataSession,
         heartbeats::HeartbeatReward,
-        speedtests::{Speedtest, SpeedtestAverages},
+        speedtests::Speedtest,
+        speedtests_average::{SpeedtestAverage, SpeedtestAverages},
         subscriber_location::SubscriberValidatedLocations,
     };
     use chrono::{Duration, Utc};
+    use file_store::speedtest::CellSpeedtest;
     use futures::stream;
     use helium_proto::services::poc_mobile::mobile_reward_share::Reward as MobileReward;
     use prost::Message;
-    use std::collections::{HashMap, VecDeque};
+    use std::collections::HashMap;
 
     fn valid_shares() -> RadioShares {
         let mut radio_shares: HashMap<String, Decimal> = Default::default();
@@ -526,7 +527,7 @@ mod test {
         assert_eq!(data_transfer_rewards.reward_scale().round_dp(1), dec!(0.5));
     }
 
-    fn bytes_per_s(mbps: i64) -> i64 {
+    fn bytes_per_s(mbps: u64) -> u64 {
         mbps * 125000
     }
 
@@ -536,39 +537,55 @@ mod test {
             .reward_weight()
     }
 
-    fn acceptable_speedtest(timestamp: DateTime<Utc>) -> Speedtest {
+    fn acceptable_speedtest(pubkey: PublicKeyBinary, timestamp: DateTime<Utc>) -> Speedtest {
         Speedtest {
-            timestamp,
-            upload_speed: bytes_per_s(10),
-            download_speed: bytes_per_s(100),
-            latency: 25,
+            report: CellSpeedtest {
+                pubkey,
+                timestamp,
+                upload_speed: bytes_per_s(10),
+                download_speed: bytes_per_s(100),
+                latency: 25,
+                serial: "".to_string(),
+            },
         }
     }
 
-    fn degraded_speedtest(timestamp: DateTime<Utc>) -> Speedtest {
+    fn degraded_speedtest(pubkey: PublicKeyBinary, timestamp: DateTime<Utc>) -> Speedtest {
         Speedtest {
-            timestamp,
-            upload_speed: bytes_per_s(5),
-            download_speed: bytes_per_s(60),
-            latency: 60,
+            report: CellSpeedtest {
+                pubkey,
+                timestamp,
+                upload_speed: bytes_per_s(5),
+                download_speed: bytes_per_s(60),
+                latency: 60,
+                serial: "".to_string(),
+            },
         }
     }
 
-    fn failed_speedtest(timestamp: DateTime<Utc>) -> Speedtest {
+    fn failed_speedtest(pubkey: PublicKeyBinary, timestamp: DateTime<Utc>) -> Speedtest {
         Speedtest {
-            timestamp,
-            upload_speed: bytes_per_s(1),
-            download_speed: bytes_per_s(20),
-            latency: 110,
+            report: CellSpeedtest {
+                pubkey,
+                timestamp,
+                upload_speed: bytes_per_s(1),
+                download_speed: bytes_per_s(20),
+                latency: 110,
+                serial: "".to_string(),
+            },
         }
     }
 
-    fn poor_speedtest(timestamp: DateTime<Utc>) -> Speedtest {
+    fn poor_speedtest(pubkey: PublicKeyBinary, timestamp: DateTime<Utc>) -> Speedtest {
         Speedtest {
-            timestamp,
-            upload_speed: bytes_per_s(2),
-            download_speed: bytes_per_s(40),
-            latency: 90,
+            report: CellSpeedtest {
+                pubkey,
+                timestamp,
+                upload_speed: bytes_per_s(2),
+                download_speed: bytes_per_s(40),
+                latency: 90,
+                serial: "".to_string(),
+            },
         }
     }
 
@@ -611,19 +628,21 @@ mod test {
 
         let last_timestamp = timestamp - Duration::hours(12);
         let g1_speedtests = vec![
-            acceptable_speedtest(last_timestamp),
-            acceptable_speedtest(timestamp),
+            acceptable_speedtest(g1.clone(), last_timestamp),
+            acceptable_speedtest(g1.clone(), timestamp),
         ];
         let g2_speedtests = vec![
-            acceptable_speedtest(last_timestamp),
-            acceptable_speedtest(timestamp),
+            acceptable_speedtest(g2.clone(), last_timestamp),
+            acceptable_speedtest(g2.clone(), timestamp),
         ];
-        let mut speedtests = HashMap::new();
-        speedtests.insert(g1.clone(), VecDeque::from(g1_speedtests));
-        speedtests.insert(g2.clone(), VecDeque::from(g2_speedtests));
-        let speedtest_avgs = SpeedtestAverages { speedtests };
+        let g1_average = SpeedtestAverage::from(&g1_speedtests);
+        let g2_average = SpeedtestAverage::from(&g2_speedtests);
+        let mut averages = HashMap::new();
+        averages.insert(g1.clone(), g1_average);
+        averages.insert(g2.clone(), g2_average);
+        let speedtest_avgs = SpeedtestAverages { averages };
 
-        let rewards = PocShares::aggregate(stream::iter(heartbeats).map(Ok), speedtest_avgs)
+        let rewards = PocShares::aggregate(stream::iter(heartbeats).map(Ok), &speedtest_avgs)
             .await
             .unwrap();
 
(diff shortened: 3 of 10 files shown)


mobile_verifier/migrations/15_speedtests_one_to_one.sql
```diff
@@ -0,0 +1,18 @@
...
+       latency integer,
+       timestamp timestamptz NOT NULL
+);
+CREATE INDEX idx_speedtests_pubkey on speedtests_migration (pubkey);
+
+INSERT INTO speedtests_migration (pubkey, upload_speed, download_speed, latency, timestamp)
+SELECT id, (st).upload_speed, (st).download_speed, (st).latency, (st).timestamp
+FROM (select id, unnest(speedtests) as st from speedtests) as tmp;
+
+ALTER TABLE speedtests RENAME TO speedtests_old;
```
andymck, line 15: I should change this to drop the old table but for safety sake renaming for now.  Will change to drop before merge

mobile_verifier/src/speedtests.rs
```diff
@@ -96,580 +87,117 @@ impl SpeedtestDaemon {
...
+    .bind(report.report.download_speed as i64)
+    .bind(report.report.latency as i64)
+    .bind(report.report.timestamp)
+    .execute(exec)
+    .await?;
+    Ok(())
+}
+
+// Clear the speedtests table of tests older than hours defined by SPEEDTEST_LAPSE
+pub async fn clear_speedtests(
```
maplant, line 195: This function should have the same signature as `clear_heartbeats`; you should clear out the speedtests that are before the reward period

mobile_verifier/src/speedtests.rs
```diff
@@ -96,580 +87,117 @@ impl SpeedtestDaemon {
...
-                    if gateway_client
-                        .resolve_gateway_info(&rolling_average.id)
-                        .await?
-                        .is_none()
-                    {
-                        return Ok(None);
-                    }
-                    Ok(Some((rolling_average, cell_speedtests)))
+            .map(anyhow::Ok)
+            .try_fold(transaction, |mut transaction, report| async {
```
maplant, line 103: I think that using a try fold, and separating out the functions is less clear and incongruous with how heartbeats (and eventually coverage object) daemons are structured, this function should match how the heartbeats function operates. That is, have a `validate_speedtests` function on `Speedtest` that you can call `write` and `save` on in a while loop.  

Basically, I think that the structure of the function before doesn't need to be changed, because it matches the way heartbeats work and how the coverage object work. Changing this function to look differently makes the function less clear

mobile_verifier/src/speedtests_average.rs
```diff
@@ -0,0 +1,463 @@
...
+            }
+        }
+    }
+}
+
+impl SpeedtestAverage {
+    pub async fn write(
+        &self,
+        filesink: &FileSinkClient,
+        speedtests: Vec<Speedtest>,
```
maplant, line 99: Instead of having speedtests be an argument, it makes more sense to me to include that in `SpeedtestAverage`

mobile_verifier/src/speedtests_average.rs
```diff
@@ -0,0 +1,463 @@
...
+}
+
+impl SpeedtestAverages {
+    pub fn get_average(&self, pub_key: &PublicKeyBinary) -> Option<SpeedtestAverage> {
+        self.averages.get(pub_key).cloned()
+    }
+
+    pub async fn aggregate_epoch_averages<'a>(
+        epoch_end: DateTime<Utc>,
+        exec: impl sqlx::PgExecutor<'a> + Copy + 'a,
```
maplant, line 226: This essentially force the executor to be a `PgPool`, so you might as well just make it that to clean up the argument

mobile_verifier/src/speedtests.rs
```diff
@@ -96,580 +87,117 @@ impl SpeedtestDaemon {
...
-                40,
-            ),
-        ];
-        let disjoint_speedtests =
-            speedtests_without_lapsed(disjoint_speedtests.iter(), speedtest_cutoff);
-
-        assert_eq!(contiguous_speedtests.count(), 8);
-        assert_eq!(disjoint_speedtests.count(), 1);
-    }
+pub async fn get_latest_speedtests_for_pubkey<'a>(
```
maplant, line 129:
```suggestion
pub async fn get_latest_speedtests_for_pubkey(
```

mobile_verifier/src/speedtests.rs
andymck: addressed here: a9b971c45caf62acb2c7b3c439ab824bad3ba01e

mobile_verifier/src/speedtests_average.rs
andymck: The averages are utilzed in multiple points in the code, but the only time the source speedtests from which the average was derived is required is at the point of writing out the proto.  Seems overkill to be throwing the vec of speedtests around unnecessarily

mobile_verifier/src/speedtests_average.rs
andymck: addressed here: a9b971c45caf62acb2c7b3c439ab824bad3ba01e
NOTE: heartbeats uses the same approach with the executor ( via HeartbeatReward::validated )

mobile_verifier/src/speedtests.rs
```diff
@@ -101,575 +87,122 @@ impl SpeedtestDaemon {
...
-                &self.gateway_client,
-                reports.map(|s| s.report),
-                &mut transaction,
-            )
-            .await?
-        );
-        while let Some(speedtest) = validated_speedtests.next().await.transpose()? {
-            speedtest.write(&self.file_sink).await?;
+        // todo, the double some is odd, what be the alternative ?
+        while let Some(Some(speedtest)) = validated_speedtests.next().await {
```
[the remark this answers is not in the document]
andymck: The gateway check was there prior to current changes ( https://github.com/helium/oracles/blob/main/mobile_verifier/src/speedtests.rs#L193 ).  Also the save in the changes *is* conditional, it only saves if it is validated

mobile_verifier/migrations/15_speedtests_one_to_one.sql
andymck: i am grand with that

mobile_verifier/src/speedtests.rs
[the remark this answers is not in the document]
andymck: i have refactored this again and kept it simple.  I was trying to shoehorn this into the prior pattern as you suggested ( and have a `validate_speedtests` iterator for example ) but it just becomes unnecessarily complicated and required boiler plate `impl` for the sake of them.  

Re adding a new proto for speedtest validity I am grand with that but suggest doing that as a separate PR.

mobile_verifier/src/heartbeats.rs
```diff
@@ -287,3 +287,14 @@ async fn validate_heartbeat(
 
     Ok((cell_type, proto::HeartbeatValidity::Valid))
 }
+
+pub async fn clear_heartbeats(
+    tx: &mut sqlx::Transaction<'_, sqlx::Postgres>,
+    reward_period: &Range<DateTime<Utc>>,
```
maplant, line 293: I would recommend not taking a period here since it is unnecessary, have this be `reward_start: DateTime<Utc>` (same for clear_speedtests)

andymck: addressed here: ae8fb0607f4b32f6881a7ce46e00551de941f410

bbalser (approved):

bbalser (approved):

mobile_verifier/migrations/15_speedtests_one_to_one.sql
```diff
@@ -0,0 +1,22 @@
+
```
jeffgrunewald, line 1:
```suggestion
```

mobile_verifier/migrations/15_speedtests_one_to_one.sql
```diff
@@ -0,0 +1,22 @@
...
+CREATE TABLE speedtests_migration (
+       pubkey text NOT NULL,
+       upload_speed bigint,
+       download_speed bigint,
+       latency integer,
+       serial text,
```
jeffgrunewald, line 7: Any issues with this column name being a keyword in Postgres? Should it be named `serial_num` or something?

mobile_verifier/migrations/15_speedtests_one_to_one.sql
```diff
@@ -0,0 +1,22 @@
...
+       timestamp timestamptz NOT NULL,
+       inserted_at timestamptz default now(),
+       PRIMARY KEY(pubkey, timestamp)
+);
+CREATE INDEX idx_speedtests_pubkey on speedtests_migration (pubkey);
+
+INSERT INTO speedtests_migration (pubkey, upload_speed, download_speed, latency, serial, timestamp)
+SELECT id, (st).upload_speed, (st).download_speed, (st).latency, '', (st).timestamp
+FROM (select id, unnest(speedtests) as st from speedtests) as tmp
```
jeffgrunewald, line 16: Is the second `as tmp` necessary since you don’t refer to a `tmp` table anywhere?

mobile_verifier/migrations/15_speedtests_one_to_one.sql
andymck: Never even noticed this, but all worked fine on local testing.   Still, prob best to rename and avoid

mobile_verifier/src/reward_shares.rs
```diff
@@ -1,7 +1,6 @@
+use crate::speedtests_average::SpeedtestAverages;
```
jeffgrunewald, line 1:
```suggestion
```

mobile_verifier/src/reward_shares.rs
```diff
@@ -1,7 +1,6 @@
...
 use crate::{
-    data_session::HotspotMap,
-    heartbeats::HeartbeatReward,
-    speedtests::{Average, SpeedtestAverages},
+    data_session::HotspotMap, heartbeats::HeartbeatReward, speedtests_average::SpeedtestAverage,
```
jeffgrunewald, line 3:
```suggestion
    data_session::HotspotMap, heartbeats::HeartbeatReward, speedtests_average::{SpeedtestAverage, SpeedtestAverages},
```

mobile_verifier/src/reward_shares.rs
```diff
@@ -326,14 +325,16 @@ mod test {
         data_session,
         data_session::HotspotDataSession,
         heartbeats::HeartbeatReward,
-        speedtests::{Speedtest, SpeedtestAverages},
+        speedtests::Speedtest,
+        speedtests_average::{SpeedtestAverage, SpeedtestAverages},
```
jeffgrunewald, line 329:
```suggestion
```
This line should be unnecessary since you `use super::*` to bring in those imports from the parent module already

maplant (approved):

mobile_verifier/src/rewarder.rs
```diff
@@ -214,15 +213,11 @@ impl Rewarder {
...
-        sqlx::query("DELETE FROM heartbeats WHERE truncated_timestamp < $1")
-            .bind(reward_period.start)
-            .execute(&mut transaction)
-            .await?;
-
-        // clear the db of data sessions data & subscriber location data for the epoch
-        data_session::clear_hotspot_data_sessions(&mut transaction, reward_period).await?;
-        // subscriber_location::clear_location_shares(&mut transaction, reward_period).await?;
+        // clear out the various db tables
+        heartbeats::clear_heartbeats(&mut transaction, &reward_period.start).await?;
```
jeffgrunewald, line 217: Why do we clear all the other records from the end of the reward period and before, but the heartbeats we clear from the _start_ of the period and prior? That would leave the heartbeats from the each reward period in the DB for the next reward period’s evaluation each day, no?

mobile_verifier/src/speedtests.rs
```diff
@@ -1,55 +1,41 @@
...
-            upload_speed,
-            download_speed,
-            latency,
-        }
+#[derive(Debug, Clone)]
+pub struct Speedtest {
+    pub report: CellSpeedtest,
+}
+
+impl FromRow<'_, PgRow> for Speedtest {
```
jeffgrunewald, line 27: Not sure if file_store has sqlx already defined but you should be able to derive the FromRow against the CellSpeedtest and then your queries to grab Speedtest structs just query_as CellSpeedtest and map the results to set that result as the value of the report field in the Speedtest struct. The only real data conversion I see here is the integer types and deriving FromRow allows you to annotate integer fields to `try_from` to convert from signed to unsigned integer types

mobile_verifier/src/speedtests.rs
```diff
@@ -99,577 +85,107 @@ impl SpeedtestDaemon {
...
-                &mut transaction,
-            )
-            .await?
-        );
-        while let Some(speedtest) = validated_speedtests.next().await.transpose()? {
-            speedtest.write(&self.file_sink).await?;
-            speedtest.save(&mut transaction).await?;
+        let mut speedtests = file.into_stream(&mut transaction).await?;
+        while let Some(speedtest_report) = speedtests.next().await {
+            let pubkey = speedtest_report.report.pubkey.clone();
```
jeffgrunewald, line 91: Do you need this clone here? Looks like you only pass it by reference for the rest of the operation

mobile_verifier/src/speedtests.rs
```diff
@@ -99,577 +85,107 @@ impl SpeedtestDaemon {
...
+    exec: &mut Transaction<'_, Postgres>,
+) -> Result<(), sqlx::Error> {
+    sqlx::query(
+        r#"
+        insert into speedtests (pubkey, upload_speed, download_speed, latency, serial, timestamp)
+        values ($1, $2, $3, $4, $5, $6)
+        on conflict (pubkey, timestamp) do nothing
+        "#,
+    )
+    .bind(speedtest.pubkey.clone())
```
jeffgrunewald, line 122: Do you need to clone the pubkeybin to insert the record? Same question with the text/string serial

mobile_verifier/src/speedtests.rs
```diff
@@ -99,577 +85,107 @@ impl SpeedtestDaemon {
...
+    .bind(speedtest.download_speed as i64)
+    .bind(speedtest.latency as i64)
+    .bind(speedtest.serial.clone())
+    .bind(speedtest.timestamp)
+    .execute(exec)
+    .await?;
+    Ok(())
+}
+
+pub async fn get_latest_speedtests_for_pubkey(
```
jeffgrunewald, line 133: Since you’re not performing any operations on the steamed result in the `while` loop you should be able to just make this a `fetch_all` instead of a `fetch` and get the results back in a vec without the while loop

mobile_verifier/src/speedtests.rs
```diff
@@ -99,577 +85,107 @@ impl SpeedtestDaemon {
...
+    }
+    Ok(speedtests)
+}
+
+pub async fn aggregate_epoch_speedtests<'a>(
+    epoch_end: DateTime<Utc>,
+    exec: &sqlx::Pool<sqlx::Postgres>,
+) -> Result<EpochSpeedTests, sqlx::Error> {
+    let mut speedtests = EpochSpeedTests::new();
+    // use latest speedtest which are no older than hours defined by SPEEDTEST_LAPSE
```
jeffgrunewald, line 157:
```suggestion
    // use latest speedtest which are no older than N hours, defined by SPEEDTEST_LAPSE
```

mobile_verifier/src/speedtests_average.rs
```diff
@@ -0,0 +1,478 @@
...
+        let mut id = vec![]; // eww!
+        let mut window_size = 0;
+        let mut sum_upload = 0;
+        let mut sum_download = 0;
+        let mut sum_latency = 0;
+
+        for Speedtest { report, .. } in
+            speedtests_without_lapsed(iter.into_iter(), Duration::hours(SPEEDTEST_LAPSE))
+        {
+            id = report.pubkey.as_ref().to_vec(); // eww!
```
jeffgrunewald, line 43: Instead of just overwriting the pubkey/id each iteration can you set it once based on the first element of the loop or else pass it in as an additional argument and then filter in the iterator loop based on each speedtest having the same pubkey?

mobile_verifier/src/speedtests_average.rs
```diff
@@ -0,0 +1,478 @@
...
+                    upload_speed_avg_bps: self.upload_speed_avg_bps,
+                    download_speed_avg_bps: self.download_speed_avg_bps,
+                    latency_avg_ms: self.latency_avg_ms,
+                    timestamp: Utc::now().encode_timestamp(),
+                    speedtests: speedtests_without_lapsed(
+                        speedtests.iter(),
+                        Duration::hours(SPEEDTEST_LAPSE),
+                    )
+                    .map(|st| proto::Speedtest {
+                        timestamp: st.report.timestamp.timestamp() as u64,
```
jeffgrunewald, line 109: This should probably implement the timestamp_millis feature so you don’t have to cast it to a u64, right?

mobile_verifier/src/speedtests_average.rs
```diff
@@ -0,0 +1,478 @@
...
+                        upload_speed_bps: st.report.upload_speed,
+                        download_speed_bps: st.report.download_speed,
+                        latency_ms: st.report.latency,
+                    })
+                    .collect(),
+                    validity: self.validity as i32,
+                    reward_multiplier: self.reward_multiplier.try_into().unwrap(),
```
jeffgrunewald, line 116: guaranteed safe unwrap?

mobile_verifier/src/speedtests_average.rs
```diff
@@ -0,0 +1,478 @@
...
+        Ok(())
+    }
+
+    pub fn reward_multiplier(&self) -> Decimal {
+        self.reward_multiplier
+    }
+
+    #[allow(dead_code)]
+    // function used by tests only
+    pub fn tier(&self) -> SpeedtestTier {
```
jeffgrunewald, line 130: Move into the test sub module and then remove the dead_code annotation? Or is this used in multiple testing modules?

mobile_verifier/src/speedtests_average.rs
```diff
@@ -0,0 +1,478 @@
+use crate::speedtests::{self, Speedtest};
+use chrono::{DateTime, Duration, Utc};
+use file_store::{file_sink::FileSinkClient, traits::TimestampEncode};
+use helium_crypto::PublicKeyBinary;
+use helium_proto::services::poc_mobile::{self as proto, SpeedtestAvgValidity};
```
jeffgrunewald, line 5:
```suggestion
use helium_proto::services::poc_mobile::proto;
```
Looks like you’re (typically) referencing the SpeedtestAvgValidity as `proto::SpeedtestAvgValidity` anyway so you can clean up this import a bit

mobile_verifier/src/speedtests_average.rs
andymck: copy pasted from the original speedtests module but changed now to use the timestamp trait.  Returns seconds tho not millis

mobile_verifier/src/speedtests_average.rs
andymck: I believe so and copy pasted from original, so always been there.

mobile_verifier/migrations/15_speedtests_one_to_one.sql
andymck: addressed here: 67effc3558d2cd2c56ef07956dba0b19b02ffd95

mobile_verifier/src/rewarder.rs
andymck: Both heartbeats and speedtests were meant to use the epoch start, not end.  Speedtests clearing now fixed:
addressed here: 67effc3558d2cd2c56ef07956dba0b19b02ffd95

mobile_verifier/src/speedtests.rs
andymck: addressed here: 67effc3558d2cd2c56ef07956dba0b19b02ffd95

mobile_verifier/src/speedtests.rs
andymck: both required

mobile_verifier/migrations/15_speedtests_one_to_one.sql
andymck: addressed here: 67effc3558d2cd2c56ef07956dba0b19b02ffd95

mobile_verifier/src/speedtests.rs
andymck: addressed here: 67effc3558d2cd2c56ef07956dba0b19b02ffd95

mobile_verifier/src/speedtests_average.rs
andymck: addressed here: 67effc3558d2cd2c56ef07956dba0b19b02ffd95

mobile_verifier/src/speedtests_average.rs
andymck: addressed here: 67effc3558d2cd2c56ef07956dba0b19b02ffd95

mobile_verifier/src/speedtests_average.rs
andymck: addressed here: 67effc3558d2cd2c56ef07956dba0b19b02ffd95

mobile_verifier/src/reward_shares.rs
andymck: addressed here: 67effc3558d2cd2c56ef07956dba0b19b02ffd95

mobile_verifier/src/speedtests.rs
jeffgrunewald: you can just pass by reference as `&speedtest_report.report.pubkey;`

mobile_verifier/src/speedtests_average.rs
jeffgrunewald: here's what i was thinking:
```
impl<'a, I: ?Sized> From<&'a I> for SpeedtestAverage
where
    &'a I: IntoIterator<Item = &'a Speedtest>,
{
    fn from(iter: &'a I) -> Self {
        let mut window_size = 0;
        let mut sum_upload = 0;
        let mut sum_download = 0;
        let mut sum_latency = 0;

        let mut speedtest_iter = speedtests_without_lapsed(iter.into_iter(), Duration::hours(SPEEDTEST_LAPSE)).peekable();
        let id = speedtest_iter
            .peek()
            .map(|Speedtest { report, .. }| report.pubkey.as_ref().to_vec()).unwrap_or_default();
        for Speedtest { report, .. } in speedtest_iter 
        {
            if report.pubkey.as_ref().to_vec() == id {
                sum_upload += report.upload_speed;
                sum_download += report.download_speed;
                sum_latency += report.latency;
                window_size += 1;
            }
        }

        if window_size > 0 {
            let upload_speed_avg_bps = sum_upload / window_size as u64;
            let download_speed_avg_bps = sum_download / window_size as u64;
            let latency_avg_ms = sum_latency / window_size as u32;
            let validity = validity(
                window_size as usize,
                upload_speed_avg_bps,
                download_speed_avg_bps,
                latency_avg_ms,
            );
            let tier = SpeedtestTier::new(
                window_size as usize,
                upload_speed_avg_bps,
                download_speed_avg_bps,
                latency_avg_ms,
            );
            let reward_multiplier = tier.into_multiplier();
            SpeedtestAverage {
                pubkey: id.into(),
                window_size: window_size as usize,
                upload_speed_avg_bps,
                download_speed_avg_bps,
                latency_avg_ms,
                validity,
                reward_multiplier,
            }
        } else {
            SpeedtestAverage {
                pubkey: id.into(),
                window_size: 0,
                upload_speed_avg_bps: sum_upload,
                download_speed_avg_bps: sum_download,
                latency_avg_ms: sum_latency,
                validity: proto::SpeedtestAvgValidity::TooFewSamples,
                reward_multiplier: Decimal::ZERO,
            }
        }
    }
}
```
logically this works and it compiles. i'm not a fan of the `unwrap_or_default` since that _theoretically_ could result in aggregating the speedtest average under a `SpeedtestAverage` with an empty Vec<u8> as the `Pubkey` but since that seems really unlikely that we'd have a valid `CellSpeedtest` with an empty pubkey than you should be safe to just replace this with an `unwrap()`.

i like the `peekable` implementation on the iterator to grab the first element of the list without consuming it and then checking subsequent members of the list to ensure the pubkeys match before including them in the average, but that's probably unnecessary and if overwriting the id on each iteration of the for loop over the collection of CellSpeedtests isn't an expensive operation this is probably unnecessary so i'll leave it to you to take or ignore

jeffgrunewald (approved):

PR #583, opened 2023-07-29
Add stream permission for hpr listing skfs or euis

michaeldjeffrey:
Allows a sanity check from routing infrastructure to re-request information for a specific route without requiring reloading the entire database of routing information.

There can be times where the updating of a routes skfs or euis can be weird. 
An EUI/SKF can be added and removed in an order where HPRs end up with the record, but it no longer exists in the database. Subsequent requests to remove the record from the Config Service will "succeed", but not send any updates to HPRs because no update was made to the database.

Adding HPR permission to `list_skfs` and `get_euis` allows them to do a sanity check of a single route.

### DIFF
--- a/iot_config/src/route_service.rs
+++ b/iot_config/src/route_service.rs
@@ -122,6 +122,21 @@ impl RouteService {
         }
     }
 
+    async fn verify_request_signature_or_stream<'a, R>(
+        &self,
+        signer: &PublicKey,
+        request: &R,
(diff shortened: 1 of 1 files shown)


jeffgrunewald (approved):
seems reasonable to me; only question: should this change apply to the `get_devaddrs()` request as well?

bbalser (approved):

PR #585, opened 2023-08-01
Add deduping to mobile packet verifier

bbalser:


mobile_packet_verifier/src/event_ids.rs
```diff
@@ -0,0 +1,73 @@
...
+    received_timestamp: DateTime<Utc>,
+) -> anyhow::Result<()> {
+    sqlx::query("INSERT INTO event_ids(event_id, received_timestamp) VALUES($1,$2)")
+        .bind(event_id)
+        .bind(received_timestamp)
+        .execute(tx)
+        .await
+        .map(|_| ())
+        .map_err(anyhow::Error::from)
+}
```
jeffgrunewald, line 31: worth the speed boost to make this an upsert statement returning whether or not the insert happened and then do it in a single query?

jeffgrunewald: i've done this in a different app which attempts to do an upsert and then returns the boolean if the row was inserted or not
```
    query_with_bindings.execute(&mut tx).await?.rows_affected() > 0;
```

mobile_packet_verifier/src/event_ids.rs
```diff
@@ -0,0 +1,73 @@
...
+
+    Ok(result)
+}
+
+pub async fn record(
+    tx: &mut Transaction<'_, Postgres>,
+    event_id: String,
+    received_timestamp: DateTime<Utc>,
+) -> anyhow::Result<()> {
+    sqlx::query("INSERT INTO event_ids(event_id, received_timestamp) VALUES($1,$2)")
```
andymck, line 24:
```suggestion
    sqlx::query("INSERT INTO event_ids(event_id, received_timestamp) VALUES ($1, $2)")
```

andymck (approved):

mobile_packet_verifier/src/event_ids.rs
maplant: I agree with @jeffgrunewald here, this should be an upsert

mobile_packet_verifier/migrations/4_event_ids.sql
```diff
@@ -0,0 +1,4 @@
+CREATE TABLE event_ids (
+    event_id TEXT NOT NULL PRIMARY KEY,
```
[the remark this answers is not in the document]
bbalser: There are unique according to Dmytro.

mobile_packet_verifier/src/event_ids.rs
maplant: This is would the query would look like:
```rust
let is_duplicate = sqlx::query(
    r#"
    INSERT INTO event_ids (event_id, received_timestamp) VALUES ($1, $2) 
    ON CONFLICT DO NOTHING
    "#
)
.bind(event_id)
.bind(received_timestamp)
.execute(tx)
.await?
.rows_affected() > 0;
```

bbalser: Agreed, upsert is better 5c66469

mobile_packet_verifier/src/event_ids.rs
bbalser: Fixed 5c66469

jeffgrunewald (approved):

PR #587, opened 2023-08-04
Make file_store usable outside of oracles

jeffgrunewald:
Convert the internal FileInfo struct from file-store to have a generic "prefix" instead of an internally defined enum FileType to allow external apps/libraries that use file-store as a dependency to build FileInfos for their own types (when using for searching the S3 records for instance).

This is intended as a minimal refactor; a longer-term solution would be to implement a FileInfo trait that encapsulates the functionality in question and better exposes some common functionality used in the file store cli commands for extension (such as "listing" a collection of files into a stream, filtered by an input type and decoded)

### DIFF
--- a/file_store/src/file_store.rs
+++ b/file_store/src/file_store.rs
@@ -1,5 +1,5 @@
 use crate::{
-    error::DecodeError, BytesMutStream, Error, FileInfo, FileInfoStream, FileType, Result, Settings,
+    error::DecodeError, BytesMutStream, Error, FileInfo, FileInfoStream, Result, Settings,
 };
 use aws_config::meta::region::RegionProviderChain;
 use aws_sdk_s3::{types::ByteStream, Client, Endpoint, Region};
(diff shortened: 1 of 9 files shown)


file_store/src/file_store.rs
```diff
@@ -57,27 +57,27 @@ impl FileStore {
...
+    pub async fn list_all<A, B, P>(
         &self,
-        file_type: F,
+        file_type: P,
         after: A,
         before: B,
     ) -> Result<Vec<FileInfo>>
     where
-        F: Into<FileType> + Copy,
+        P: ToString + std::fmt::Display,
```
maplant, line 67: I believe this Display bound is unnecessary.

Generally, I would not suggest having things impl ToString, and instead just have the function take a `&str`

andymck (approved):

bbalser (approved):

file_store/src/file_store.rs
jeffgrunewald: addressed; i thought we had steered toward a slightly more permissive `impl ToString` and `impl AsRef<str>` but maybe i'm thinking of something else

PR #592, opened 2023-08-10
add filestore new support

andymck:
Support non settings based creation of a filestore. Specifically added to allow the subscriber location service to instantiate a filestore without using a settings file

### DIFF
--- a/file_store/src/file_store.rs
+++ b/file_store/src/file_store.rs
@@ -1,5 +1,7 @@
 use crate::{
-    error::DecodeError, BytesMutStream, Error, FileInfo, FileInfoStream, Result, Settings,
+    error::DecodeError,
+    settings::{self, Settings},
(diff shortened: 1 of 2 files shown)


bbalser (approved):

file_store/src/file_store.rs
```diff
@@ -57,6 +59,36 @@ impl FileStore {
...
+                .map(Endpoint::immutable)
+                .map(Some)
+                .map_err(DecodeError::from)?,
+            _ => None,
+        };
+        let region = if region.is_some() {
+            Region::new(region.unwrap())
+        } else {
+            Region::new(settings::default_region())
+        };
```
jeffgrunewald, line 78:
```suggestion
        let region = Region::new(region.unwrap_or_else(|| settings::default_region()));
```

jeffgrunewald (approved):
one minor cleanup opportunity on setting the region i see otherwise good!

PR #597, opened 2023-08-11
Don't allow constructing a RouteId with invalid uuid

michaeldjeffrey:
I'm not sure if this is the right way to go about this since the result is being thrown away and reparsed later.

A few days ago, a User was trying to list a routes Euis and being met with an `unauthorized request signature` error, when it was a route_id typo.

Making sure a user provides a valid Uuid as early as possible can provide better guidance towards using the system. Normally the route id is not attempted to be parsed until the db is about to be hit where other concerns are more pertinent. This leaves Uuid parsing errors mostly to be overwritten as general authentication errors.

### DIFF
--- a/iot_config/src/route_service.rs
+++ b/iot_config/src/route_service.rs
@@ -47,6 +47,15 @@ enum OrgId<'a> {
     RouteId(&'a str),
 }
 
+impl<'a> OrgId<'a> {
+    fn route_id(route_id: &'a str) -> Result<Self, Status> {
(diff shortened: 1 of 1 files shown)


jeffgrunewald:
i think this is definitely a reasonable thing to expose to a user to improve their ability to do error handling but i'd lean more toward the alternative implementation i put up in https://github.com/helium/oracles/pull/598 which just reuses the existing error instead of swallowing the details in the rpc handler's `map_err()`

jeffgrunewald:
Closing in favor of https://github.com/helium/oracles/pull/598

PR #600, opened 2023-08-14
iot packet verifier ignores join packets

jeffgrunewald:
Make lora join packets free for the iot network by having the iot packet verifier ignore them when determining if packets have been paid for. Join packets are skipped over when reviewing packet reports so they are not burned against the sending OUI's wallet and are likewise not marked as "invalid"


bbalser (approved):

andymck (approved):
lgtm, only suggestion I would add is that a test would be useful here

PR #602, opened 2023-08-15
Add integration test for mobile verifier

maplant:


mobile_verifier/tests/test_poc_shares.rs
```diff
@@ -0,0 +1,158 @@
...
+        timestamp,
+        upload_speed: bytes_per_s(10),
+        download_speed: bytes_per_s(100),
+        latency: 25,
+        // Unused:
+        serial: String::new(),
+    }
+}
+
+async fn rewards(
```
andymck, line 85: Whilst this function does some of what the actual rewarder does and tests key functions of that flow, an integration test should be using the actual rewarder flow itself rather than shadowing it.  Its fine to have prep functions to generate the data etc but i think this rewards function negates a lot of the point of these being integration tests

PR #605, opened 2023-08-18
Check the chain for transactions if there are attempted burns that haven't been completed

maplant:


iot_packet_verifier/migrations/4_saved_balance.sql
```diff
@@ -0,0 +1,4 @@
+CREATE TABLE saved_balances (
+        payer TEXT PRIMARY KEY,
+        balance BIGINT NOT NULL
```
andymck, line 3: Be good to have a last_update or insert_date here

iot_packet_verifier/src/pending_burns.rs
```diff
@@ -216,3 +332,22 @@ pub struct Burn {
     pub payer: PublicKeyBinary,
     pub amount: i64,
 }
+
+#[derive(Clone, FromRow, Debug)]
+pub struct BurnAttempt {
+    pub payer: PublicKeyBinary,
+    amount: i64,
+    latest_transaction_signature: String,
```
andymck, line 340: insert date would be useful here, even if only to assist manual auditing, verification when browsing the table

solana/src/lib.rs
```diff
@@ -227,6 +248,73 @@ impl SolanaNetwork for SolanaRpc {
 
         Ok(())
     }
+
+    async fn latest_transaction(&self) -> Result<Signature, Self::Error> {
+        let signer = Keypair::from_bytes(&self.keypair).unwrap();
```
andymck, line 253: this unwrap safe?

iot_packet_verifier/migrations/4_saved_balance.sql
```diff
@@ -0,0 +1,4 @@
+CREATE TABLE saved_balances (
```
andymck, line 1: is this table redundant now?

iot_packet_verifier/src/burner.rs
```diff
@@ -58,6 +61,42 @@ where
...
+            tracing::info!(%since, %amount, "Checking chain for burn transaction");
+            if self
+                .solana
+                .has_burn_transaction(&attempted_burn.payer, amount, &since)
+                .await
+                .map_err(BurnError::SolanaError)?
+            {
+                tracing::info!("Found a matching transaction. Removing burn attempt");
+                self.pending_burns
+                    .complete_burn_attempt(&attempted_burn.payer, amount)
```
andymck, line 84: On the regular burn function, the payer balance is updated after a complete_burn_attempt is successful. Does the same need performed here after this run of complete_burn_attempt?

solana/src/lib.rs
maplant: Yup, we do this in the burn function as well

iot_packet_verifier/migrations/4_saved_balance.sql
maplant: Yup, forgot to delete

Cargo.toml
```diff
@@ -65,6 +65,7 @@ hextree = "*"
 solana-client = "1.14"
 solana-sdk = "1.14"
 solana-program = "1.11"
+solana-transaction-status = "1.14"
```
jeffgrunewald, line 68: friendly periodic reminder that the solana family of dependencies is in the 1.16.x range now. we'll have to struggle through an update eventually

iot_packet_verifier/src/pending_burns.rs
```diff
@@ -48,25 +59,62 @@ impl PendingBurns for Pool<Postgres> {
             .await
     }
 
-    async fn subtract_burned_amount(
+    async fn fetch_incomplete_burns(&mut self) -> Result<Vec<BurnAttempt>, Self::Error> {
+        sqlx::query_as("SELECT * FROM attempted_burns")
```
andymck, line 63: you forget to checkin the migration for this table ?

andymck (approved):
Other than the missing attempted_burns table migration, lgtm

PR #606, opened 2023-08-18
Add taskman support to IOT packet verifier

andymck:
Adds a task manager to control the start and stop order of processes within an oracle

Refactors the filestore and DB libs to enable oracles to use either the new taskmanager flow or to continue to use the existing non taskmanager flow.  This requires minor changes to the filestore and DB apis to support initialisation with or without taskmanager.

The taskmanager is then implemented for the IOT packet verifier.

All other oracles continue to use existing (the non taskmanager) startup and shutdown flows.


bbalser (approved):

maplant (approved):

file_store/src/file_info_poller.rs
```diff
@@ -54,51 +56,86 @@ pub struct FileInfoPoller<T> {
...
+    config: FileInfoPollerConfig<T>,
+    sender: Sender<FileInfoStream<T>>,
+}
+
+impl<T> FileInfoPollerConfigBuilder<T>
+where
+    T: Clone,
+{
+    pub fn create(self) -> Result<(Receiver<FileInfoStream<T>>, FileInfoPollerServer<T>)> {
+        let config = self.build().unwrap(); // TODO: fix the unwrap
```
jeffgrunewald, line 70: does this TODO need fixed before the initial release or are we cool merging this now and finishing later?

file_store/src/file_sink.rs
```diff
@@ -63,28 +67,37 @@ pub struct FileSinkBuilder {
...
+            max_size: 50_000_000,
+            roll_time: Duration::minutes(DEFAULT_SINK_ROLL_MINS),
+            deposits: None,
+            file_upload: None,
+            auto_commit: true,
+            metric,
+        }
+    }
+
+    pub fn new_tm(prefix: impl ToString, target_path: &Path, metric: &'static str) -> Self {
```
jeffgrunewald, line 90: `new` and `new_tm` seem to be identical; are they both needed?

file_store/src/file_upload.rs
```diff
@@ -18,20 +19,67 @@ pub async fn upload_file(tx: &MessageSender, file: &Path) -> Result {
...
+    }
+}
+
+impl FileUploadServer {
     pub async fn run(self, shutdown: &triggered::Listener) -> Result {
-        tracing::info!("starting file uploader 1");
+        self.run_tm(shutdown.clone()).await
+    }
+
+    pub async fn run_tm(self, shutdown: triggered::Listener) -> Result {
```
jeffgrunewald, line 81: this also looks superfluous; this seems like you could just remove `run` and rename `run_tm` to `run`

iot_config/src/main.rs
```diff
@@ -136,11 +130,7 @@ impl Daemon {
             .serve_with_shutdown(listen_addr, shutdown_listener)
             .map_err(Error::from);
 
-        tokio::try_join!(
-            db_join_handle.map_err(Error::from),
-            md_pool_handle.map_err(Error::from),
-            server
-        )?;
+        tokio::try_join!(server)?;
```
jeffgrunewald, line 133: nothing to join here anymore; i think you can just await this

file_store/src/file_info_poller.rs
andymck: fixed here: 0fab396f15e3ef14da478b36dec63ae7f32441f1

file_store/src/file_sink.rs
andymck: Fixed, there were different until the a few commits ago.  Forgot to update these, addressed here 3f8c94b66b069cdb497c986892a97a6fe909d47d

file_store/src/file_upload.rs
andymck: Fixed, there were different until the a few commits ago.  Forgot to update these, addressed here 3f8c94b66b069cdb497c986892a97a6fe909d47d

iot_config/src/main.rs
andymck: fixed here 683b9d51710e605db82d0a2dc358c69d4380c5c3

mobile_config/src/main.rs
```diff
@@ -112,11 +106,7 @@ impl Daemon {
             .serve_with_shutdown(listen_addr, shutdown_listener)
             .map_err(Error::from);
 
-        tokio::try_join!(
-            pool_handle.map_err(Error::from),
-            md_pool_handle.map_err(Error::from),
-            server,
-        )?;
+        tokio::try_join!(server)?;
```
jeffgrunewald, line 109: same deal as iot_config; you can just await the server builder task instead of the try_join

task_manager/src/lib.rs
```diff
@@ -0,0 +1,279 @@
...
+    pub fn builder() -> TaskManagerBuilder {
+        TaskManagerBuilder { tasks: Vec::new() }
+    }
+
+    pub fn add(&mut self, task: impl ManagedTask + 'static) {
+        self.tasks.push(Box::new(task));
+    }
+
+    pub async fn start(self) -> anyhow::Result<()> {
+        let mut sigterm = signal::unix::signal(signal::unix::SignalKind::terminate()).unwrap();
```
jeffgrunewald, line 73: avoid the unwrap here since we're inside something that should be able to take `?` since it returns an anyhow result?

task_manager/src/select_all.rs
```diff
@@ -0,0 +1,57 @@
+use core::mem;
+use core::pin::Pin;
```
jeffgrunewald, line 2: do these differ from `std::mem` and `std::pin::Pin`?

mobile_config/src/main.rs
andymck: i see you fixed, ty

jeffgrunewald (approved):

PR #607, opened 2023-08-21
retry iot_config client rpcs

jeffgrunewald:


iot_config/src/client/mod.rs
```diff
@@ -35,6 +35,34 @@ pub struct Client {
...
+        let mut attempt = 1;
+        loop {
+            let response = $rpc.await;
+            match response {
+                Ok(resp) => break Ok(resp),
+                Err(status) => match status.code() {
+                    Code::Cancelled | Code::DeadlineExceeded | Code::Unavailable => {
+                        if attempt < 3 {
+                            attempt += 1;
+                            tokio::time::sleep(Duration::from_millis(attempt * 100)).await;
```
andymck, line 51: delay between attempts is very short.  Doesnt give much time for any transitory issue to resolve

bbalser: @andymck What do you think it should be?

andymck: 5 secs ?

jeffgrunewald: We default the rpc timeout in the client endpoint builder to 5 sec already so it should probably be less then that since we’re making three calls but I can skip the increasing back off  and just retry 3 times at the default. That reminds me though we’d talked about modifying the request to set the timeout header which is going to shoot this whole thing down

bbalser: Should we start with 1 second, so then do 1,2,3??

bbalser (approved):

iot_config/src/client/mod.rs
andymck: works for me

andymck (approved):

maplant (approved):

PR #610, opened 2023-08-22
drop hardcoded sign key, replace with support for N keys via settings

andymck:
Linked ops PR: https://github.com/novalabsxyz/ops/pull/775

Replaces a single hardcoded sign key used to verify the signature of denylist updates with N keys defined via settings

### DIFF
--- a/denylist/src/denylist.rs
+++ b/denylist/src/denylist.rs
@@ -125,12 +125,12 @@ pub fn filter_from_bin(bin: &Vec<u8>, sign_keys: &[PublicKey]) -> Result<Xor32> 
     sign_keys
         .iter()
         .any(|key| {
-            if key.verify(buf, &signature).is_ok() {
-                tracing::info!(pubkey = %key, "valid denylist signer");
-                true
-            } else {
-                false
-            }
+            key.verify(buf, &signature)
+                .map(|res| {
+                    tracing::info!(pubkey = %key, "valid denylist signer");
+                    res
+                })
+                .is_ok()
         })
         .then(|| {
             let _serial = buf.get_u32_le();
--- a/denylist/src/denylist.rs
+++ b/denylist/src/denylist.rs
@@ -123,21 +123,18 @@ pub fn filter_from_bin(bin: &Vec<u8>, valid_sign_keys: &Vec<PublicKey>) -> Resul
     let _version = buf.get_u8();
     let signature_len = buf.get_u16_le() as usize;
     let signature = buf.copy_to_bytes(signature_len).to_vec();
-    for pubkey in valid_sign_keys {
-        match pubkey.verify(buf, &signature) {
-            Ok(_) => {
-                tracing::info!(%pubkey, "updating filter to latest");
-                let _serial = buf.get_u32_le();
-                let xor = bincode::deserialize::<Xor32>(buf)?;
-                return Ok(xor);
-            }
-            Err(_) => continue,
-        }
-    }
-    tracing::warn!("filter signature verification failed");
-    Err(Error::InvalidBinary(
-        "filter signature verification failed".to_string(),
-    ))
+    valid_sign_keys
+        .iter()
+        .any(|key| key.verify(buf, &signature).is_ok())
+        .then(|| {
+            let _serial = buf.get_u32_le();
+            bincode::deserialize::<Xor32>(buf)
+        })
+        .transpose()?
+        .ok_or_else(|| {
+            tracing::warn!("filter signature verification failed");
+            Error::InvalidBinary("filter signature verification failed".to_string())
+        })
 }
 
 fn public_key_hash<R: AsRef<[u8]>>(public_key: R) -> u64 {
--- a/denylist/src/denylist.rs
+++ b/denylist/src/denylist.rs
@@ -19,7 +19,7 @@ pub struct DenyList {
     pub client: DenyListClient,
     #[serde(skip_serializing)]
     pub filter: Xor32,
-    pub valid_sign_keys: Vec<PublicKey>,
+    pub sign_keys: Vec<PublicKey>,
 }
 
 impl TryFrom<Vec<PublicKeyBinary>> for DenyList {
@@ -32,7 +32,7 @@ impl TryFrom<Vec<PublicKeyBinary>> for DenyList {
             tag_name: 0,
             client,
             filter,
-            valid_sign_keys: vec![],
+            sign_keys: vec![],
         })
     }
 }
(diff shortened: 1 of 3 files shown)


denylist/src/settings.rs
```diff
@@ -15,9 +15,10 @@ pub struct Settings {
     /// Cadence at which we poll for an updated denylist (secs)
     #[serde(default = "default_trigger_interval")]
     pub trigger: u64,
-    // comma seperated string of b58 encoded pubkeys
+    // vec of b58 encoded pubkeys
```
madninja, line 18: specifically b58 helium encoded pub keys!

andymck: updated

denylist/src/denylist.rs
```diff
@@ -122,21 +123,21 @@ pub fn filter_from_bin(bin: &Vec<u8>) -> Result<Xor32> {
...
-        }
-        Err(_) => {
-            tracing::warn!("filter signature verification failed");
-            Err(Error::InvalidBinary(
-                "filter signature verification failed".to_string(),
-            ))
+    for pubkey in valid_sign_keys {
+        match pubkey.verify(buf, &signature) {
+            Ok(_) => {
+                tracing::info!("updating filter to latest");
```
madninja, line 129: add `%pubkey`  to indicate which public key verified this?

denylist/src/settings.rs
```diff
@@ -36,7 +40,7 @@ impl Settings {
     /// Environemnt overrides have the same name as the entries in the settings
     /// file in uppercase and prefixed with "DENYLIST_". For example
     /// "DENYLIST_LOG" will override the log setting.
-    pub fn new<P: AsRef<Path>>(path: Option<P>) -> Result<Self> {
+    pub fn new<P: AsRef<Path>>(path: Option<P>) -> Result<Self, config::ConfigError> {
```
madninja, line 43: Why did this change? Seems unneeded? If you want to include this change you should also remove config::ConfigError from this crate's  Error definition?

denylist/src/denylist.rs
```diff
@@ -35,12 +32,13 @@ impl TryFrom<Vec<PublicKeyBinary>> for DenyList {
...
             client,
             filter,
+            valid_sign_keys: vec![],
         })
     }
 }
 
 impl DenyList {
-    pub fn new() -> Result<Self> {
+    pub fn new(settings: Settings) -> Result<Self> {
```
madninja, line 41: Take `&Settings` so you don't have to clone settings when using this

iot_verifier/src/runner.rs
```diff
@@ -73,7 +73,7 @@ impl Runner {
         let beacon_max_retries = settings.beacon_max_retries;
         let witness_max_retries = settings.witness_max_retries;
         let deny_list_latest_url = settings.denylist.denylist_url.clone();
-        let mut deny_list = DenyList::new()?;
+        let mut deny_list = DenyList::new(settings.denylist.clone())?;
```
madninja, line 76: use `&settings`

denylist/src/denylist.rs
```diff
@@ -23,6 +19,7 @@ pub struct DenyList {
     pub client: DenyListClient,
     #[serde(skip_serializing)]
     pub filter: Xor32,
+    pub valid_sign_keys: Vec<PublicKey>,
```
jeffgrunewald, line 22: any significance to `valid_` in `valid_sign_keys`? seems like any configured signing key would be expected to be valid

denylist/src/denylist.rs
```diff
@@ -122,21 +123,21 @@ pub fn filter_from_bin(bin: &Vec<u8>) -> Result<Xor32> {
...
-            ))
+    for pubkey in valid_sign_keys {
+        match pubkey.verify(buf, &signature) {
+            Ok(_) => {
+                tracing::info!("updating filter to latest");
+                let _serial = buf.get_u32_le();
+                let xor = bincode::deserialize::<Xor32>(buf)?;
+                return Ok(xor);
+            }
+            Err(_) => continue,
```
jeffgrunewald, line 134: sign_keys
    .iter()
    .any(|key| key.verify(buf, &signature).is_ok())
    .then(|| {
        let _serial = buf.get_u32_le();
        bincode::deserialize::<Xor32>(buf)
    })
    .transpose()?
    .ok_or_else(|| {
        tracing::warn!("filter signature verification failed");
        Err(Error::InvalidBinary("filter signature verification_failed".to_string()))
    })

maplant: ^ this is much better IMO

denylist/src/denylist.rs
```diff
@@ -122,21 +123,21 @@ pub fn filter_from_bin(bin: &Vec<u8>) -> Result<Xor32> {
...
-            tracing::warn!("filter signature verification failed");
-            Err(Error::InvalidBinary(
-                "filter signature verification failed".to_string(),
-            ))
+    for pubkey in valid_sign_keys {
+        match pubkey.verify(buf, &signature) {
+            Ok(_) => {
+                tracing::info!("updating filter to latest");
+                let _serial = buf.get_u32_le();
+                let xor = bincode::deserialize::<Xor32>(buf)?;
```
maplant, line 131: Since buf is just a slice, you can create a subslice and I think that's better than manually moving the pointer forward:
```rust
buf = buf[4..];
```

denylist/src/denylist.rs
bbalser: Addressed dbf8dd7

denylist/src/settings.rs
bbalser: Reverted in dbf8dd7

denylist/src/denylist.rs
bbalser: Fixed dbf8dd7

iot_verifier/src/runner.rs
bbalser: Fixed dbf8dd7

denylist/src/denylist.rs
jeffgrunewald: you mean instead of the `buf.get_u32_le()`?

madninja: huh? you _want_ to move the cursor forward?

denylist/src/denylist.rs
```diff
@@ -123,21 +123,25 @@ pub fn filter_from_bin(bin: &Vec<u8>, valid_sign_keys: &Vec<PublicKey>) -> Resul
...
-            Ok(_) => {
-                tracing::info!("updating filter to latest");
-                let _serial = buf.get_u32_le();
-                let xor = bincode::deserialize::<Xor32>(buf)?;
-                return Ok(xor);
+    valid_sign_keys
+        .iter()
+        .any(|key| {
+            if key.verify(buf, &signature).is_ok() {
+                tracing::info!(pubkey = %key, "valid denylist signer");
```
madninja, line 130: Just name it pubkey as the any argument to avoid the extra noise here?

denylist/src/denylist.rs
maplant: > huh? you _want_ to move the cursor forward? which get_u32_le would do no?

I mean manually as in "through a contrived function that doesn't explain what it's doing"

denylist/src/denylist.rs
bbalser: Renamed 0b9db6e

denylist/src/denylist.rs
madninja: it's documented in get_u32_le and part of Buf behavior? But if we wanted to make that absolutely clear here and since we're not using the result, you could just call `advance(4)` instead?

bbalser: I like that idea the best, so I made the change 923a641

jeffgrunewald (approved):

PR #611, opened 2023-08-22
broadcast route changes in spawn task

jeffgrunewald:
This change is to further guard against heavy load on the iot config service preventing route updates from successfully broadcasting those updates out on the HPR stream or from returning a success result to the calling client.

When the service is under load and the broadcast channel it uses to send updates to the connected HPRs is saturated, the route update process will go into a backoff loop until the channel has capacity. While previously done in-line to the grpc handler receiving the update request, this backoff _could_ at times of high load wait long enough for the grpc handler to timeout, either on the client or the server, which causes the request to fail even when the change was successfully persisted to the database.

### DIFF
--- a/iot_config/src/lib.rs
+++ b/iot_config/src/lib.rs
@@ -52,7 +52,7 @@ pub async fn broadcast_update<T: std::fmt::Debug>(
 fn enqueue_update(queue_size: usize) -> bool {
     // enqueue the message for broadcast if
     // the current queue is <= 80% full
-    (queue_size * 100) / BROADCAST_CHANNEL_QUEUE <= 80
+    (queue_size * 100) / BROADCAST_CHANNEL_QUEUE <= 85
 }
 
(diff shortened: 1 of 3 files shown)


iot_config/src/lib.rs
```diff
@@ -52,7 +52,7 @@ pub async fn broadcast_update<T: std::fmt::Debug>(
 fn enqueue_update(queue_size: usize) -> bool {
     // enqueue the message for broadcast if
     // the current queue is <= 80% full
-    (queue_size * 100) / BROADCAST_CHANNEL_QUEUE <= 80
+    (queue_size * 100) / BROADCAST_CHANNEL_QUEUE <= 85
```
bbalser, line 55: why the change to 85% for the capacity check?

bbalser (approved):

andymck (approved):

PR #612, opened 2023-08-22
Switch to new denylist key, see helium/denylist#8356

Vagabond:


jeffgrunewald:
addressed by https://github.com/helium/oracles/pull/610 which pulls potential multi-sig public keys from configuration

PR #615, opened 2023-08-23
taskman up iot config

andymck:
Applies taskmanager control the the mobile config oracle

Removed some internal shutdown listeners on some of the services.  Impact of these removals needs considered further and if required then will need to implement and alternative pattern

### DIFF
--- a/iot_config/Cargo.toml
+++ b/iot_config/Cargo.toml
@@ -35,8 +35,11 @@ sqlx = {workspace = true}
 thiserror = {workspace = true}
 tokio = {workspace = true}
 tokio-stream = {workspace = true}
+tokio-util = { workspace = true }
 tonic = {workspace = true}
 tower-http = {workspace = true}
 tracing = {workspace = true}
 tracing-subscriber = {workspace = true}
 triggered = {workspace = true}
+task-manager = { path = "../task_manager" }
+
--- a/iot_config/Cargo.toml
+++ b/iot_config/Cargo.toml
@@ -42,4 +42,3 @@ tracing = {workspace = true}
 tracing-subscriber = {workspace = true}
 triggered = {workspace = true}
 task-manager = { path = "../task_manager" }
-
--- a/iot_config/src/route_service.rs
+++ b/iot_config/src/route_service.rs
@@ -389,35 +389,23 @@ impl iot_config::Route for RouteService {
         let mut route_updates = self.subscribe_to_routes();
 
         tokio::spawn(async move {
-            tokio::select! {
-                result = stream_existing_routes(&pool, &signing_key, tx.clone())
-                             .and_then(|_| stream_existing_euis(&pool, &signing_key, tx.clone()))
-                             .and_then(|_| stream_existing_devaddrs(&pool, &signing_key, tx.clone()))
-                             .and_then(|_| stream_existing_skfs(&pool, &signing_key, tx.clone())) => {
-                    if let Err(error) = result {
-                        tracing::error!(?error, "Error occurred streaming current routing configuration");
-                        return;
-                    }
-                }
+            if stream_existing_routes(&pool, &signing_key, tx.clone())
+                .and_then(|_| stream_existing_euis(&pool, &signing_key, tx.clone()))
+                .and_then(|_| stream_existing_devaddrs(&pool, &signing_key, tx.clone()))
+                .and_then(|_| stream_existing_skfs(&pool, &signing_key, tx.clone()))
+                .await
+                .is_err()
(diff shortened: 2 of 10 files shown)


iot_config/Cargo.toml
```diff
@@ -35,8 +35,11 @@ sqlx = {workspace = true}
...
 tokio = {workspace = true}
 tokio-stream = {workspace = true}
+tokio-util = { workspace = true }
 tonic = {workspace = true}
 tower-http = {workspace = true}
 tracing = {workspace = true}
 tracing-subscriber = {workspace = true}
 triggered = {workspace = true}
+task-manager = { path = "../task_manager" }
+
```
jeffgrunewald, line 45:
```suggestion
```

andymck:
@jeffgrunewald if you can pay particular attention to the removal of the shutdown listeners embedded within some of the services....

iot_config/src/route_service.rs
```diff
@@ -390,15 +383,13 @@ impl iot_config::Route for RouteService {
...
         let pool = self.pool.clone();
-        let shutdown_listener = self.shutdown.clone();
         let (tx, rx) = tokio::sync::mpsc::channel(20);
         let signing_key = self.signing_key.clone();
 
         let mut route_updates = self.subscribe_to_routes();
 
         tokio::spawn(async move {
             tokio::select! {
-                _ = shutdown_listener.clone() => return,
```
jeffgrunewald, line 401: this select statement can just go away if we're not selecting between two futures and the stream of combinators can just be awaited with a map_err at the end. that's what this used to do before i tried selecting on the shutdown to attempt to exit more gracefully

iot_config/src/route_service.rs
```diff
@@ -413,13 +404,7 @@ impl iot_config::Route for RouteService {
             tracing::info!("existing routes sent; streaming updates as available");
             telemetry::route_stream_subscribe();
             loop {
-                let shutdown = shutdown_listener.clone();
```
jeffgrunewald, line 416: this can also avoid the select macro and probably just have a `while let Ok(update) = stream.recv() {}` inside the loop

jeffgrunewald (commented):
there are several instances where i was passing a clone of the shutdown listener into a short-lived stream where i was assured it wasn't necessary and those _didn't_ prove to solve the shutdown issues we have with the config service so i'm totally fine with those going. i'm still unclear about the long-lived streams spawned in the `route_service.rs` module that represent the grpc route updates subscription an HPR has to the config service but i believe @maplant was pretty sure those are also safe to stop monitoring and the tokio runtime would take care of their shutdown signals so if he can confirm that i'm okay simplifying that as well.

iot_config/src/route_service.rs
andymck: addressed here: ec80d34b0873a8a45fb277efa7f5b64b8cb28bf8

iot_config/src/route_service.rs
andymck: addressed here: ec80d34b0873a8a45fb277efa7f5b64b8cb28bf8

andymck:
@jeffgrunewald I addressed your comments but if you can give another pass over those changes.  I dont have enough context on that route service to feel comfortable with what is there right now

iot_config/src/route_service.rs
```diff
@@ -390,49 +383,29 @@ impl iot_config::Route for RouteService {
...
-                        tracing::error!(?error, "Error occurred streaming current routing configuration");
-                        return;
-                    }
-                }
+            if stream_existing_routes(&pool, &signing_key, tx.clone())
+                .and_then(|_| stream_existing_euis(&pool, &signing_key, tx.clone()))
+                .and_then(|_| stream_existing_devaddrs(&pool, &signing_key, tx.clone()))
+                .and_then(|_| stream_existing_skfs(&pool, &signing_key, tx.clone()))
+                .await
+                .is_err()
```
jeffgrunewald, line 397: why not just call this combinator chain, map_err at the end where we keep the error trace and then `await?`?

jeffgrunewald (commented):
unfortunately @bbalser 's testing with the new streaming ingestor rpcs proved we _do_ need shutdown listeners inside the spawned tasks that handle indefinite streaming RPCs so we'll need to keep the ones in each of the `router_service.rs` methods that take a `GrpcStreamRequest` or return a `GrpcStreamResult` type in order to shut down when OS sends the signal

bbalser (approved):

jeffgrunewald (approved):

PR #617, opened 2023-08-25
Remove heartbeat delete non-determinism

maplant:


Cargo.toml
```diff
@@ -22,6 +22,7 @@ members = [
     "solana",
     "task_manager",
 ]
+resolver = "2"
```
jeffgrunewald: looks like this needs a rebase

mobile_verifier/src/heartbeats.rs
```diff
@@ -226,17 +254,6 @@ impl Heartbeat {
...
-        // If the heartbeat is not valid, do not save it
-        if self.validity != proto::HeartbeatValidity::Valid {
-            return Ok(false);
-        }
-
-        sqlx::query("DELETE FROM heartbeats WHERE cbsd_id = $1 AND hotspot_key != $2")
-            .bind(&self.cbsd_id)
-            .bind(&self.hotspot_key)
-            .execute(&mut *exec)
-            .await?;
```
jeffgrunewald, line 238: is this safe now b/c the query above that pulls valid heartbeats is going to ignore records if the cbsd_id isn't matched to the "current/latest" hotspot_key (so we avoid counting radio heartbeats across multiple gateways?

mobile_verifier/tests/heartbeats.rs
```diff
@@ -0,0 +1,68 @@
...
+        .try_collect()
+        .await?;
+
+    assert_eq!(
+        heartbeat_reward,
+        vec![HeartbeatReward {
+            hotspot_key: hotspot_2,
+            cbsd_id,
+            reward_weight: Decimal::ONE
+        }]
```
jeffgrunewald, line 64: these timestamps all fall in the same day so we should probably also have an assertion that hortspot_1 _doesnt_ get any `HeartbeatReward`

task_manager/src/lib.rs
```diff
@@ -135,6 +135,7 @@ fn start_futures(
 }
 
 async fn stop_all(futures: Vec<StopableLocalFuture>) -> anyhow::Result<()> {
+    #[allow(clippy::manual_try_fold)]
```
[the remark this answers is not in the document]
maplant: It could be, but I just wanted to satisfy clippy

jeffgrunewald (approved):

Cargo.toml
maplant: alright resolved

mobile_verifier/src/heartbeats.rs
maplant: That's not quite right - the current query fetches the latest hotspot key for a cbsd_id and joins that with the previous query, so the table will only return the latest hotspot key

mobile_verifier/tests/heartbeats.rs
maplant: It doesn't. I'm asserting that there are no rewards

jeffgrunewald: i missed the `vec!` so this probably does properly assert that only the one hotspot was rewarded but perhaps modifying it or adding additional scenarios that validate what we discussed with the business such as having a radio produce 11 heartbeats attached to hotspot_1 and 1 heartbeat attached to hotspot_2 after that and asserting once again that hotspot_2 and only hotspot_2 received the reward?

task_manager/src/lib.rs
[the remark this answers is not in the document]
jeffgrunewald:
```suggestion
```
rebasing fixed this for the moment; we addressed this for real in your taskmanager PR for the iot-verifier @andymck

mobile_verifier/tests/heartbeats.rs
bbalser: maybe add a test case where there are not 12 heartbeats for any hotspot but more than 12 when it counts heartbeats for both hotspots??

jeffgrunewald (approved):

PR #618, opened 2023-08-26
fix witness metadata on invalid reports

jeffgrunewald:
At the stage at the end of `do_witness_verifications` in the iot_verifier poc.rs module where we assign non-zero values to invalid witness metadata we should have validated the witness's on-chain metadata at least (the invalid witness report was invalid for a different reason) so we should assign the witnesses's metadata to its section of the invalid report rather than the beaconer's.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,6 +22,7 @@ members = [
     "solana",
     "task_manager",
 ]
+resolver = "2"
 
 [workspace.package]
 authors = ["Nova Labs <info@nova-labs.com>"]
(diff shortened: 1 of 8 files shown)


Cargo.toml
```diff
@@ -22,6 +22,7 @@ members = [
     "solana",
     "task_manager",
 ]
+resolver = "2"
```
madninja, line 25: the changes in this file seem unneeded

jeffgrunewald: The latest rust (1.72.0) has started throwing some noisy warnings/notes on compile without this

bbalser (approved):

Cargo.toml
madninja: how incredibly annoying that they just randomly started requiring explicitly using a next gen resolver.. oh well

jeffgrunewald: They didn’t, just required you explicitly declare one. If this should “1” that’s cool too

andymck (approved):
ffs, couldnt believe this one when i saw it

PR #620, opened 2023-08-28
Modeled Coverage Phase 3 

maplant:


mobile_verifier/src/coverage.rs
```diff
@@ -402,47 +456,72 @@ impl CoveredHexes {
...
+            radios
+                .into_sorted_vec()
+                .into_iter()
+                .take(MAX_OUTDOOR_RADIOS_PER_HEX)
+                .enumerate()
+                .map(|(rank, cl)| CoverageReward {
+                    points: cl.coverage_points() * OUTDOOR_REWARD_WEIGHTS[rank],
+                    hotspot: cl.hotspot,
+                    radio_key: cl.radio_key,
+                })
```
bbalser, line 504: I think it is easier to understand if you use `zip`

```rust
                .zip(OUTDOOR_REWARD_WEIGHTS)
                .map(|(cl, weight)| CoverageReward {
                    points: cl.coverage_points() * weight,
                    hotspot: cl.hotspot,
                    radio_key: cl.radio_key,
                })
```

mobile_verifier/src/heartbeats/mod.rs
```diff
@@ -107,6 +114,37 @@ pub enum OwnedKeyType {
...
+impl From<PublicKeyBinary> for OwnedKeyType {
+    fn from(w: PublicKeyBinary) -> Self {
+        Self::Wifi(w)
+    }
+}
+
+impl PartialEq<KeyType<'_>> for OwnedKeyType {
+    fn eq(&self, rhs: &KeyType<'_>) -> bool {
+        match (self, rhs) {
+            (Self::Cbrs(lhs), KeyType::Cbrs(rhs)) => lhs == rhs,
```
jeffgrunewald, line 141: why do you need to deref the wifi keytype but not the cbrs keytype? b/c rust already implements PartialEq between Strings and &strs but `helium-crypto-rs` doesn't do the same between `PublicKeyBinary` and `&'a PublicKeyBinary`?

mobile_verifier/src/reward_shares.rs
```diff
@@ -203,52 +204,147 @@ pub fn dc_to_mobile_bones(dc_amount: Decimal, mobile_bone_price: Decimal) -> Dec
...
-    ) -> anyhow::Result<Self> {
-        let mut poc_shares = Self::default();
-        let mut heartbeat_rewards = std::pin::pin!(heartbeat_rewards);
-        while let Some(heartbeat_reward) = heartbeat_rewards.next().await.transpose()? {
-            let speedmultiplier = speedtest_averages
-                .get_average(&heartbeat_reward.hotspot_key)
+impl HotspotPoints {
+    pub fn new(speedtest_multiplier: Decimal) -> Self {
+        Self {
+            speedtest_multiplier,
```
jeffgrunewald, line 241: maybe return a result here and check that the passed-in multiplier is greater than Decimal::ZERO?

mobile_verifier/src/reward_shares.rs
```diff
@@ -203,52 +204,147 @@ pub fn dc_to_mobile_bones(dc_amount: Decimal, mobile_bone_price: Decimal) -> Dec
...
-        self.radio_shares
-            .values()
-            .fold(Decimal::ZERO, |sum, amount| sum + amount)
+impl RadioPoints {
+    fn new(heartbeat_multiplier: Decimal, coverage_object: Uuid, seniority: DateTime<Utc>) -> Self {
+        Self {
+            heartbeat_multiplier,
+            seniority,
+            coverage_object,
+            points: Decimal::ZERO,
```
jeffgrunewald, line 221: we instantiate this to ZERO but where does this value ever get set to anything else?

mobile_verifier/src/reward_shares.rs
```diff
@@ -203,52 +204,147 @@ pub fn dc_to_mobile_bones(dc_amount: Decimal, mobile_bone_price: Decimal) -> Dec
...
+
+        Ok(Self { coverage_points })
+    }
+
+    /*
+    pub fn is_valid(&self, hotspot: &PublicKeyBinary) -> bool {
+        if let Some(coverage_points) = self.coverage_points.get(hotspot) {
+            !coverage_points.total_points().is_zero()
+        } else {
+            false
```
jeffgrunewald, line 330: delete?

mobile_verifier/src/reward_shares.rs
```diff
@@ -203,52 +204,147 @@ pub fn dc_to_mobile_bones(dc_amount: Decimal, mobile_bone_price: Decimal) -> Dec
...
         }
-        Ok(poc_shares)
+    }
+     */
+
+    /// Only used for testing
```
jeffgrunewald, line 335:
```suggestion
#[cfg(test)]
```
?

jeffgrunewald (commented):
Am I understanding correctly that the split between indoor and outdoor covered hexes is intended to be generic across both CBRS and Wifi hotspots, since CBRS gateways can have multiple radios, both indoor and outdoor (hypothetically) and thus might have results in both hashmaps while a wifi hotspot should only have a single set of either indoor or outdoor hashmap entries in any nearby covered hexes since they map 1-to-1 hotspot key/gateway-to-radio?

mobile_verifier/src/reward_shares.rs
```diff
@@ -318,22 +457,24 @@ mod test {
...
-    use futures::stream;
+    use futures::stream::{self, BoxStream};
     use helium_proto::services::poc_mobile::mobile_reward_share::Reward as MobileReward;
     use prost::Message;
     use std::collections::HashMap;
     use uuid::Uuid;
 
     #[test]
-    fn bytes_to_bones() {
+    fn ensure_correct_conversion_of_bytes_to_bones() {
```
andymck, line 477: weird naming choice, do we ever use an incorrect conversion?

mobile_verifier/src/reward_shares.rs
```diff
@@ -203,52 +204,147 @@ pub fn dc_to_mobile_bones(dc_amount: Decimal, mobile_bone_price: Decimal) -> Dec
...
+    fn new(heartbeat_multiplier: Decimal, coverage_object: Uuid, seniority: DateTime<Utc>) -> Self {
+        Self {
+            heartbeat_multiplier,
+            seniority,
+            coverage_object,
+            points: Decimal::ZERO,
+        }
+    }
+
+    fn points(&self) -> Decimal {
```
andymck, line 225: it is a lil confusing have the var points in the struct and then this points function.  Difficult to distinguish in their usage and could be error prone where one is used instead of the other.  Maybe rename points function to points_multiplier() or other ?

jeffgrunewald: i think it makes sense in this case since the `points` field in the struct is private, so this method is the only way to access that value

mobile_verifier/src/reward_shares.rs
```diff
@@ -203,52 +204,147 @@ pub fn dc_to_mobile_bones(dc_amount: Decimal, mobile_bone_price: Decimal) -> Dec
...
+        } in covered_hexes.into_coverage_rewards()
+        {
+            // Guaranteed that points contains the given hotspot.
+            coverage_points
+                .get_mut(&hotspot)
+                .unwrap()
+                .radio_points
+                .get_mut(&radio_key.into_cbsd_id())
+                .unwrap()
+                .points += points;
```
jeffgrunewald, line 319: i'm not clear how this works with the points field not being public

mobile_verifier/src/reward_shares.rs
maplant: The ensure? This just properly describes what the test does

mobile_verifier/src/heartbeats/mod.rs
```diff
@@ -236,11 +274,18 @@ pub struct HeartbeatReward {
     // cell hb only
     pub cbsd_id: Option<String>,
     pub reward_weight: Decimal,
```
bbalser, line 276: Can we rename this to location_trust_score_multiplier?

mobile_verifier/src/reward_shares.rs
maplant: Every field is public to its own module

mobile_verifier/src/heartbeats/mod.rs
```diff
@@ -383,7 +382,7 @@ impl ValidatedHeartbeat {
                 proto::Heartbeat {
                     cbsd_id: self.heartbeat.cbsd_id.clone().unwrap_or_default(),
                     pub_key: self.heartbeat.hotspot_key.as_ref().into(),
-                    reward_multiplier: self.cell_type.reward_weight().to_f32().unwrap_or_default(),
+                    reward_multiplier: 1.0,
```
jeffgrunewald, line 385: can this be deprecated and then just ignored? if not, please comment "this is a legacy field that's been superseded by algorithmic coverage weights" or something

jeffgrunewald (commented):
I recognize it went in already from a previous PR but I would really like `KeyType` and `OwnedKeyType` to get renamed to something that makes them distinct from the public key "key type" that PublicKeyBinary might actually refer to (RSA, Ed25519, EccCompact, etc). Something more like "Address" or "Identifier" which I think more clearly illustrates that the type is used to find a printable (or easily converted to printable) identifier

mobile_verifier/src/reward_shares.rs
maplant: can't, used in integration tests

mobile_verifier/src/coverage.rs
```diff
@@ -239,48 +239,86 @@ pub struct HexCoverage {
...
+            .cmp(&other.signal_power)
+            .reverse()
+            .then_with(|| self.coverage_claim_time.cmp(&other.coverage_claim_time))
+    }
+}
+
+impl OutdoorCoverageLevel {
+    fn coverage_points(&self) -> Decimal {
+        match self.signal_level {
+            SignalLevel::High => dec!(16),
```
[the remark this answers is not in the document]
maplant: Yeah, these are taken straight from the HIP

mobile_verifier/src/coverage.rs
```diff
@@ -239,48 +239,86 @@ pub struct HexCoverage {
...
+        }
+    }
+}
+
+#[derive(Eq, Debug)]
+struct OutdoorCoverageLevel {
+    radio_key: OwnedKeyType,
+    coverage_claim_time: DateTime<Utc>,
+    hotspot: PublicKeyBinary,
+    signal_power: i32,
```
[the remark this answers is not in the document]
maplant: it's only used in calculating coverage points by outdoor hotspots, for now

mobile_verifier/src/coverage.rs
```diff
@@ -239,48 +239,86 @@ pub struct HexCoverage {
...
-            (false, SignalLevel::High) => dec!(16),
-            (false, SignalLevel::Medium) => dec!(8),
-            (false, SignalLevel::Low) => dec!(4),
-            (_, SignalLevel::None) => dec!(0),
-            _ => anyhow::bail!("Indoor radio cannot have a signal level of medium"),
-        })
+impl IndoorCoverageLevel {
+    fn coverage_points(&self) -> Decimal {
+        match self.signal_level {
+            SignalLevel::High => dec!(400),
```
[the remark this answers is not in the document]
maplant: correct

mobile_verifier/src/coverage.rs
```diff
@@ -291,16 +329,23 @@ pub struct CoverageReward {
     pub hotspot: PublicKeyBinary,
 }
 
-pub const MAX_RADIOS_PER_HEX: usize = 5;
+pub const MAX_INDOOR_RADIOS_PER_HEX: usize = 5;
+pub const MAX_OUTDOOR_RADIOS_PER_HEX: usize = 3;
```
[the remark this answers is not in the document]
maplant: yes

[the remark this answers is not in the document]
jeffgrunewald: can we call that out for reference, either as a comment or in the name of the const?
`MAX_OUTDOOR_RADIOS_PER_RES12_HEX` or similar comment

mobile_verifier/src/heartbeats/mod.rs
maplant: yeah it's an impl thing

mobile_verifier/src/reward_shares.rs
jeffgrunewald: was not aware the `/test` directory tests weren't compiled for the test environment. that'd just make too much sense

mobile_verifier/src/coverage.rs
```diff
@@ -402,47 +456,72 @@ impl CoveredHexes {
...
-                .push(CoverageLevel {
-                    radio_key,
-                    coverage_claim_time,
-                    indoor,
-                    signal_level,
-                    hotspot: hotspot.clone(),
-                });
+            if indoor {
+                self.indoor
+                    .entry(CellIndex::try_from(hex as u64).unwrap())
```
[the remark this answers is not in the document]
maplant: I don't know about h3o vs heron, but we instantiate the cell index using the same method before putting it in the DB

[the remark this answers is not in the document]
jeffgrunewald: i guess the question then wouldn't really apply here where you're reading out of the DB but rather at the insert point; if the u64 being turned into a CellIndex there isn't a valid H3 cell h3o will perform that validation and fail, which would cause the loading process to crash whereas working with heron previously the library would allow you to instantiate invalid cells

mobile_verifier/src/heartbeats/mod.rs
maplant: We are planning on deprecating this field at some point

mobile_verifier/src/reward_shares.rs
maplant: I don't think that's particularly useful? It wouldn't be very relevant to this PR either

bbalser (approved):

jeffgrunewald (approved):

andymck (approved):

mobile_verifier/src/reward_shares.rs
```diff
@@ -1142,21 +1190,21 @@ mod test {
...
         // the emissions actually distributed for the hour
         let mut distributed_total_rewards = 0;
         for val in owner_rewards.values() {
             distributed_total_rewards += *val
         }
-        assert_eq!(distributed_total_rewards, 2_049_180_327_863);
+        assert_eq!(distributed_total_rewards, 2_049_180_327_865);
 
-        let diff = expected_total_rewards - distributed_total_rewards;
+        let diff = expected_total_rewards as i128 - distributed_total_rewards as i128;
```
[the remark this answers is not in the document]
maplant: the expected could be greater or less, and these are u64s not i64s

PR #625, opened 2023-09-04
file_store (helium-crypto) compiling issue

kurotych:
Fixes the same compiling issue https://github.com/helium/helium-crypto-rs/issues/63 with rsa 0.4.0 
But now it happens when I use `file_store` as a dependency

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -59,7 +59,7 @@ sqlx = {version = "0", features = [
   "macros",
   "runtime-tokio-rustls"
 ]}
-helium-crypto = {version = "0.8.0", features=["sqlx-postgres", "multisig"]}
+helium-crypto = {version = "0.8.0", default-features=false, features=["sqlx-postgres", "multisig"]}
 helium-proto = {git = "https://github.com/helium/proto", branch = "master", features = ["services"]}
 hextree = "*"
 solana-client = "1.14"


jeffgrunewald:
Unfortunately in the case of the Oracles we upgraded to the helium-crypto-rs to version 0.8.0 and left default features enabled explicitly to include the new RSA feature b/c the mobile network will be onboarding devices with RSA keys in the next month and this new feature was needed to begin testing.

I don't see any reason why we couldn't disable default features and then include RSA in the features list to explicitly document that the feature is in use in Oracles but reading the referenced issue that doesn't seem like it will solve your problem.

kurotych:
Hello @jeffgrunewald.
Is the RSA feature needed for the `file-store` crate? 

If not, could you enable the RSA feature only in creates (workspaces) that will use it?

kurotych:
Ok, I see, Thank you for your response.  
(I deleted the previous message to ensure that the approach will work, but now it won't help anyway).

So I have two ways. 
Fixing it somehow locally on my side or keeping fork...

kurotych:
I've created related PR to [helium-crypto-rs](https://github.com/helium/helium-crypto-rs/pull/65) that should fix the problem

kurotych:
Bug fix it merged.
@madninja Is it possible to release a new minor version of `helium-crypto-rs` to link this repository with the fixed version? 
Thanks

jeffgrunewald:
Your merged change to helium-crypto-rs has been released in 0.8.1 and https://github.com/helium/oracles/pull/642 should bring it into the file-store crate and resolve this issue.

PR #626, opened 2023-09-11
Add bi directional stream for beacons/witnesses

bbalser:
Proto: https://github.com/helium/proto/pull/370

### DIFF
--- a/ingest/src/server_iot.rs
+++ b/ingest/src/server_iot.rs
@@ -59,10 +59,7 @@ impl StreamState {
         }
     }
 
-    async fn send_offer(
-        &mut self,
-        tx: &Sender<Result<LoraStreamResponseV1, Status>>,
-    ) -> Result<(), Status> {
+    async fn send_offer(&mut self, tx: &Sender<Result<LoraStreamResponseV1, Status>>) -> bool {
         let nonce: Nonce = rand::random();
         self.nonce = Some(nonce);
         self.timeout = Instant::now() + self.session_key_offer_timeout;
@@ -73,7 +70,7 @@ impl StreamState {
             })),
         }))
         .await
-        .map_err(|_| Status::unknown("unable to send LoraStreamSessionOffer"))
+        .is_ok()
     }
 
     fn initialize(&mut self, init: LoraStreamSessionInitV1) -> Result<(), Status> {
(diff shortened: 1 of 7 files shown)


ingest/src/server_iot.rs
```diff
@@ -7,58 +7,268 @@ use file_store::{
...
+    timeout: Instant,
+    nonce: Option<Nonce>,
+    session_key_offer_timeout: std::time::Duration,
+    session_key_timeout: std::time::Duration,
+}
+
+impl StreamState {
+    fn new(server: &GrpcServer) -> StreamState {
+        StreamState {
+            beacon_report_sink: server.beacon_report_sink.clone(),
```
andymck, line 47: maybe just clone the GrpcServer struct itself into StreamState, given you are coping all the fields anyway

andymck (approved):

ingest/src/server_iot.rs
bbalser: my preference would be to leave it as is, as I think it is clearer what data the StreamState needs, but I am not super attached to it, so if you or others feel strongly about it, I will make the change.

andymck: Its more of a nit, so no big deal either way

ingest/src/server_iot.rs
```diff
@@ -35,38 +183,99 @@ impl ManagedTask for GrpcServer {
...
+                .map_err(|_| Status::invalid_argument("invalid signature"))
+        })
+}
+
+async fn handle_beacon_report(
+    file_sink: &FileSinkClient,
+    timestamp: u64,
+    report: LoraBeaconReportReqV1,
+    signing_key: Option<&PublicKey>,
+    expected_pubkey_bytes: Option<&Vec<u8>>,
```
jeffgrunewald, line 236: can this be `Option<&[u8]>`? maybe with `bytes == report.pub_key.as_ref() | report.pub_key.as_slice()`?

bbalser: not sure why this is better since the pub_key comes as a `Vec<u8>` and we just store it to compare it to the same field in subsequent messages.

maplant (approved):

ingest/src/server_iot.rs
bbalser: Ok, Rondo convinced why this is better, will change

bbalser: Fixed ff72f1b

andymck (approved):

PR #627, opened 2023-09-14
Add support for denied edges in the denylist

Vagabond:
This change adds support for denying of hotspot-hotspot edges when the denylist system considers an edge invalid because of one of its classifiers but is unable to assign blame to a specific hotspot. This allows the denylist to deny earnings for bad edges without accidentally denying hotspots that are simply adjacent or enmeshed in a cluster of badly behaving hotspots, or anoymous radio repeater devices.

Hotspot edges are not directional, so A->B is considered the same as B->A. To accomplish this, the hotspot keys are always lexiographically ordered when constructing the hash to check the denylist.

### DIFF
--- a/denylist/src/denylist.rs
+++ b/denylist/src/denylist.rs
@@ -37,6 +37,22 @@ impl TryFrom<Vec<PublicKeyBinary>> for DenyList {
     }
 }
 
+impl TryFrom<Vec<(PublicKeyBinary, PublicKeyBinary)>> for DenyList {
+    type Error = Error;
+    fn try_from(v: Vec<(PublicKeyBinary, PublicKeyBinary)>) -> Result<Self> {
+        let keys: Vec<u64> = v.into_iter().map(|e| edge_hash((e.0.as_ref(), e.1.as_ref()))).collect();
+        let filter = Xor32::from(&keys);
+        let client = DenyListClient::new()?;
+        Ok(Self {
+            tag_name: 0,
+            client,
+            filter,
+            sign_keys: vec![],
+        })
+    }
+}
+
+
 impl DenyList {
     pub fn new(settings: &Settings) -> Result<Self> {
         tracing::debug!("initializing new denylist");
--- a/denylist/src/denylist.rs
+++ b/denylist/src/denylist.rs
@@ -110,6 +110,17 @@ impl DenyList {
         }
         self.filter.contains(&public_key_hash(pub_key))
     }
+
+    pub fn check_edge(&self, beaconer: &[u8], witness: &[u8]) -> bool {
+        if self.filter.len() == 0 {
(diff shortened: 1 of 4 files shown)


Vagabond:
Related to helium/oracles#22

denylist/src/denylist.rs
```diff
@@ -150,6 +161,13 @@ fn public_key_hash<R: AsRef<[u8]>>(public_key: R) -> u64 {
     hasher.finish()
 }
 
+fn edge_hash(edge_key: (&[u8], &[u8])) -> u64 {
+    let mut hasher = XxHash64::default();
+    hasher.write(edge_key.0);
```
andymck, line 166: I assume the denylist filter is being updated or has already to include these edge keys ?

iot_verifier/src/poc.rs
```diff
@@ -489,6 +494,36 @@ fn verify_denylist(pub_key: &PublicKeyBinary, deny_list: &DenyList) -> GenericVe
     //
     Ok(())
 }
+
+/// verify if gateway-gateway edge is on the deny list
+/// note that the order of the gateway keys is unimportant as edges are not considered directional
+fn verify_edge_denylist(
```
andymck, line 500: need to add a test here to cover this ( see the existing `test_verify_denylist` test for reference )

iot_verifier/src/poc.rs
```diff
@@ -489,6 +494,36 @@ fn verify_denylist(pub_key: &PublicKeyBinary, deny_list: &DenyList) -> GenericVe
...
+        tracing::debug!(
+            "report verification failed, reason: {:?}.
+            beacon: {}, witness {}, tagname: {}",
+            InvalidReason::Denied,
+            beaconer,
+            witness,
+            deny_list.tag_name
+        );
+        return Err(InvalidResponse {
+            reason: InvalidReason::Denied,
```
andymck, line 515: Should we add a new invalid reason such as DeniedEdge or similar ? Given there are two scenarios in which a hotspot can be denied do we need to distinguish the denied reason here?

bbalser (approved):

iot_verifier/src/poc.rs
andymck: Let me put this differently, when owners complain about being on the denylist....do you want to readily determine whether its because its due to the edge check or the regular check ?

riobah: Taking Andy's comment one step further, if multiple classifiers will be generating these denied edges, it could be nice to provide the classifier ID to the users for clarification. Like this witness report is "denied because of distance insensitivity". This would require more than one new denied reasons, like `denied_edge_distance_insensitivity`, `denied_edge_witness_reciprocity`, etc

jeffgrunewald (approved):

PR #632, opened 2023-09-28
verify asserted device type against wifi HB cell type

andymck:


mobile_config/src/gateway_info.rs
```diff
@@ -33,11 +33,15 @@ impl From<GatewayInfoProto> for GatewayInfo {
     fn from(info: GatewayInfoProto) -> Self {
         let metadata = if let Some(metadata) = info.metadata {
             u64::from_str_radix(&metadata.location, 16)
-                .map(|location| GatewayMetadata { location })
+                .map(|location| GatewayMetadata {
```
jeffgrunewald, line 36: not sure this implementation makes sense any longer since a mobile hotspot should _always_ have a device type the struct going forward should probably be
```
GatewayInfo {
    address: PublicKeyBinary,
    metadata: GatewayMetadata {
        location: Option<u64>,
        device_type: DeviceTypeEnum
    }
}
```
if we wanna minimize definitions we could probably pull the device type enum in from the HPL since it's already a dependency of Oracles but at the very least I'd make this an enum rather than a string

mobile_verifier/src/lib.rs
```diff
@@ -15,10 +15,11 @@ pub use settings::Settings;
 
 use async_trait::async_trait;
 
+use mobile_config::gateway_info::GatewayMetadata;
 pub enum GatewayResolution {
     GatewayNotFound,
     GatewayNotAsserted,
-    AssertedLocation(u64),
+    GatewayAsserted(GatewayMetadata),
```
jeffgrunewald, line 22: does it make sense to include the entire metadata struct in the `GatewayAsserted` variant when it should have metadata to include the device type even if it's not asserted?

PR #638, opened 2023-10-02
Use Jemalloc for ingest, with profiling enabled

maplant:
This PR created just for visibility purposes.

In addition to this branch, `libunwind` needs to be installed on the system, and ingest needs to be run with the following environmental variables:

```export _RJEM_MALLOC_CONF=prof:true,lg_prof_interval:32,lg_prof_sample:19```

This is based off of https://gist.github.com/ordian/928dc2bd45022cddd547528f64db9174


jeffgrunewald:
is the only reason we're not rebasing the https://github.com/helium/oracles/pull/465 b/c we want to limit the changes to _just_ the ingest service? i'm unclear the difference between `tikv-jemallocator` and `jemallocator` crate in the previous PR but according to crates.io they appear to be maintained by the same person, pointing to the same github repo and updated on the same schedule so i'm guessing they're equivalent?

PR #647, opened 2023-10-13
Andymck/beacon fixed windows

andymck:
With this change hotspots can now beacon once and only once within N time buckets per day.  By default those buckets are 6 hours in width and based on UTC time.  
A beacon will pass the schedule verification if it is the first beacon for any one bucket irrespective of when it was received within that bucket window.  Any subsequent beacon within the same bucket will be rejected

### DIFF
--- a/iot_verifier/src/poc.rs
+++ b/iot_verifier/src/poc.rs
@@ -7,7 +7,7 @@ use crate::{
     region_cache::{RegionCache, RegionCacheError},
 };
 use beacon;
-use chrono::{DateTime, Duration, Utc};
+use chrono::{DateTime, Duration, DurationRound, Utc};
 use denylist::denylist::DenyList;
 use file_store::{
     iot_beacon_report::{IotBeaconIngestReport, IotBeaconReport},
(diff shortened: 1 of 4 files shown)


bbalser (approved):

iot_verifier/src/poc.rs
```diff
@@ -447,17 +439,37 @@ pub fn do_witness_verifications(
...
     beacon_received_ts: DateTime<Utc>,
-    beacon_interval: Duration,
-    beacon_interval_tolerance: Duration,
+    beacon_interval: u64,
 ) -> GenericVerifyResult {
     match last_beacon {
         Some(last_beacon) => {
-            let interval_since_last_beacon = beacon_received_ts - last_beacon.timestamp;
-            if interval_since_last_beacon < (beacon_interval - beacon_interval_tolerance) {
+            let last_bucket = last_beacon
```
jeffgrunewald, line 446: This operation feels like a good place for a helper function since you’re performing the same logic on the last and current timestamp or else pull the whole thing out into a helper that returns a boolean

jeffgrunewald (approved):

PR #650, opened 2023-10-17
Two phase commit for iot packet verifier

maplant:
TODO:
- [x] Fix integration tests for IOT packet verifier
- [x] Add unit test for two phase commits
~~Integrate into mobile packet verifier~~ (moving to separate PR)


andymck (approved):
lgtm, given the degree of changes should be fully tested on devnet before hitting mainnet

iot_packet_verifier/src/pending.rs
```diff
@@ -0,0 +1,464 @@
...
+pub async fn confirm_pending_txns<S>(
+    pending_tables: &impl PendingTables,
+    solana: &S,
+) -> Result<(), ConfirmPendingError<S::Error>>
+where
+    S: SolanaNetwork,
+{
+    // Fetch all pending transactions and confirm them
+    let pending = pending_tables.fetch_all_pending_txns().await?;
+    for pending in pending {
```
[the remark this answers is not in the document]
maplant: Two reasons. The first is for testing, so that I can create a unit test that completes instantly.
The second is that it allows us to start the iot packet verifier instantly when there are no outstanding transactions

bbalser (approved):

iot_packet_verifier/src/burner.rs
```diff
@@ -85,33 +82,37 @@ where
...
 
-        let amount = amount as u64;
-
+        let txn = self
+            .solana
+            .make_burn_transaction(&payer, amount)
+            .await
+            .map_err(BurnError::SolanaError)?;
+        self.pending_tables
+            .submit_txn(&payer, amount, txn.get_signature())
```
jeffgrunewald, line 100: if this is one of the places @bbalser mentioned where naming got tricky, i'd agree: `submit_txn` immediately followed by `submit_transaction`

iot_packet_verifier/src/burner.rs
```diff
@@ -85,33 +82,37 @@ where
...
         self.solana
-            .burn_data_credits(&payer, amount)
+            .submit_transaction(&txn)
             .await
             .map_err(BurnError::SolanaError)?;
 
         // Now that we have successfully executed the burn and are no longer in
         // sync land, we can remove the amount burned:
-        self.pending_burns
+        let mut pending_tables_txn = self.pending_tables.begin().await?;
```
jeffgrunewald, line 109: this comment above looks like it could use an update

iot_packet_verifier/src/pending.rs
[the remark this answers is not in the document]
jeffgrunewald: what about `if pending.len() > 0 { sleep_for_1_min_only_once }`? that should address the second point and if you have `fetch_all_pending_txns` order the results by `time_of_submission` then you can just grab the first or last (depending on how you want to order the results) and only sleep for 1 min iff the list is non-empty and the most recent submission was less than 1 min ago

iot_packet_verifier/src/pending.rs
```diff
@@ -0,0 +1,467 @@
...
+        // give the transaction plenty of time to be confirmed
+        let time_since_submission = Utc::now() - pending.time_of_submission;
+        if Duration::minutes(1) > time_since_submission {
+            tokio::time::sleep((Duration::minutes(1) - time_since_submission).to_std()?).await;
+        }
+
+        let mut txn = pending_tables.begin().await?;
+        // We remove the transaction regardless of whether it has been confirmed
+        // or not:
+        txn.confirm_txn(&pending.signature).await?;
```
jeffgrunewald, line 75: what does it mean to "remove the transaction regardless of whether or not it's been confirmed" if we only remove the the burn amount from the pending burns table if it _is_ confirmed? what are we removing regardless?

iot_packet_verifier/src/pending.rs
```diff
@@ -0,0 +1,467 @@
...
+    async fn submit_txn(
+        &self,
+        payer: &PublicKeyBinary,
+        amount: u64,
+        signature: &Signature,
+    ) -> Result<(), sqlx::Error> {
+        sqlx::query(
+            r#"
+            INSERT INTO pending_txns (signature, payer, amount, time_of_submission)
+            VALUES ($1, $2, $3, $4)
```
jeffgrunewald, line 144: is there any chance a txn could be submitted more than once that you need to have an `ON CONFLICT` here?

maplant: I don't believe so, signatures should be unique even with the same inputs

madninja: that's correct.. I'm assuming signature is enforced with a unique constraint?

iot_packet_verifier/src/burner.rs
maplant: I renamed the pending tables methods to `add_pending_transaction` and `remove_pending_transaction`

iot_packet_verifier/src/pending.rs
maplant: I explained this comment a little further, but I think the renamed methods should also help quite significantly here

PR #652, opened 2023-10-20
assume beacon interval defined in seconds

andymck:
Assume beacon interval is defined in settings as seconds
Convert beacon interval from seconds to Duration via settings accessor
Use beacon interval as duration throughout iot verifier

### DIFF
--- a/iot_verifier/src/settings.rs
+++ b/iot_verifier/src/settings.rs
@@ -1,3 +1,4 @@
+use anyhow::bail;
 use chrono::Duration;
 use config::{Config, Environment, File};
 use serde::Deserialize;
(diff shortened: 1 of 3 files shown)


bbalser (approved):

iot_verifier/src/settings.rs
```diff
@@ -37,7 +37,7 @@ pub struct Settings {
     pub reward_offset_minutes: i64,
     #[serde(default = "default_max_witnesses_per_poc")]
     pub max_witnesses_per_poc: u64,
-    /// The cadence at which hotspots are permitted to beacon (in hours)
+    /// The cadence at which hotspots are permitted to beacon (in seconds)
```
jeffgrunewald, line 40: if this should always be a factor of 24 then add a check to the accessor that validates it evenly divides by 24 or else panics before returning the i64 converted to a duration

andymck: addressed here: b62179d0dbc7d16b8f4bd2321ab20a7dc8f80454

jeffgrunewald (approved):

PR #655, opened 2023-10-30
Add traits for iot and mobile config clients

bbalser:


mobile_verifier/src/lib.rs
```diff
@@ -14,16 +14,17 @@ mod telemetry;
...
     GatewayNotFound,
     GatewayNotAsserted,
     AssertedLocation(u64),
 }
 
-#[async_trait]
-pub trait GatewayResolver {
-    type Error: std::error::Error + Send + Sync + 'static;
+#[async_trait::async_trait]
+pub trait GatewayResolver: Clone + Send + Sync + 'static {
```
maplant, line 26: Can these traits not get outright replaced by the mock mobile config traits?

andymck: they could yes and in fact I had dropped these ones in the original rebase and used only the mobile config traits but then re-added as it necessitates some changes to the flow in heartbeat validations.  In this PR i want to minimise unnecessary code changes.  I have another PR which coming later which adds additional integration tests for the rewarder etc and I suspect these mobile verifier traits may get dropped then.

maplant: As long as there is a plan for replacing these then I'm Ok approving, although I think it would be totally appropriate to remove these now, considering other similar local traits have been removed in this PR

maplant (approved):

jeffgrunewald (approved):

mobile_verifier/src/lib.rs
```diff
@@ -14,16 +14,17 @@ mod telemetry;
 pub use settings::Settings;
 
 use async_trait::async_trait;
+use std::error::Error;
```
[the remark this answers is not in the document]
andymck: it is not

PR #659, opened 2023-11-03
Change mobile packet verifier to support rewardable bytes

bbalser:
Still in draft, want to think through what to do in cancelled invalid data transfer sessions and the switchover

### DIFF
--- a/file_store/src/mobile_session.rs
+++ b/file_store/src/mobile_session.rs
@@ -178,10 +178,9 @@ impl From<DataTransferEvent> for DataTransferEventProto {
 #[derive(Serialize, Clone, Debug)]
 pub struct DataTransferSessionReq {
     pub data_transfer_usage: DataTransferEvent,
-    pub reward_cancelled: bool,
+    pub rewardable_bytes: u64,
     pub pub_key: PublicKeyBinary,
     pub signature: Vec<u8>,
-    pub rewardable_bytes: u64,
 }
 
 impl MsgDecode for DataTransferSessionReq {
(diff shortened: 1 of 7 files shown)


file_store/src/mobile_session.rs
```diff
@@ -203,14 +203,16 @@ impl TryFrom<DataTransferSessionReqV1> for DataTransferSessionReq {
...
 impl From<DataTransferSessionReq> for DataTransferSessionReqV1 {
     fn from(v: DataTransferSessionReq) -> Self {
         let report: DataTransferEventProto = v.data_transfer_usage.into();
         Self {
             data_transfer_usage: Some(report),
-            reward_cancelled: v.reward_cancelled,
+            rewardable_bytes: v.rewardable_bytes,
             pub_key: v.pub_key.into(),
             signature: v.signature,
+            reward_cancelled: false,
```
andymck, line 215: wont this would break the ability to replay packets prior to this change ?
You could address this by adding a method to generate rewardable_bytes value for older packets from the existing reward_cancelled/downloaded_bytes/uploaded_bytes fields

bbalser: I could also just continue reading the reward_cancelled field from the proto as longs as we are ok with marking the function `#[allow(deprecated)]`

andymck: perhaps but a helper method that generates the rewardable_bytes value from a sum of upload/download bytes for those packets which do not include the rewardable_bytes field and which have reward_cancelled: false feels like a better long term fit

andymck (approved):

jeffgrunewald (approved):

PR #660, opened 2023-11-04
Update locate command to support validated heartbeats and speedtest a…

bbalser:
…verages

### DIFF
--- a/file_store/src/heartbeat.rs
+++ b/file_store/src/heartbeat.rs
@@ -88,49 +88,55 @@ impl MsgTimestamp<Result<DateTime<Utc>>> for CellHeartbeatIngestReportV1 {
     }
 }
 
-#[derive(Serialize, Deserialize, Debug, Clone)]
-pub struct ValidatedHeartbeat {
-    pub cbsd_id: String,
-    pub pub_key: PublicKeyBinary,
-    pub reward_multiplier: f32,
-    pub timestamp: DateTime<Utc>,
-    pub cell_type: CellType,
-    pub validity: HeartbeatValidity,
-    pub lat: f64,
-    pub lon: f64,
-    pub coverage_object: Vec<u8>,
-    pub location_validation_timestamp: DateTime<Utc>,
-    pub distance_to_asserted: u64,
-}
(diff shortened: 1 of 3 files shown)


file_store/src/heartbeat.rs
```diff
@@ -76,6 +79,51 @@ impl MsgTimestamp<Result<DateTime<Utc>>> for CellHeartbeatIngestReportV1 {
     }
 }
 
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct ValidatedHeartbeat {
```
andymck, line 83: NOTE, there is an existing ValidatedHeartbeat struct in mobile verifier...which does already include all the data you require here.  Have you considered if these can be harmonized ?

bbalser: missed that those exist, will look at pulling those into file_store

file_store/src/heartbeat.rs
```diff
@@ -76,6 +79,51 @@ impl MsgTimestamp<Result<DateTime<Utc>>> for CellHeartbeatIngestReportV1 {
...
+    pub cbsd_id: String,
+    pub pub_key: PublicKeyBinary,
+    pub reward_multiplier: f32,
```
jeffgrunewald, line 86: should we be keeping all the numbers that factor into rewards calculations be kept as integers or decimals? we've allowed a _few_ floats but I think those have so far always been relative to things like transmit values and other RF-things

jeffgrunewald (commented):
assuming you want to remove the similar ValidatedHeartbeat type from the mobile verifier in a separate change

andymck (approved):

bbalser:
> assuming you want to remove the similar ValidatedHeartbeat type from the mobile verifier in a separate change

They are different enough, I am not sure they should be combined. This new one is meant as the rust version of the proto, where the one in the mobile verifier is the struct representing a validated heartbeat during the validation process and has other stuff in needed for that process.  For the moment, I moved these new ones into a cli module.

jeffgrunewald (approved):

PR #661, opened 2023-11-06
Update proto to include rewardable bytes

bbalser:


andymck (approved):

file_store/src/mobile_session.rs
```diff
@@ -190,6 +191,7 @@ impl MsgDecode for DataTransferSessionReq {
 impl TryFrom<DataTransferSessionReqV1> for DataTransferSessionReq {
     type Error = Error;
 
+    #[allow(deprecated)]
```
maplant, line 194: Instead of doing this, don't set the deprecated field and do ..Default::default() at the end of the struct

maplant (approved):

file_store/src/mobile_session.rs
bbalser: Addressed in d368f75

maplant (approved):

PR #663, opened 2023-11-09
add service provider rewards support

andymck:
related hip: https://github.com/helium/HIP/blob/main/0087-proportional-service-provider-rewards.md

related proto pr: https://github.com/helium/proto/pull/377

### DIFF
--- a/mobile_config/migrations/5_carrier_service.sql
+++ b/mobile_config/migrations/5_carrier_service.sql
@@ -1,0 +1,7 @@
+create table carrier_keys (
+    pubkey text not null,
+    entity_key text not null,
+    created_at timestamptz not null default now(),
+    updated_at timestamptz not null default now(),
+    PRIMARY KEY(pubkey)
+);
--- a/mobile_config/migrations/5_carrier_service.sql
+++ b/mobile_config/migrations/5_carrier_service.sql
@@ -1,7 +1,6 @@
 create table carrier_keys (
-    pubkey text not null,
+    pubkey text primary key not null,
     entity_key text not null,
     created_at timestamptz not null default now(),
-    updated_at timestamptz not null default now(),
-    PRIMARY KEY(pubkey)
+    updated_at timestamptz not null default now()
 );
--- a/mobile_config/src/client/carrier_service_client.rs
+++ b/mobile_config/src/client/carrier_service_client.rs
@@ -3,7 +3,7 @@ use async_trait::async_trait;
 use file_store::traits::MsgVerify;
 use helium_crypto::{Keypair, PublicKey, Sign};
 use helium_proto::{
-    services::{mobile_config, poc_mobile::ServiceProvider, Channel},
+    services::{mobile_config, Channel},
     Message,
 };
 use retainer::Cache;
@@ -12,17 +12,15 @@ use std::{sync::Arc, time::Duration};
 #[async_trait]
 pub trait CarrierServiceVerifier {
     type Error;
-    async fn key_to_rewardable_entity<'a>(
-        &self,
-        entity_id: &'a str,
-    ) -> Result<ServiceProvider, Self::Error>;
+    async fn key_to_rewardable_entity<'a>(&self, entity_id: &'a str)
+        -> Result<String, Self::Error>;
 }
 #[derive(Clone)]
 pub struct CarrierServiceClient {
     client: mobile_config::CarrierServiceClient<Channel>,
     signing_key: Arc<Keypair>,
     config_pubkey: PublicKey,
-    cache: Arc<Cache<String, ServiceProvider>>,
+    cache: Arc<Cache<String, String>>,
     cache_ttl: Duration,
 }
 
@@ -30,12 +28,9 @@ pub struct CarrierServiceClient {
 impl CarrierServiceVerifier for CarrierServiceClient {
     type Error = ClientError;
 
-    async fn key_to_rewardable_entity<'a>(
-        &self,
-        pubkey: &'a str,
-    ) -> Result<ServiceProvider, ClientError> {
+    async fn key_to_rewardable_entity<'a>(&self, pubkey: &'a str) -> Result<String, ClientError> {
         if let Some(carrier_found) = self.cache.get(&pubkey.to_string()).await {
-            return Ok(*carrier_found.value());
+            return Ok(carrier_found.value().clone());
         }
 
         let mut request = mobile_config::CarrierKeyToEntityReqV1 {
@@ -49,12 +44,12 @@ impl CarrierServiceVerifier for CarrierServiceClient {
             Ok(verify_res) => {
                 let response = verify_res.into_inner();
                 response.verify(&self.config_pubkey)?;
-                to_service_provider(&response.entity_key)?
+                response.entity_key
             }
             Err(status) => Err(status)?,
         };
         self.cache
-            .insert(pubkey.to_string(), response, self.cache_ttl)
+            .insert(pubkey.to_string(), response.clone(), self.cache_ttl)
             .await;
         Ok(response)
     }
@@ -77,12 +72,5 @@ impl CarrierServiceClient {
             cache_ttl: settings.cache_ttl(),
             cache,
         })
-    }
-}
-
-fn to_service_provider(key: &str) -> Result<ServiceProvider, ClientError> {
-    match key {
-        "helium_mobile" => Ok(ServiceProvider::HeliumMobile),
-        _ => Err(ClientError::UnknownServiceProvider)?,
     }
 }
--- a/mobile_config/src/client/carrier_service_client.rs
+++ b/mobile_config/src/client/carrier_service_client.rs
@@ -3,7 +3,7 @@ use async_trait::async_trait;
 use file_store::traits::MsgVerify;
 use helium_crypto::{Keypair, PublicKey, Sign};
 use helium_proto::{
-    services::{mobile_config, Channel},
+    services::{mobile_config, poc_mobile::ServiceProvider, Channel},
     Message,
 };
 use retainer::Cache;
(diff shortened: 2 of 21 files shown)


mobile_verifier/src/reward_shares.rs
```diff
@@ -197,6 +204,112 @@ impl MapperShares {
...
+impl ServiceProviderShares {
+    pub fn new(shares: Vec<ServiceProviderDataSession>) -> Self {
+        Self { shares }
+    }
+
+    pub async fn from_payers_dc(
+        payer_shares: HashMap<String, u64>,
+        client: &dyn CarrierServiceVerifier<Error = ClientError>,
+    ) -> anyhow::Result<ServiceProviderShares> {
+        let mut shares = vec![];
```
jeffgrunewald, line 221: since `ServiceProviderShares` derives the default trait you can just `ServiceProviderShares::default()` here and then return that once you're done loading it up

mobile_verifier/src/reward_shares.rs
```diff
@@ -197,6 +204,112 @@ impl MapperShares {
...
+            })
+        }
+        Ok(ServiceProviderShares { shares })
+    }
+
+    fn total_dc(&self) -> Decimal {
+        self.shares.iter().map(|v| v.total_dcs).sum()
+    }
+
+    pub fn get_scheduled_tokens_for_service_providers(&self, duration: Duration) -> Decimal {
```
jeffgrunewald, line 236: should this be a method off the `ServiceProviderShares` struct or a free function alongside `get_total_scheduled_tokens`?

reward_index/src/indexer.rs
```diff
@@ -150,6 +152,19 @@ impl Indexer {
                         },
                         r.discovery_location_amount,
                     )),
+                    Some(MobileReward::ServiceProviderReward(r)) => {
+                        if let Some(sp) = ServiceProvider::from_i32(r.service_provider_id) {
+                            Ok((
+                                RewardKey {
+                                    key: sp.as_str_name().to_string(),
+                                    reward_type: RewardType::MobileSubscriber,
```
jeffgrunewald, line 160: is this the right reward type for a service provider?

mobile_config/migrations/5_carrier_service.sql
```diff
@@ -0,0 +1,7 @@
+create table carrier_keys (
+    pubkey text not null,
```
jeffgrunewald, line 2:
```suggestion
    pubkey text primary key not null,
```

mobile_config/migrations/5_carrier_service.sql
```diff
@@ -0,0 +1,7 @@
...
+    entity_key text not null,
+    created_at timestamptz not null default now(),
+    updated_at timestamptz not null default now(),
+    PRIMARY KEY(pubkey)
```
jeffgrunewald, line 6:
```suggestion
```

mobile_config/src/client/carrier_service_client.rs
```diff
@@ -0,0 +1,76 @@
...
+    type Error;
+    async fn key_to_rewardable_entity<'a>(&self, entity_id: &'a str)
+        -> Result<String, Self::Error>;
+}
+#[derive(Clone)]
+pub struct CarrierServiceClient {
+    client: mobile_config::CarrierServiceClient<Channel>,
+    signing_key: Arc<Keypair>,
+    config_pubkey: PublicKey,
+    cache: Arc<Cache<String, String>>,
```
jeffgrunewald, line 23: how frequently is this verification being called? once a day during rewards calculation or periodically? seems like we're not likely to have enough service providers on the network collecting rewards for a while such that we need a cache here

jeffgrunewald (commented):
If I understand correctly here, the `payer` that we're using to sum data transfer credits burned by a service provider is the b58 text encoding of the carrier's public key and then we're storing a mapping of that public key to the string-entity key (name) of the service provider stored in the mobile config service but is that mapping not available on-chain such that we can just treat the service provider as a "rewardable entity" and look up their entity key from the metadata database the foundation populates?

mobile_verifier/src/reward_shares.rs
andymck: these getters are all moved to free functions in #683 which will go live with service provider rewards

mobile_config/migrations/5_carrier_service.sql
andymck: addressed here: 3f5924313ce8efe32a580e1981be7f4018355ee8

mobile_config/migrations/5_carrier_service.sql
andymck: addressed here: 3f5924313ce8efe32a580e1981be7f4018355ee8

mobile_verifier/src/reward_shares.rs
andymck: addressed here: 3f5924313ce8efe32a580e1981be7f4018355ee8

reward_index/src/indexer.rs
andymck: it most def is not, addressed here: 3f5924313ce8efe32a580e1981be7f4018355ee8

andymck:
> If I understand correctly here, the `payer` that we're using to sum data transfer credits burned by a service provider is the b58 text encoding of the carrier's public key and then we're storing a mapping of that public key to the string-entity key (name) of the service provider stored in the mobile config service but is that mapping not available on-chain such that we can just treat the service provider as a "rewardable entity" and look up their entity key from the metadata database the foundation populates?

good question, my understanding at the time was that we needed to add this to our internally managed DB.  I dont recall the exact reason but i think it relates to the fact we only have the payer public key and that is not available as a direct translation on the metadata db populated by  foundation ?  But i may have just misunderstood and my context of the metadata DB is not that great.  @bbalser thoughts?

mobile_verifier/src/reward_shares.rs
```diff
@@ -197,6 +204,112 @@ impl MapperShares {
...
+        } else {
+            Decimal::ZERO
+        }
+    }
+
+    async fn payer_key_to_service_provider(
+        payer: &str,
+        client: &dyn CarrierServiceVerifier<Error = ClientError>,
+    ) -> anyhow::Result<ServiceProvider> {
+        tracing::info!("getting entity key for pubkey {:?}", payer);
```
jeffgrunewald, line 300:
```suggestion
        tracing::info!(payer, "getting entity key for service provider");
```

jeffgrunewald (approved):

mobile_verifier/src/reward_shares.rs
```diff
@@ -197,6 +204,112 @@ impl MapperShares {
...
+            (total_rewards / total_shares)
+                .round_dp_with_strategy(DEFAULT_PREC, RoundingStrategy::MidpointNearestEven)
+        } else {
+            Decimal::ZERO
+        }
+    }
+
+    async fn payer_key_to_service_provider(
+        payer: &str,
+        client: &dyn CarrierServiceVerifier<Error = ClientError>,
```
bbalser, line 298: isn't this easier to read as &impl CarrierServiceVerifier<Error = ClientError>

mobile_config/migrations/5_carrier_service.sql
```diff
@@ -0,0 +1,6 @@
+create table carrier_keys (
+    pubkey text primary key not null,
```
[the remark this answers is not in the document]
andymck: its not a signing key, but i can rename to 'payer' or something more equivalent to its function

mobile_verifier/src/data_session.rs
```diff
@@ -140,6 +148,31 @@ pub async fn aggregate_hotspot_data_sessions_to_dc<'a>(
...
+pub async fn sum_data_sessions_to_dc_by_payer<'a>(
+    exec: impl sqlx::PgExecutor<'a> + Copy + 'a,
+    epoch: &'a Range<DateTime<Utc>>,
+) -> Result<HashMap<String, u64>, sqlx::Error> {
+    Ok(sqlx::query(
+        r#"
+        SELECT payer as sp, sum(num_dcs)::bigint as total_dcs
+        FROM hotspot_data_transfer_sessions
+        WHERE received_timestamp >= $1 and received_timestamp < $2
+        GROUP BY payer
```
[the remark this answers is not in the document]
andymck: GROUP by is required as it used an aggregate function

mobile_verifier/src/reward_shares.rs
andymck: fixed on #669 to avoid rebase issue later

mobile_config/migrations/5_carrier_service.sql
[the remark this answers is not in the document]
andymck: in further thinking i think its fine as is, the table is named as 'carrier_keys` and the carrier has a pubkey which is used to find its associated entity key

mobile_verifier/src/reward_shares.rs
andymck: addressed here: e4b9890f0b20b7272b15c56710ae389dde8d573a

mobile_verifier/src/reward_shares.rs
andymck: addressed here: e4b9890f0b20b7272b15c56710ae389dde8d573a

mobile_verifier/src/data_session.rs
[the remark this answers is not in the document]
jeffgrunewald: i'm asking if the sql works properly when you've aliased the `payer` column to `sp` in this query but then referenced the `payer` column by its original name.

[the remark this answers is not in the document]
andymck: yip, the sql works as is

bbalser (approved):

maplant (approved):

PR #672, opened 2023-11-19
draft change to scale wifi reward weights vs cbrs weights

andymck:
wip....

### DIFF
--- a/mobile_verifier/src/reward_shares.rs
+++ b/mobile_verifier/src/reward_shares.rs
@@ -1384,7 +1384,8 @@ mod test {
 
         // setup heartbeats
         // add 2 wifi indoor HBs
-        // 2 with a valid location timestamp but a distance to asserted > max
+        // 1 with an invalid location timestamp
+        // 1 with a valid location timestamp but a distance to asserted > max
         // this results in a final location trust score of 0.25, made up as follows:
         // HB1 location score = 0.25
         // HB1 location score = 0.25
@@ -1396,8 +1397,8 @@ mod test {
                 cell_type: CellType::NovaGenericWifiIndoor,
                 coverage_object: Some(Uuid::new_v4()),
                 latest_timestamp: DateTime::<Utc>::MIN_UTC,
-                location_validation_timestamp: Some(timestamp),
-                distance_to_asserted: Some(1000),
+                location_validation_timestamp: None,
+                distance_to_asserted: None,
             },
             HeartbeatRow {
                 cbsd_id: None,
@@ -1518,7 +1519,8 @@ mod test {
         // setup heartbeats
         // add 4 wifi indoor HBs
         // 2 with a valid location timestamp and a distance to asserted < max
-        // 2 with a valid location timestamp but a distance to asserted > max
+        // 1 with an invalid location timestamp
+        // 1 with a valid location timestamp but a distance to asserted > max
         // this results in a final location trust score of 0.625, made up as follows:
         // HB1 location score = 1.0
         // HB2 location score = 1.0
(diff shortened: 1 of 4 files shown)


mobile_verifier/src/reward_shares.rs
```diff
@@ -203,16 +203,38 @@ pub fn dc_to_mobile_bones(dc_amount: Decimal, mobile_bone_price: Decimal) -> Dec
...
+pub struct Share {
+    cell_type_weight: Decimal,
+    speed_multiplier: Decimal,
+    aggregated_location_trust_weight: Decimal,
+    aggregated_hb_count: Decimal,
+}
+
+impl Share {
+    fn update(&mut self, heartbeat_reward: HeartbeatReward, speed_multiplier: Decimal) {
+        self.cell_type_weight = heartbeat_reward.cell_type_weight;
```
jeffgrunewald, line 216: seems like the `cell_type_weight` shouldn't change once set. If this single `update` method is supposed to act as an upsert operation that field should probably be an `cell_type_weight: Option<Decimal>` and then the `update` method checks and only sets the value once if it's not already set.

The perpetual assignment implies this value should actually change with the input of each new heartbeat when I don't think that makes sense; the weight should be stable for the life of the radio.

mobile_verifier/src/reward_shares.rs
```diff
@@ -233,13 +255,14 @@ impl PocShares {
...
-            *poc_shares
+            poc_shares
                 .hotspot_shares
                 .entry(heartbeat_reward.hotspot_key.clone())
                 .or_default()
                 .radio_shares
                 .entry(heartbeat_reward.cbsd_id.clone())
-                .or_default() += heartbeat_reward.reward_weight() * speedmultiplier;
+                .or_default()
+                .update(heartbeat_reward, speedmultiplier);
```
jeffgrunewald, line 265: so are we ever going to update the value of the given key and if so are we just overwriting the speedtest multiplier for that share? does the answer to that change drastically based on if we're talking about wifi rewards and cbrs rewards where we have one collection of rewards per device for wifi radios but a true aggregate of all radio rewards attached to a given gateway for cbrs radios?

mobile_verifier/src/reward_shares.rs
```diff
@@ -267,8 +290,9 @@ impl PocShares {
                 self.hotspot_shares
                     .into_iter()
                     .flat_map(move |(hotspot_key, RadioShares { radio_shares })| {
-                        radio_shares.into_iter().map(move |(cbsd_id, amount)| {
-                            let poc_reward = poc_rewards_per_share * amount;
+                        radio_shares.into_iter().map(move |(cbsd_id, share)| {
+                            let poc_reward = poc_rewards_per_share * share.total();
+                            println!("share: {:?}", share);
```
jeffgrunewald, line 295: remove before merging

mobile_verifier/src/reward_shares.rs
andymck: Call type weight and speed multiplier will not change.  They are located here for simplicity...the values are stable.  They will not change across updates.  I dont think wrapping each in an Option provides any real additional value and just makes the code more verbose.  Ideally they could be moved out of the Share struct and into the overall RadioShares struct but that complicates things further.  
The best thing to do here is see how this all gets refactored as part of making it compatible with modelled coverage v3 and see if a better home can be found then

mobile_verifier/src/reward_shares.rs
andymck: the update is called once per heartbeat reward.  This will be a single time for CBRS as there should only be a single heartbeat per hotspot/radio combo.  For wifi hotspots there will be N heartbeat rewards here and thus update will be called N times.

PR #673, opened 2023-11-22
Fix wifi rewards to calculate location trust score average

bbalser:


mobile_verifier/src/heartbeats/mod.rs
```diff
@@ -256,17 +256,54 @@ impl HeartbeatReward {
...
+                            .push(row);
+
+                        Ok(map)
+                    },
+                )
+                .await?;
+
+        Ok(
+            futures::stream::iter(heartbeat_rows).map(move |((hotspot_key, cbsd_id), rows)| {
+                let first = rows.first().unwrap();
```
jeffgrunewald, line 284: there's no chance this could ever be a None?

mobile_verifier/src/heartbeats/mod.rs
```diff
@@ -256,17 +256,54 @@ impl HeartbeatReward {
...
+                            row.distance_to_asserted,
+                            max_distance_to_asserted,
+                        )
+                    })
+                    .sum::<Decimal>()
+                    / Decimal::new(rows.len() as i64, 0);
+
+                HeartbeatReward {
+                    hotspot_key,
+                    cell_type: first.cell_type,
```
jeffgrunewald, line 299: guessing this is the thing you weren't happy about? feels like this is probably not the right abstraction if we're arbitrarily peeling off the first value of the list to grab these field values but understand now's not the time to fix it. probably something where the `HeartbeatReward` is just an envelope struct that contains the contents of the `HeartbeatRow` but then adds in the newly computed `reward_weight`

maplant (approved):

mobile_verifier/src/heartbeats/mod.rs
bbalser: since the map is created from the rows, at least one row has to exist for this to be called on the vec.

mobile_verifier/src/heartbeats/mod.rs
bbalser: I don't like this, but it works and the all the fields we are reading from the first row are the same in all rows.

mobile_verifier/src/heartbeats/mod.rs
jeffgrunewald: sure but is there definitely not a world where that query would never return any rows so you wouldn't be passing an empty/default map into this `stream::iter`?

bbalser: The function passed to `map` won't be executed if there are no entries in the iterator.

PR #675, opened 2023-11-30
add support for unallocated IOT rewards

andymck:
Linked proto PR: https://github.com/helium/proto/pull/384


iot_verifier/Cargo.toml
```diff
@@ -55,3 +55,4 @@ price = { path = "../price" }
 tokio-util = { workspace = true }
 tokio-stream = { workspace = true }
 task-manager = { path = "../task_manager" }
+time = "0.3.17"
```
bbalser, line 58: We don't usually pin deps to a specific version in the Cargo.toml, is this exact version needed?

andymck: actually this is not even required...will remove

bbalser (approved):

iot_verifier/src/reward_share.rs
```diff
@@ -329,66 +340,68 @@ impl GatewayShares {
...
-        // compute the awards per hotspot
+
         tracing::info!(
             %total_dc_shares,
             %total_dc_transfer_rewards_used,
             %dc_transfer_rewards_unused,
             %dc_transfer_rewards_per_share,
             "data transfer rewards"
         );
+        self.beacon_rewards_per_share = beacon_rewards_per_share;
```
maplant, line 357: Are these values defaulted before this function is called? If so, this function should be a constructor, not a &mut function. Could be a pretty dangerous footgun

iot_verifier/src/reward_share.rs
```diff
@@ -898,14 +934,22 @@ mod test {
             reward_shares_in_dec(dec!(150), dec!(350), gw6_dc_spend),
         ); // 0.0150, 0.0350
 
-        let gw_shares = GatewayShares { shares };
+        let mut gw_shares = GatewayShares {
```
maplant, line 937: Instead of doing this, construct a new set of gateway shares from an iterator of hashmaps

iot_verifier/src/reward_share.rs
```diff
@@ -329,66 +340,68 @@ impl GatewayShares {
...
-            .map(|gateway_reward| proto::IotRewardShare {
-                start_period: reward_period.start.encode_timestamp(),
-                end_period: reward_period.end.encode_timestamp(),
-                reward: Some(ProtoReward::GatewayReward(gateway_reward)),
+            .map(|gateway_reward| {
+                let total_gateway_reward = gateway_reward.dc_transfer_amount
+                    + gateway_reward.beacon_amount
+                    + gateway_reward.witness_amount;
+                (
+                    total_gateway_reward,
```
jeffgrunewald, line 394: does including this total serve any purpose other than testing? wouldn't it be cleaner to just grab the reward message and sum the three reward categories in the test if that's the only place it's used?

iot_verifier/tests/rewarder_operations.rs
```diff
@@ -0,0 +1,35 @@
+mod common;
+use chrono::{Duration as ChronoDuration, Utc};
+use iot_verifier::{reward_share, rewarder};
+use rust_decimal::{prelude::ToPrimitive, Decimal, RoundingStrategy};
+use rust_decimal_macros::dec;
+use sqlx::PgPool;
+
+#[sqlx::test]
+async fn test_operations(_pool: PgPool) -> anyhow::Result<()> {
```
jeffgrunewald, line 9: this `PgPool` is not used? if this test isn't an impl of a trait that requires it is this argument not needed anymore?

iot_verifier/src/rewarder.rs
```diff
@@ -1,11 +1,14 @@
...
+    reward_share::{self, GatewayShares},
     telemetry,
 };
 use chrono::{DateTime, Duration, TimeZone, Utc};
 use db_store::meta;
 use file_store::{file_sink, traits::TimestampEncode};
 use futures::future::LocalBoxFuture;
+use helium_proto::services::poc_lora as proto;
+use helium_proto::services::poc_lora::iot_reward_share::Reward as ProtoReward;
+use helium_proto::services::poc_lora::{UnallocatedReward, UnallocatedRewardType};
```
jeffgrunewald, line 11:
```suggestion
use helium_proto::services::poc_lora as proto;
use helium_proto::services::poc_lora::iot_reward_share::Reward as ProtoReward;
use helium_proto::services::poc_lora::{UnallocatedReward, UnallocatedRewardType};
```

and then reference throughout as `proto::iot_reward_share::Reward`, `proto::UnallocatedReward`, `proto::UnallocatedRewardType`

iot_verifier/src/reward_share.rs
```diff
@@ -216,6 +222,10 @@ impl RewardShares {
 #[derive(Default)]
 pub struct GatewayShares {
     pub shares: HashMap<PublicKeyBinary, RewardShares>,
+    pub beacon_rewards_per_share: Decimal,
+    pub witness_rewards_per_share: Decimal,
+    pub dc_transfer_rewards_per_share: Decimal,
+    pub total_rewards_for_poc_and_dc: Decimal,
```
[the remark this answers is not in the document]
jeffgrunewald: are these necessary when you can just call the `total_shares()` method on a `GatewayShares` struct that returns a 3-tuple of (Decimal, Decimal, Decimal) for the total of each category and then you just sum those three and subtract from total available to get the unallocated amount? seems less intensive than keeping the running total of each one inside when you can just call that method once at the end

iot_verifier/src/reward_share.rs
andymck: this is used in the rewarder, to accumulate the final u64 value of each gateway reward and from this to work out the final unallocated reward amount

iot_verifier/src/reward_share.rs
andymck: Addressed in sep PR via this commit: d1238dc8e5e255b4aa03f3c4a41fc6da62053b04

iot_verifier/src/reward_share.rs
andymck: Addressed in sep PR via this commit: d1238dc8e5e255b4aa03f3c4a41fc6da62053b04

iot_verifier/src/reward_share.rs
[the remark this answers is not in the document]
andymck: Addressed in sep PR via this commit: d1238dc8e5e255b4aa03f3c4a41fc6da62053b04

iot_verifier/tests/rewarder_operations.rs
andymck: Addressed in sep PR via this commit: d1238dc8e5e255b4aa03f3c4a41fc6da62053b04

iot_verifier/src/rewarder.rs
andymck: Addressed in sep PR via this commit: d1238dc8e5e255b4aa03f3c4a41fc6da62053b04

andymck:
> approved. i still think at a later time i'd like to see the way we calculate total allocated rewards more closely mirror how we calculate the `total_shares()` method on `GatewayRewardShares`, i.e. all at one time and independently of other operations in the same method since the `into_iot_reward_shares` method now seems a bit overloaded computing a total decimal of bones, converting the reward shares to the proto type _and_ writing them to the file sink client.

`total_allocated_rewards` is accumulated as part of the into_iot_reward_shares iterator in order to avoid having to make a third or distrinct pass over the shares map.  `into_iot_reward_shares` generates the reward proto after converting the reward decimal value to a u64 value.  This conversion is done as we cannot reward fractional bones.  Both the proto definition and `total_allocated_rewards` require this u64 value, hence why the iterator returns that u64 value. 

Note: `into_iot_reward_shares` It doesnt actually write the proto to the filesink.  That is performed in the rewarder itself.

PR #680, opened 2023-12-05
Fixes for slow HB processing

maplant:


jeffgrunewald (approved):
Not sure about that `inserted_at` method name, but I don’t have anything better that’s not painfully verbose like `fetch_last_valid_timestamp` or something

PR #684, opened 2023-12-06
Add processing time histograms to prevent anomalous processing times

maplant:


mobile_verifier/src/coverage.rs
```diff
@@ -86,7 +87,11 @@ impl CoverageDaemon {
                         tracing::info!("CoverageDaemon shutting down");
                         break;
                     }
-                    Some(file) = self.coverage_objs.recv() => self.process_file(file).await?,
+                    Some(file) = self.coverage_objs.recv() => {
+			let start = Instant::now();
+			self.process_file(file).await?;
+			metrics::histogram!("coverage_object_processing_time", Instant::now() - start);
```
jeffgrunewald, line 93: is there any functional benefit to computing two different `Instant`s and then getting the difference vs. calling the `elapsed()` method on the first instant?

also, i assume the `histogram!` macro is automatically coercing the resulting duration to the necessary `f64`?

jeffgrunewald (approved):
looks good to me; any reason not to move the histogram emissions inside the implementation of each `process_file()`?

maplant:
> looks good to me; any reason not to move the histogram emissions inside the implementation of each `process_file()`?

Yeah; they can return early with errors

PR #685, opened 2023-12-07
Outdoor WiFi radios

maplant:
As far as I can tell, this is the only thing we need to do in order to add WiFi. I assumed the calculations would be different, and indeed they are, but only _after_ mapping events are live. Until then, unless I am mistaken, the trust score calculations for both outdoor and indoor radios are the exact same. Please confirm or refute this for me @andymck 

Based on @bbalser's work

### DIFF
--- a/mobile_verifier/migrations/22_coverage_objects.sql
+++ b/mobile_verifier/migrations/22_coverage_objects.sql
@@ -18,14 +18,15 @@ CREATE TABLE hexes (
     PRIMARY KEY (uuid, hex)
 );
 
-
 INSERT INTO coverage_objects(uuid, radio_type, radio_key, indoor, coverage_claim_time, inserted_at, invalidated_at)
 SELECT distinct uuid, radio_type, radio_key, indoor, coverage_claim_time, inserted_at, invalidated_at
 FROM hex_coverage;
 
 INSERT INTO hexes(uuid, hex, signal_level, signal_power)
 SELECT uuid, hex, signal_level, signal_power
-from hex_coverage;
+FROM hex_coverage;
 
 ALTER TABLE hex_coverage RENAME TO old_hex_coverage;
+
+ALTER TYPE cell_type RENAME VALUE 'novagenericwifiindoor' TO 'novagenericwifi';
 
--- a/mobile_verifier/migrations/22_coverage_objects.sql
+++ b/mobile_verifier/migrations/22_coverage_objects.sql
@@ -28,5 +28,4 @@ FROM hex_coverage;
 
 ALTER TABLE hex_coverage RENAME TO old_hex_coverage;
 
-ALTER TYPE cell_type RENAME VALUE 'novagenericwifiindoor' TO 'novagenericwifi';
-
+ALTER TYPE cell_type ADD VALUE 'novagenericwifioutdoor' AFTER 'novagenericwifiindoor';
--- a/mobile_verifier/migrations/22_coverage_objects.sql
+++ b/mobile_verifier/migrations/22_coverage_objects.sql
@@ -19,7 +19,7 @@ CREATE TABLE hexes (
 );
 
 INSERT INTO coverage_objects(uuid, radio_type, radio_key, indoor, coverage_claim_time, inserted_at, invalidated_at)
-SELECT distinct uuid, radio_type, radio_key, indoor, coverage_claim_time, inserted_at, invalidated_at
+SELECT DISTINCT uuid, radio_type, radio_key, indoor, coverage_claim_time, inserted_at, invalidated_at
 FROM hex_coverage;
 
 INSERT INTO hexes(uuid, hex, signal_level, signal_power)
--- a/mobile_verifier/migrations/22_coverage_objects.sql
+++ b/mobile_verifier/migrations/22_coverage_objects.sql
@@ -1,13 +1,12 @@
 CREATE TABLE coverage_objects (
-    uuid UUID NOT NULL,
+    uuid UUID PRIMARY KEY,
     radio_type radio_type NOT NULL,
     radio_key TEXT NOT NULL,
     indoor BOOLEAN NOT NULL,
     coverage_claim_time TIMESTAMPTZ NOT NULL,
     trust_score INTEGER,
     inserted_at TIMESTAMPTZ NOT NULL,
-    invalidated_at TIMESTAMPTZ,
-    PRIMARY KEY (uuid)
+    invalidated_at TIMESTAMPTZ
 );
 
 CREATE TABLE hexes (
--- a/mobile_verifier/src/cell_type.rs
+++ b/mobile_verifier/src/cell_type.rs
@@ -20,7 +20,7 @@ pub enum CellType {
     SercommIndoor = 3,
     SercommOutdoor = 4,
     CellTypeNone = 5,
-    NovaGenericWifiIndoor = 6,
+    NovaGenericWifi = 6,
 }
 
 #[derive(PartialEq)]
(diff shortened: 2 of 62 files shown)


mobile_verifier/src/cell_type.rs
```diff
@@ -49,8 +49,21 @@ impl CellType {
...
+    }
+
+    pub fn to_proto(&self, indoor: bool) -> CellTypeProto {
+        match self {
+            Self::Nova436H => CellTypeProto::Nova436h,
+            Self::Nova430I => CellTypeProto::Nova430i,
+            Self::Neutrino430 => CellTypeProto::Neutrino430,
+            Self::SercommIndoor => CellTypeProto::SercommIndoor,
+            Self::SercommOutdoor => CellTypeProto::SercommOutdoor,
+            Self::NovaGenericWifi if indoor => CellTypeProto::NovaGenericWifiIndoor,
```
andymck, line 64: Why have only a single `NovaGenericWifi` covering both indoor and outdoor for the internal struct but distinct values on the proto struct ?  I suspect its because we dont want to change the proto as its already out in the wild ?  

In general I dont really see the benefit in the change  to a single value, especially if the external representations use distinct values ( ie the protos and the onchain representations ).  It also requires any code which needs to distinguish between indoor and outdoor to rely on the coverage object being present/available...that may be a safe assumption ( not entirely sure yet ) but doesnt seem ideal.

mobile_verifier/tests/heartbeats.rs
```diff
@@ -281,18 +281,18 @@ async fn ensure_wifi_hotspots_are_rewarded(pool: PgPool) -> anyhow::Result<()> {
...
-    ($1, 'novagenericwifiindoor', '2023-08-25 03:00:00+00', '2023-08-25 03:00:00+00', NOW(), 300, $2),
-    ($1, 'novagenericwifiindoor', '2023-08-25 04:00:00+00', '2023-08-25 04:00:00+00', NOW(), 300, $2),
-    ($1, 'novagenericwifiindoor', '2023-08-25 05:00:00+00', '2023-08-25 05:00:00+00', NOW(), 300, $2),
-    ($1, 'novagenericwifiindoor', '2023-08-25 06:00:00+00', '2023-08-25 06:00:00+00', NOW(), 300, $2),
-    ($1, 'novagenericwifiindoor', '2023-08-25 07:00:00+00', '2023-08-25 07:00:00+00', NOW(), 300, $2),
-    ($1, 'novagenericwifiindoor', '2023-08-25 08:00:00+00', '2023-08-25 08:00:00+00', NOW(), 300, $2),
-    ($1, 'novagenericwifiindoor', '2023-08-25 09:00:00+00', '2023-08-25 09:00:00+00', NOW(), 300, $2),
-    ($1, 'novagenericwifiindoor', '2023-08-25 10:00:00+00', '2023-08-25 10:00:00+00', NOW(), 300, $2),
-    ($1, 'novagenericwifiindoor', '2023-08-25 11:00:00+00', '2023-08-25 11:00:00+00', NOW(), 300, $3)
+    ($1, 'novagenericwifi', '2023-08-25 00:00:00+00', '2023-08-25 00:00:00+00', NOW(), 300, $2),
```
andymck, line 284: Not solely related to this PR but noting it here...where possible we should avoid using raw SQL in integration tests as part of seeding and instead use the existing struct methods ie ValidatedHeartbeat.save().  Means we exercise more of the actual codebase in the tests, reduces the amount of changes to tests should a schema change and forces a cleaner and more robust implementation of the test itself

andymck (approved):

andymck:
@maplant from my reading of the HIP, the location trust scores for indoor vs outdoor are different.  Indoor has either 0.25 or 1.0 ( prior to mappers going live ) whilst outdoor should be 1.0 if it has asserted its location ( prior to mappers going live ).  The existing trust score uses the location validation timestamp and the distance to asserted.  I am not sure if the same factors should be used for outdoor, that should be verified with Mario.

mobile_verifier/migrations/22_coverage_objects.sql
```diff
@@ -0,0 +1,32 @@
+CREATE TABLE coverage_objects (
+    uuid UUID NOT NULL,
+    radio_type radio_type NOT NULL,
+    radio_key TEXT NOT NULL,
+    indoor BOOLEAN NOT NULL,
+    coverage_claim_time TIMESTAMPTZ NOT NULL,
+    trust_score INTEGER,
```
jeffgrunewald, line 7: isn't trust score a float? or is the intention to store it as a whole number multiple of the functional trust score?

mobile_verifier/migrations/22_coverage_objects.sql
```diff
@@ -0,0 +1,32 @@
...
+    inserted_at TIMESTAMPTZ NOT NULL,
+    invalidated_at TIMESTAMPTZ,
+    PRIMARY KEY (uuid)
```
jeffgrunewald, line 10:
```suggestion
```

mobile_verifier/migrations/22_coverage_objects.sql
```diff
@@ -0,0 +1,32 @@
+CREATE TABLE coverage_objects (
+    uuid UUID NOT NULL,
```
jeffgrunewald, line 2:
```suggestion
    uuid UUID PRIMARY KEY NOT NULL,
```

mobile_verifier/migrations/22_coverage_objects.sql
```diff
@@ -0,0 +1,32 @@
...
+SELECT DISTINCT uuid, radio_type, radio_key, indoor, coverage_claim_time, inserted_at, invalidated_at
+FROM hex_coverage;
+
+INSERT INTO hexes(uuid, hex, signal_level, signal_power)
+SELECT uuid, hex, signal_level, signal_power
+FROM hex_coverage;
+
+ALTER TABLE hex_coverage RENAME TO old_hex_coverage;
+
+ALTER TYPE cell_type RENAME VALUE 'novagenericwifiindoor' TO 'novagenericwifi';
```
jeffgrunewald, line 31: agree with mack, this seems contrary to the intention of the cell type enum in the proto and potentially limiting; given how quickly the cbrs and wifi mobile hotspots are diverging and the differences in the way indoor and outdoor wifi radios trust scores and other coverage factors are considered it seems like maintaining a distinction is valuable.

mobile_verifier/src/cell_type.rs
maplant: I think that's reasonable, I'll change around the order

bbalser: I am not sure I understand why we care about cell_type anymore.  Isn't it just cbrs or wifi and indoor or outdoor?

mobile_verifier/migrations/22_coverage_objects.sql
bbalser: we receive it as an integer so I was just planning on storing it as one and convert to float when needed.

maplant: :+1:

mobile_verifier/src/cell_type.rs
maplant: We need it for our proto output.
I think it generally cleans up the implementation so I'm fine doing it.

mobile_verifier/src/coverage.rs
```diff
@@ -601,31 +636,36 @@ impl CoverageClaimTimeCache {
...
 }
 
-impl CoveredHexCache {
+impl CoverageObjects {
     pub fn new(pool: &Pool<Postgres>) -> Self {
         Self { pool: pool.clone() }
     }
 
-    pub async fn inserted_at(
+    pub async fn coverage_info(
```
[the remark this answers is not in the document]
maplant: I'll go with `CoverageObjectSummary`

mobile_verifier/src/reward_shares.rs
```diff
@@ -955,7 +955,7 @@ mod test {
             HeartbeatRow {
```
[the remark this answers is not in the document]
maplant: I have just received a mandate from heaven that the calculation for both should be the same, so I'll probably just duplicate one of the tests and mark everything as outdoor

bbalser (approved):

andymck (approved):

PR #688, opened 2023-12-14
include device type in mobile gateway info

jeffgrunewald:


andymck:
why the dup pr, when there is #632 ( albeit requiring updating )?

jeffgrunewald:
> why the dup pr, when there is #632 ( albeit requiring updating )?

This is a pressing need for the app team (builder app needs the data) but @bbalser said we need to reevaluate the logic of the asserted location distance check in addition to the rebase so I wanted to get this change in independently.

andymck:
> > why the dup pr, when there is #632 ( albeit requiring updating )?
> 
> This is a pressing need for the app team (builder app needs the data) but @bbalser said we need to reevaluate the logic of the asserted location distance check in addition to the rebase so I wanted to get this change in independently.

gotcha

mobile_config/src/gateway_info.rs
```diff
@@ -1,34 +1,46 @@
...
 pub struct GatewayMetadata {
-    pub location: u64,
+    pub device_type: DeviceType,
+    pub location: Option<u64>,
 }
 
 #[derive(Clone, Debug)]
 pub struct GatewayInfo {
     pub address: PublicKeyBinary,
-    pub metadata: Option<GatewayMetadata>,
```
andymck, line 17: so it used to be that if metadata was `Some` then all its fields would be available as they were populated upon assertion ( convention carried across from iot config )  Now its up to each field to be optional where applicable.  This was somewhat the reason I moved device_type out of metadata so as to maintain the existing rule but I dont mind this approach of optional sub fields either...the only thing I would note is that it takes it into a different convention that iot configs metadata

andymck (approved):

bbalser (approved):

mobile_config/src/gateway_info.rs
madninja: yeah the inconsistency is not ideal here..

jeffgrunewald: @andymck 's half-convincing me here. i can go either way and it's easy to flip it back if we prefer it. Semantically I think it makes some sense since for the purposes of Oracles an IoT hotspot _has_ to be asserted to be rewarded whereas that field is literally optional in a mobile hotspot for the purposes of the mobile poc.

On the other hand, unlike an asserted location, which can changes over time, the device type is intrinsic to the mobile hotspot forever, so in that sense it could easily be viewed similar to the `is_full_hotspot` boolean that is part of the main IoT hotspot info struct.

Lemme know your preferences @madninja and @bbalser ; arguably reasons to put it in either place and seems like the Options might be slightly easier handled if it _wasn't_ in the metadata sub-struct but then it doesn't "match" the way the info is stored on-chain. For reference is the iot hotspot's `is_full_hotspot` boolean in the metadata for those devices?

PR #690, opened 2023-12-15
do not stream updates for routes that have been deleted

michaeldjeffrey:
In a scenario such as:

1. Add Route
2. Add SKF
3. Remove Route

If a stream is created requesting updates since before the first action, it will receive:

1. Remove Route
2. Add SKF

With these changes, the route removal will be streamed, allowing HPR to clean up related configuration, and the Adds afterwards will be ignored.

### DIFF
--- a/iot_config/src/route.rs
+++ b/iot_config/src/route.rs
@@ -560,7 +560,8 @@ pub fn eui_stream<'a>(
         r#"
         select eui.route_id, eui.app_eui, eui.dev_eui, eui.deleted
         from route_eui_pairs eui
-        where eui.updated_at >= $1
+        join routes r on eui.route_id = r.id
+        where eui.updated_at >= $1 and r.deleted = false
         "#,
     )
     .bind(since)
@@ -578,7 +579,8 @@ pub fn devaddr_range_stream<'a>(
         r#"
         select devaddr.route_id, devaddr.start_addr, devaddr.end_addr, devaddr.deleted
         from route_devaddr_ranges devaddr
-        where devaddr.updated_at >= $1
+        join routes r on devaddr.route_id = r.id
+        where devaddr.updated_at >= $1 and r.deleted = false
         "#,
     )
     .bind(since)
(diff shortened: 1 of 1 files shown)


jeffgrunewald (commented):
i think i understand how this avoids the problem you've outlined in the PR description but what mechanism ensures that the SKF, EUI, and DevAddr records are cleaned up now that the records relative to the route are being omitted from the stream?

when the HPRs had fully ephemeral state, if you deleted a route all of its child-records would be auto cleaned up no later than the next reboot of the HPR but now when the LNS operator deletes a route the child records will just stay in the HPR unless the LNS operator explicitly deletes the child records as well.

this further compounds the HPR's stale data problem by the fact that all of those child records automatically purge from the config service DB when their parent route is garbage collected after 2 days. so the CS garbage collector deletes the "deleted" route, the delete cascades to the child records and they're removed from the DB but the HPR still has them saved unless they're explicitly deleted via the RPC.

would it not be better to have a `route_delete` operation involve a cascade `child_record_delete` that applies the deleted boolean to all of the child records and thus sends a stream update to HPR that all of those child records have been marked for removal as well? then they would _all_ get garbage collected after the 2 day tombstone marker

michaeldjeffrey:
Route cleanup of child records is also a part of HPR. 

Route deletes happening after the timestamp are still included in the route portion of the stream. Related DevAddrs, EUIs, and SKFs will be removed at that point.

An HPR could be down for more than 2 days and miss a route delete. At that point it would make sense to completely reset the HPR and start from the beginning.

If it happens that an HPR is able to delete a Route but not any of the related configuration, packets matching those routes would be ignored as they don't terminate at a Route. In the case of SKFs where it stores a unique dets file, that file would not be rehydrated into consideration for routing. If that continued, file system limits would alarm and we would then make the decision to up the storage limit on a server, or reset configuration from the beginning and find the reason for the leaving behind of the SKF files.

jeffgrunewald:
given that, i don't see a problem here but would it do any harm to the HPR to broadcast the cascade delete of route child records and perhaps have the HPR ignore them if this problem is already solved on that end? for the moment i think HPR is the only authorized consumer of these streams but i worry that this could change in the future and this cascading delete being sort of buried in the database table migration that established the persistent records could be lost on other consumers of the streams. food for thought

jeffgrunewald (approved):

bbalser (approved):

PR #693, opened 2023-12-16
Update file_info_poller to track a process name and make sqlx an optional dependency

bbalser:
- Add process_name, defaulted to 'default', to file_info_poller to allow 2 instances reading the same source data to completely independent of each other.
- Added traits and feature(sqlx-postgres) to make sqlx an optional dependency of the file-store crate


file_store/src/file_info_poller.rs
```diff
@@ -227,91 +265,92 @@ fn create_cache() -> MemoryFileCache {
...
-            Ok(None)
-        } else {
-            Ok(Some(result))
-        }
+                SELECT MAX(file_timestamp) FROM files_processed where process_name = $1 and file_type = $2
+            "#,
+            )
+            .bind(process_name)
+            .bind(file_type)
+            .fetch_one(self)
```
jeffgrunewald, line 319: could this ever get called with _no_ results such that it needs to be `fetch_optional`?

bbalser: No, when then are no rows that match the where clause, calling the max function will return 1 row where the value is NULL hence mapping the single value to a `<Option<DateTime<Utc>>>`

andymck (approved):

file_store/src/file_source.rs
```diff
@@ -8,11 +8,11 @@ use std::path::{Path, PathBuf};
 use tokio::{fs::File, io::BufReader};
 use tokio_util::codec::{length_delimited::LengthDelimitedCodec, FramedRead};
 
-pub fn continuous_source<T>() -> FileInfoPollerConfigBuilder<T>
+pub fn continuous_source<T, S>() -> FileInfoPollerConfigBuilder<T, S>
```
maplant, line 11: I would provide S as an argument to continuous source, since it's necessary

maplant (approved):

file_store/Cargo.toml
```diff
@@ -57,3 +57,4 @@ tempfile = "3"
 
 [features]
 local = ["aws-types"]
+sqlx-postgres = ["sqlx"]
```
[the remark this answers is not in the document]
bbalser: Fixed ed2b337

PR #694, opened 2023-12-19
Update main query for mobile rewards to simplify query and location t…

bbalser:
…rust score calculation


mobile_verifier/src/heartbeats/valid_radios.sql
```diff
@@ -0,0 +1,101 @@
...
+            AND truncated_timestamp < $2
+        ORDER BY
+            hotspot_key,
+            truncated_timestamp DESC))
+SELECT
+    hb.hotspot_key,
+    hb.cbsd_id,
+    hb.cell_type,
+    hb.location_trust_multiplier,
+    u.uuid AS coverage_object
```
jeffgrunewald, line 93: why alias this column to uuid in the CTE and then revert that to the original column name here?

mobile_verifier/src/heartbeats/valid_radios.sql
```diff
@@ -0,0 +1,101 @@
...
+        AND truncated_timestamp < $2
+    ORDER BY
+        cbsd_id,
+        latest_timestamp DESC
+),
+heartbeats AS (
+    SELECT
+        lch.hotspot_key,
+        ch.cbsd_id,
+        ch.cell_type,
```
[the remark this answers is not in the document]
bbalser: This PR is not deprecating cell_type, so here I was just returning the field in the query that the code already expects to receive

mobile_verifier/src/heartbeats/valid_radios.sql
bbalser: it made sense to my brain at the time, but yeah, doesn't make sense

Fixed 622aed2

PR #696, opened 2023-12-21
process all cbrs heartbeats, remove filtering

andymck:
Removes the filtering of CBRS heartbeats to discard a HB when we had already seen a HB for a given hour.  This results in all cbrs HBs being processed and validated and the results being outputted to S3.

The filtering was added as a temp measure due to slow processing on mainnet.  That slowness was properly addressed by a refactor of DB usage as part of #685.  

Local performance testing resulted the following average times for the processing of each cbrs HB ingest file:

With filtering:  14 seconds
Without filtering: 24 seconds

24 seconds without filtering is an acceptable time and results in the benefit of all HBs being processed

### DIFF
--- a/mobile_verifier/src/heartbeats/cbrs.rs
+++ b/mobile_verifier/src/heartbeats/cbrs.rs
@@ -94,7 +94,7 @@ where
     async fn process_file(
         &self,
         file: FileInfoStream<CbrsHeartbeatIngestReport>,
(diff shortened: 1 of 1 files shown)


maplant (approved):

maplant:
Fantastic! Thanks for looking into this @andymck!

bbalser (approved):

PR #697, opened 2024-01-03
ics returns responses with millisecond timestamps

jeffgrunewald:
per the iot-config service proto definitions, the expected input timestamps for rpc requests are in millisecond precision but the responses are in second precision. this pushes the two into alignment so responses are now also in millisecond precision.

doesn't require <https://github.com/helium/proto/pull/383> as the data type from the perspective of the proto definitions is unchanged but should be aligned shortly to provide clarity for rpc consumers


iot_config/tests/route_service.rs
```diff
@@ -195,7 +195,7 @@ async fn stream_only_sends_data_modified_since(pool: Pool<Postgres>) {
     let response = client
         .stream(route_stream_req_v1(
             &client_keypair,
-            since.timestamp_millis() as u64,
+            since.timestamp() as u64,
```
andymck, line 198: the description says responses will now be in millisecs, but yet this and others are changed to secs ?

andymck (approved):

michaeldjeffrey (approved):

ccall48:
Sorry I missed this...

Just to clarify, second and millisecond precision will both be accepted or do I have to recode all the iot rpc's to be second precision now?

ccall48:
Ok thanks Jeff, i was already sending them with signed time as millisecond precision. if i can switch all these to be second precision before this is enforced on the iot side this should be fine.

Thank you for the clarification.

PR #698, opened 2024-01-03
Reward granularity

maplant:


mobile_verifier/src/reward_shares.rs
```diff
@@ -121,10 +141,15 @@ impl TransferRewards {
...
                         hotspot_key: hotspot_key.into(),
-                        dc_transfer_reward: (reward * reward_scale)
+                        dc_transfer_reward: (reward.bones * reward_scale)
                             .round_dp_with_strategy(0, RoundingStrategy::ToZero)
                             .to_u64()
-                            .unwrap_or(0),
+                            .unwrap_or_default(),
+                        rewardable_bytes: reward.bytes_rewarded,
+                        // Can someone confirm what this should be?
+                        price: (self.mobile_bone_price * dec!(1_000_000))
```
andymck, line 150: i think you are going to need to multiply that again, to get it the mobile token price at 10^6 otherwise you will still have a value < 0 at the point of converting to a float and thus end up with a zero price value, so basically:

price: (self.mobile_bone_price * dec!(1_000_000) * dec!(1_000_000)

This will take you back to Mobile price @ 10^6 which fits with the u64 proto definition

mobile_verifier/src/data_session.rs
```diff
@@ -146,7 +152,9 @@ pub async fn data_sessions_to_dc<'a>(
     tokio::pin!(stream);
     let mut map = HotspotMap::new();
     while let Some(session) = stream.try_next().await? {
-        *map.entry(session.pub_key).or_default() += session.num_dcs as u64
+        let rewards = map.entry(session.pub_key).or_default();
+        rewards.rewardable_dc += session.num_dcs as u64;
+        rewards.rewardable_bytes += session.upload_bytes as u64 + session.download_bytes as u64;
```
andymck, line 157: Should we not be carrying rewardable_bytes from the proto:ValidDataTransferSession into this session object and use that to accumulate rewardable_bytes rather than the older upload and download bytes values ?

mobile_verifier/src/heartbeats/mod.rs
```diff
@@ -414,16 +415,27 @@ impl ValidatedHeartbeat {
         })
     }
 
-    pub async fn write(&self, heartbeats: &FileSinkClient) -> file_store::Result {
+    pub async fn write(
+        &self,
+        max_distance_to_asserted: u32,
+        heartbeats: &FileSinkClient,
+    ) -> file_store::Result {
+        let location_trust_score_multiplier = self.cell_type.location_weight(
```
bbalser, line 423: Maybe we should persist the `location_trust_score_multiplier` into the database.  That way we can ensure that the average calculation uses the same score that was recorded with the heartbeat.

mobile_verifier/src/data_session.rs
bbalser: I added rewardable_bytes to the `gateway_reward` proto, I was thinking just to make it clearer how many bytes were considered rewardable for the gateway

mobile_verifier/src/heartbeats/mod.rs
```diff
@@ -414,16 +415,27 @@ impl ValidatedHeartbeat {
...
                 proto::Heartbeat {
                     cbsd_id: self.heartbeat.cbsd_id.clone().unwrap_or_default(),
                     pub_key: self.heartbeat.hotspot_key.as_ref().into(),
-                    reward_multiplier: 1.0,
                     cell_type: self.cell_type as i32,
                     validity: self.validity as i32,
                     timestamp: self.heartbeat.timestamp.timestamp() as u64,
+                    location_trust_score_multiplier: (location_trust_score_multiplier * dec!(1000))
+                        .to_u32()
+                        .unwrap_or_default(),
```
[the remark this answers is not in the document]
maplant: It's zero, but this being zero would also indicate an error condition

bbalser (approved):

mobile_verifier/src/data_session.rs
andymck: further note on using rewardable_bytes vs upload + download bytes....cant the sum of upload+download bytes exceed that rewardable...hence why rewardable bytes was added ?

mobile_verifier/src/heartbeats/mod.rs
```diff
@@ -328,37 +327,186 @@ impl ValidatedHeartbeat {
...
+                heartbeat,
+                CellType::CellTypeNone,
+                dec!(0),
+                None,
+                None,
+                proto::HeartbeatValidity::NoSuchCoverageObject,
+            ));
+        };
+
+        let cell_type = match heartbeat.hb_type {
```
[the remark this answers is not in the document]
maplant: It's not deprecated right now, that's will be a separate PR

mobile_verifier/src/heartbeats/mod.rs
```diff
@@ -328,37 +327,186 @@ impl ValidatedHeartbeat {
...
+                dec!(0),
+                None,
+                Some(coverage_summary),
+                proto::HeartbeatValidity::GatewayNotFound,
+            )),
+            GatewayResolution::GatewayNotAsserted if heartbeat.hb_type == HbType::Wifi => {
+                Ok(Self::new(
+                    heartbeat,
+                    cell_type,
+                    dec!(0),
```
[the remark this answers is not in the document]
maplant: The heartbeat is invalid, and therefore it doesn't have a trust score. I figure best approach would be set it to the default value

mobile_verifier/src/data_session.rs
maplant: It's also just more transparency - the rewardable_dc calculation uses a div_ceil operation, so the function is not exactly one to one

PR #705, opened 2024-01-09
constrain http

lthiery:
API changes after 0.2 breaks things


andymck (approved):
In the medium term we should look at updating our code to be compatible with the updated apis

bbalser (approved):

PR #716, opened 2024-01-22
hex boosting support

andymck:
proto PR: https://github.com/helium/proto/pull/389

- [x] update mobile config to use the hex boosting sql & pull from table ( waiting to see table schema )
- [x] refactor poc tests
- [x] add test where radio gets boosts from multiple hexes
- [x] add test where reward period is beyond end of boost end timestamp
- [x] maybe move boost structs from mobile config to filestore
- [x] rename some structs 
- [x] boost man tests
- [x] consider location trust score for wifi radios when rewarding boosted hex ( hip 93 )
- [x] add test to cover reduced location trust score
- [ ] consider activated hex table cleanup - how long we we want DB data to hang around after successfully boosting a hex
- [ ] tidy up

### DIFF
--- a/file_store/src/hex_boost.rs
+++ b/file_store/src/hex_boost.rs
@@ -1,0 +1,9 @@
+use chrono::{DateTime, Utc};
+
+#[derive(Debug, Clone)]
+pub struct BoostedHexActivation {
+    pub location: u64,
+    pub activation_ts: DateTime<Utc>,
+    pub boosted_hex_pubkey: String,
+    pub boost_config_pubkey: String,
+}
--- a/mobile_verifier/src/boosted_hexes.rs
+++ b/mobile_verifier/src/boosted_hexes.rs
@@ -1,0 +1,41 @@
+use chrono::{DateTime, Utc};
+use futures::StreamExt;
+use mobile_config::{
+    boosted_hex_info::BoostedHexInfo,
+    client::{hex_boosting_client::HexBoostingInfoResolver, ClientError},
+};
+use std::collections::HashMap;
+
+#[derive(Debug, Clone, Default)]
+pub struct BoostedHexes {
+    pub hexes: HashMap<u64, BoostedHexInfo>,
+}
+
+#[derive(PartialEq, Debug, Clone)]
+pub struct BoostedHex {
+    pub location: u64,
+    pub multiplier: u32,
+}
+
+impl BoostedHexes {
+    pub async fn new(
+        hex_service_client: &impl HexBoostingInfoResolver<Error = ClientError>,
+    ) -> anyhow::Result<Self> {
+        tracing::info!("getting boosted hexes");
+        let mut map = HashMap::new();
+        let mut stream = hex_service_client
+            .clone()
+            .stream_boosted_hexes_info()
+            .await?;
+        while let Some(info) = stream.next().await {
+            map.insert(info.location, info);
+        }
+        Ok(Self { hexes: map })
+    }
+
+    pub fn get_current_multiplier(&self, location: u64, ts: DateTime<Utc>) -> Option<u32> {
+        self.hexes
+            .get(&location)
+            .and_then(|info| info.current_multiplier(ts).ok()?)
+    }
+}
--- a/mobile_verifier/src/cli/reward_from_db.rs
+++ b/mobile_verifier/src/cli/reward_from_db.rs
@@ -8,6 +8,7 @@ use anyhow::Result;
 use chrono::NaiveDateTime;
 use helium_crypto::PublicKey;
 use helium_proto::services::poc_mobile as proto;
+use mobile_config::boosted_hex_info::BoostedHexes;
 use rust_decimal::Decimal;
 use serde_json::json;
 use std::collections::HashMap;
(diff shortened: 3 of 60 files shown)


mobile_verifier/src/cli/reward_from_db.rs
```diff
@@ -38,8 +39,17 @@ impl Cmd {
         let heartbeats = HeartbeatReward::validated(&pool, &epoch);
         let speedtest_averages =
             SpeedtestAverages::aggregate_epoch_averages(epoch.end, &pool).await?;
-        let reward_shares =
-            CoveragePoints::aggregate_points(&pool, heartbeats, &speedtest_averages, end).await?;
+        let boosted_hexes = BoostedHexes {
+            hexes: HashMap::new(),
+        };
```
maplant, line 44: Give BoostedHexes as a default impl and avoid making its members public

mobile_verifier/src/coverage.rs
```diff
@@ -496,10 +498,44 @@ pub async fn clear_coverage_objects(
...
+        signal_level: SignalLevel,
+        coverage_level: IndoorCoverageLevel,
+    ) {
+        self.multiplier = multiplier;
+        self.coverage
+            .entry(signal_level)
+            .or_default()
+            .push(coverage_level)
+    }
+}
```
maplant, line 520: It doesn't make sense to me that you are setting the multiplier in the same place where you add a coverage level. 

```rust
impl CoveredIndoorHex {
    pub fn new(boosted_hex_multiplier: u32) -> Self { ... }
}
```

mobile_verifier/src/boosted_hexes.rs
```diff
@@ -0,0 +1,41 @@
...
+
+#[derive(Debug, Clone, Default)]
+pub struct BoostedHexes {
+    pub hexes: HashMap<u64, BoostedHexInfo>,
+}
+
+#[derive(PartialEq, Debug, Clone)]
+pub struct BoostedHex {
+    pub location: u64,
+    pub multiplier: u32,
```
maplant, line 17: shouldn't this be a Decimal?

mobile_verifier/src/coverage.rs
andymck: Good point, this goes back to an earlier implementation when i was applying the multiplier earlier in the flow. I can remove it from here entirely and use only at the point of going to coverage rewards

mobile_verifier/src/cli/reward_from_db.rs
andymck: was already there, hadnt thought to use it..fixed

mobile_verifier/src/coverage.rs
```diff
@@ -524,18 +562,20 @@ impl CoveredHexes {
...
                     .or_default()
-                    .entry(signal_level)
-                    .or_default()
-                    .push(IndoorCoverageLevel {
-                        radio_key,
-                        seniority_timestamp: coverage_claim_time,
-                        signal_level,
-                        hotspot: hotspot.clone(),
-                    });
+                    .add_coverage(hex_multiplier, signal_level, coverage_level);
```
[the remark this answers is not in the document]
andymck: reverted to original code, no longer applicable

mobile_verifier/src/boosted_hexes.rs
andymck: the onchain boosts values are u8's

mobile_verifier/src/coverage.rs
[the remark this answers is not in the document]
maplant: Nice, this is much much better

mobile_verifier/src/boosted_hexes.rs
maplant: yeah, but you have to multiply this with a decimal somewhere, and that requires converting it to a decimal. So my thinking is that you should convert this to a decimal while it is being used by mobile verifier, and convert it back when you're writing to a protobuf

maplant (approved):
This looks good from my view _on the mobile verifier side_

file_store/src/hex_boost.rs
```diff
@@ -0,0 +1,11 @@
+use chrono::{DateTime, Utc};
+use sqlx::FromRow;
+
+#[derive(Debug, Clone, FromRow)]
```
bbalser, line 4: not sure we should be implementing `FromRow` in a file-store struct.  sqlx is now an optional dependency of the file-store project.

andymck: yeah will address this

boost_manager/settings.toml
```diff
@@ -0,0 +1,50 @@
+log = "boost_manager=info,solana=debug"
```
bbalser, line 1: Did you mean to check in this file?

file_store/src/hex_boost.rs
andymck: addressed here 8bd66049471f667490a9cde0d8b029151301d0b0

bbalser (approved):

mobile_verifier/src/boosted_hexes.rs
andymck: i kept this as is, the input and outputs are all integers and I convert at the point of a singular need to decimal

PR #742, opened 2024-02-22
make valid packets for netid c00053 free

jeffgrunewald:


bbalser:
We should be looking at the free field in the packet report and if that is true, don't burn any dc.  No need for specific netid checks in the packet verifier

bbalser (approved):

andymck (approved):

PR #743, opened 2024-02-26
fail beacons with no witnesses

andymck:
This adds a new beacon verification which requires beacons to have at least one witness.  Any beacon reports which fail this verification will be declared invalid with a newly added invalid reason: `BeaconNoWitnesses`

This is to address the current scenario whereby a gateway which is submitting beacon reports but does not have any witnesses ( be that due to being a lone wolf or it has faulty hardware preventing it from being able to transmit) from being able to receive witnessing rewards.  

Prior to this change the gateways beacon reports would pass regular verifications and then subsequently fail reciprocity as the beacon reciprocity check would return invalid as the beacon had no witnesses.  As the last_beacon timestamp is updated should the beacon pass regular verifications it resulted in a gap whereby the gateway was able to witness and earn rewards for those witnesses, despite it never having a successful beacon ( ie passing both regular verifications + reciprocity ).

With this change the rule that the last beacon timestamp is updated should the beacon pass regular verifications is maintained but the check to enforce at least one witness is moved out of the beacon reciprocity check and into a regular standalone verification.  This results in the gateways described being prevented from being able to earn beaconing or witnessing rewards.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1590,7 +1590,7 @@ checksum = "8c3c1a368f70d6cf7302d78f8f7093da241fb8e8807c05cc9e51a125895a6d5b"
 [[package]]
 name = "beacon"
 version = "0.1.0"
-source = "git+https://github.com/helium/proto?branch=master#c2b1dfa06e6f726ab19cd50444b2a90dd6c841e0"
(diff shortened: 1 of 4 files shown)


andymck:
closed in favour of #744

PR #748, opened 2024-03-01
add ExternalError variant to file_store::error::Error

thehardbits:
Add ExternalError variant to file_store::error::Error. It will simplify external implementation of MsgDecode, StateRecord and State traits.

### DIFF
--- a/file_store/src/error.rs
+++ b/file_store/src/error.rs
@@ -37,6 +37,8 @@ pub enum Error {
     DbError(#[from] sqlx::Error),
     #[error("channel send error")]
     SendError(#[from] tokio::sync::mpsc::error::SendError<()>),
+    #[error("External Error")]
+    ExternalError(#[from] Box<dyn std::error::Error + Send + Sync>),
 }
 
 #[derive(Error, Debug)]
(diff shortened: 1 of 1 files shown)


bbalser (approved):

file_store/src/error.rs
```diff
@@ -37,6 +37,8 @@ pub enum Error {
     DbError(#[from] sqlx::Error),
     #[error("channel send error")]
     SendError(#[from] tokio::sync::mpsc::error::SendError<()>),
+    #[error("External Error")]
```
madninja, line 40: lower case error string to match the rest please

file_store/src/error.rs
```diff
@@ -37,6 +37,8 @@ pub enum Error {
...
+    ExternalError(#[from] Box<dyn std::error::Error + Send + Sync>),
```
madninja, line 41: What _is_ an eternal error? maybe addd a comment here because it's not clear that it's basically an "Other" error

In addition can you help understand what errors you are wrapping? Sometimes it's better to be explicit about your error but I can't tell if that's what should be happening here?

thehardbits: I added comments to the code. Proposed change is emulating the generic Error type for external traits implementations. More clear solution would be change of the traits definition but it may take a lot of code rework.
Wrapped Errors in my case: type conversions errors, sql errors,  wrong timestamps, etc.

madninja (approved):

PR #749, opened 2024-03-01
Adds ingestor and filestore support for hotspot threshold requests

andymck:


ingest/src/server_mobile.rs
```diff
@@ -214,6 +217,51 @@ impl poc_mobile::PocMobile for GrpcServer {
...
+            })?;
+
+        _ = self.radio_threshold_report_sink.write(report, []).await;
+
+        Ok(Response::new(RadioThresholdReportRespV1 {
+            id: timestamp.to_string(),
+        }))
+    }
+
+    async fn submit_invalidated_threshold_report(
```
maplant, line 255: Shouldn't this function also write out to the sink? Why silently drop?

bbalser: This is intentional, just a placeholder.  Another PR will fully support invalidating the radio for boosted rewards

maplant (approved):

PR #750, opened 2024-03-01
IOT price is now on Pyth

bbalser:


price/src/price_generator.rs
```diff
@@ -174,15 +170,7 @@ impl PriceGenerator {
...
-        let price_opt = match if matches!(
-            self.token_type,
-            BlockchainTokenTypeV1::Hnt | BlockchainTokenTypeV1::Mobile
-        ) {
-            // HNT and Mobile prices are currently supported by pyth
-            self.get_pyth_price(key).await
-        } else {
-            self.get_standard_price(key).await
-        } {
+        let price_opt = match self.get_pyth_price(key).await {
```
maplant, line 173: rename `get_pyth_price` to just `get_price`?

maplant (approved):

price/src/price_generator.rs
bbalser: I think it is better as `get_pyth_price`, feels more descriptive to me.

andymck (approved):

PR #760, opened 2024-03-12
Automatically download data sets

maplant:
Creates two new daemons for checking for new data set files and downloading and processing them once their timestamp has passed.  This should be fairly robust, I prevent the download daemon from shutting down in the middle of downloading, but should that occur for whatever reason no errors should be incurred from restart.


mobile_verifier/src/boosting_oracles/mod.rs
```diff
@@ -1,10 +1,305 @@
...
+        loop {
+            #[rustfmt::skip]
+            tokio::select! {
+                biased;
+                _ = shutdown.clone() => {
+                    tracing::info!("CheckForNewDataSetDaemon shutting down");
+                    break;
+                }
+                _ = tokio::time::sleep(poll_duration.to_std()?) => {
+                    let mut new_data_sets = self.store.list(prefix, latest_file_date, None);
```
andymck, line 268: Might it be worth considering when a new dataset is generated and uploaded to S3 ( by whoever and whatever process performs that process ) that a `manifest` is included alongside the dataset.  That would avoid the need to poll S3 and list files in this way and allow a flow using our regular S3 fileinfo poller to be used and thus reusing existing conventions.  This daemon would consume the manifest and based on the manifest data trigger a download of the new dataset file

mobile_verifier/src/boosting_oracles/mod.rs
```diff
@@ -1,10 +1,305 @@
...
+}
+
+// Better name welcome
+pub struct CheckForNewDataSetDaemon {
+    pool: PgPool,
+    store: FileStore,
+    data_set: DataSetType,
+}
+
+impl ManagedTask for CheckForNewDataSetDaemon {
```
andymck, line 225: Why not push the Downloader and CheckForNew daemons out into sep modules ?

mobile_verifier/src/boosting_oracles/mod.rs
```diff
@@ -1,10 +1,305 @@
+//! Module for handling the downloading of new data sets from S3.
+//!
+//! We've split this task into two parts, the CheckForNewDataSetsDaemon and the DataSetDownloaderDaemon.
```
andymck, line 3: Good to explain all this but might be a better home in the PR description

mobile_verifier/src/boosting_oracles/mod.rs
```diff
@@ -1,10 +1,305 @@
...
+//!
+//! But it would be extremely complicated. For one thing, the DataSetDownloaderDaemon needs to update the
+//! data sets at a particular time, specified by the data set's timestamp. Keeping tracking of everything
+//! in one place would make things more prone to bugs, and the implementation becomes a lot simpler.
+
 pub mod assignment;
+pub mod urbanization;
 
 use std::collections::HashMap;
+use std::path::Path;
```
andymck, line 23: can group all these into one use statement

mobile_verifier/src/boosting_oracles/mod.rs
```diff
@@ -1,10 +1,305 @@
...
+                        file.write_all(&bytes).await?;
+                    }
+                    // Set the status to be downloaded
+                    sqlx::query("UPDATE data_sets SET status = 'downloaded' WHERE file_name = $1")
+                        .bind(&latest_data_set.file_name)
+                        .execute(&self.pool)
+                        .await?;
+                }
+
+                // Now that we've downloaded the file, load it into the urbanization data set
```
andymck, line 157: Should all DB writes from this point forward not be wrapped in a txn ?  Commit that at same time as the filestore commit

mobile_verifier/src/boosting_oracles/mod.rs
madninja: Actually this is a great question.. no manifest needed, but why doesn't this just use the normal file poller setup?

andymck: I was assuming this isnt using the normal file poller as the files are too large and we only want to be processing the latest.  Each file replaces what was before, unlike more of our ingested files which are additive...

madninja: but that's just describing how you handle a new file when the poller sees it.. there's nothing additive about the poller itself

andymck: The point i was making is that info poller will auto decode the file and as these things are large we may not want to necessarily do that for every file....

madninja: ah yes.. true.. these hextrees are disk based and shouldn't be read/streamed

mobile_verifier/src/boosting_oracles/mod.rs
```diff
@@ -1,10 +1,305 @@
...
+    pub async fn run(self, shutdown: triggered::Listener) -> anyhow::Result<()> {
+        // Another option I considered instead of polling was to use ENOTIFY, but that seemed
+        // not as good, as it would hog a pool connection.
+        //
+        // We set the poll duration to 30 minutes since that seems fine.
+        let poll_duration = Duration::minutes(30);
+
+        loop {
+            // Find the latest urbanization file
+            tracing::info!("Checking for new data sets");
```
andymck, line 124: Be good to push this chunk of code  to determine and download the file  out to its own function

mobile_verifier/src/boosting_oracles/mod.rs
```diff
@@ -1,67 +1,388 @@
...
+                }
+
+                // Now that we've downloaded the file, load it into the data set
+                data_set.update(Path::new(&path), latest_data_set.time_to_use)?;
+
+                drop(data_set);
+
+                // Update the hexes
+                let boosting_reports = set_oracle_boosting_assignments(
+                    UnassignedHex::fetch_all(&self.pool),
```
bbalser, line 208: Does this mean that every time the mobile-verifier boots it will update all the hexes with the latest dataset assignments?

bbalser: after re-reading I see that it doesn't since if won't find a non processed dataset in the db.  I think we really need to work on making this more readable.

mobile_verifier/src/boosting_oracles/mod.rs
```diff
@@ -1,184 +1,600 @@
...
+
+    fn is_ready(&self) -> bool;
 }
 
-pub struct HexBoostData {
-    urbanized: DiskTreeMap,
-    usa_geofence: Geofence,
-    footfall: DiskTreeMap,
-    landtype: DiskTreeMap,
+pub struct DataSetDownloaderDaemon<T, A, B, C> {
```
andymck, line 37: this code would be more readable if the DataSetDownloaderDaemon struct and its impls were in their own module.  Likewise the DB functions would also be pushed to its own module

mobile_verifier/src/boosting_oracles/mod.rs
```diff
@@ -1,184 +1,600 @@
...
+    }
+}
+
+async fn download_data_set(
+    store: &FileStore,
+    in_file_name: &str,
+    out_path: &Path,
+) -> anyhow::Result<()> {
+    tracing::info!("Downloading new data set: {}", out_path.to_string_lossy());
+    // TODO: abstract this out to a function
```
andymck, line 252: This todo still relevant ? looks not as download data set does what it says on the tin ?

mobile_verifier/src/boosting_oracles/mod.rs
```diff
@@ -1,184 +1,600 @@
...
+            WHERE
+                urbanized IS NULL
+                OR footfall IS NULL
+                OR landtype IS NULL",
+        )
+        .fetch(pool)
+    }
+}
+
+pub async fn set_oracle_boosting_assignments<'a>(
```
andymck, line 495: this function would be more palatable with some elements broken out

mobile_verifier/src/cli/server.rs
```diff
@@ -101,10 +104,63 @@ impl Cmd {
...
+            .await?;
+
+        let urbanization: Urbanization<DiskTreeMap, _> = Urbanization::new(usa_geofence.clone());
+        let footfall: Footfall<DiskTreeMap> = Footfall::new();
+        let landtype: Landtype<DiskTreeMap> = Landtype::new();
+        let hex_boost_data = HexBoostData::new(footfall, landtype, urbanization);
+
+        // Data sets and downloaders
+        let data_sets_file_store = FileStore::from_settings(&settings.data_sets).await?;
+        let footfall_data_set_downloader = DataSetDownloaderDaemon::new(
```
andymck, line 126: Might want to follow the new convention used to setup new daemons as part of https://github.com/helium/oracles/pull/796

mobile_verifier/src/boosting_oracles/mod.rs
maplant: the DB functions are in their own module

mobile_verifier/src/boosting_oracles/mod.rs
maplant: sure, but this function has been in the verifier since urbanization dropped

mobile_verifier/src/boosting_oracles/mod.rs
andymck: Yeah i meant to say push that mod out to its own file

maplant: and I have move the downloader  to a separate module as well

mobile_verifier/src/settings.rs
```diff
@@ -20,6 +20,7 @@ pub struct Settings {
     pub ingest: file_store::Settings,
     pub data_transfer_ingest: file_store::Settings,
     pub output: file_store::Settings,
+    pub data_sets: file_store::Settings,
```
michaeldjeffrey, line 23: Can we get some comments on these settings?
`data_sets` makes some sense in the boostings modules. But just looking at settings, its hard to suss out what type values should be going here.

Same with `data_sets_directory`. 

It seems like these two settings should be closer together. They're more closely related to each other than other settings of the same type.

mobile_verifier/src/rewarder.rs
```diff
@@ -218,6 +219,11 @@ where
                 tracing::info!("No speedtests found past reward period");
                 return Ok(false);
             }
+
+            if check_for_unprocessed_data_sets(&self.pool, reward_period.end).await? {
```
michaeldjeffrey, line 223: This call would benefit from using the qualified name.

mobile_verifier/src/coverage.rs
```diff
@@ -218,9 +202,12 @@ impl CoverageDaemon {
         self.coverage_obj_sink.commit().await?;
         transaction.commit().await?;
 
-        // After writing all of the coverage objects, we set their oracle boosting assignments. This is
-        // done in two steps to improve the testability of the assignments.
-        let unassigned_hexes = UnassignedHex::fetch(&self.pool);
+        // After writing all of the coverage objects, we set their oracle boosting assignments.
+        // If the data sets are not ready, we leave everything as NULL.
+        if !self.hex_boost_data.is_ready().await {
+            return Ok(());
```
michaeldjeffrey, line 208: If it's okay to not do anything when boosting data isn't ready, it makes me wonder if the `CoverageDaemon` needs a copy of `HexBoostData` at all.

mobile_verifier/src/boosting_oracles/mod.rs
```diff
@@ -1,170 +1,110 @@
...
-            other => anyhow::bail!("unexpected landtype disktree value: {other:?}"),
-        };
-        Ok(val)
+pub struct MockDiskTree;
+
+impl DiskTreeLike for MockDiskTree {
+    fn get(&self, cell: hextree::Cell) -> hextree::Result<Option<(hextree::Cell, &[u8])>> {
+        Ok(Some((cell, &[])))
     }
 }
```
michaeldjeffrey, line 109: This appears unused.

mobile_verifier/src/boosting_oracles/mod.rs
```diff
@@ -1,170 +1,110 @@
...
-            Landtype::Water => "Water",
-            Landtype::Wet => "HerbaceousWetland",
-            Landtype::Mangrove => "Mangroves",
-            Landtype::Moss => "MossAndLichen",
-        }
+impl DiskTreeLike for std::collections::HashSet<hextree::Cell> {
+    fn get(&self, cell: hextree::Cell) -> hextree::Result<Option<(hextree::Cell, &[u8])>> {
+        Ok(self.contains(&cell).then_some((cell, &[])))
     }
 }
```
michaeldjeffrey, line 101: This trait appears redundant.
DiskTreeMap already has this interface. No other types are being used in place of a DiskTreeMap.

mobile_verifier/src/boosting_oracles/data_sets.rs
```diff
@@ -0,0 +1,482 @@
...
+        .fetch(pool)
+    }
+}
+
+pub async fn set_oracle_boosting_assignments<'a>(
+    unassigned_hexes: impl Stream<Item = sqlx::Result<UnassignedHex>>,
+    data_sets: &HexBoostData<impl HexAssignment, impl HexAssignment, impl HexAssignment>,
+    pool: &'a PgPool,
+) -> anyhow::Result<impl Iterator<Item = proto::OracleBoostingReportV1>> {
+    const NUMBER_OF_FIELDS_IN_QUERY: u16 = 5;
```
michaeldjeffrey, line 401: This value looks to be from an old version of this function. Can we make sure no updates were missed from what's on master?

mobile_verifier/src/boosting_oracles/data_sets.rs
```diff
@@ -0,0 +1,482 @@
...
+        });
+        Box::pin(
+            handle
+                .map_err(anyhow::Error::from)
+                .and_then(|result| async move { result.map_err(anyhow::Error::from) }),
+        )
+    }
+}
+
+impl<T, A, B, C> DataSetDownloaderDaemon<T, A, B, C>
```
michaeldjeffrey, line 91: Can we get some tests around this?

mobile_verifier/src/boosting_oracles/data_sets.rs
```diff
@@ -0,0 +1,482 @@
...
+            )
+            .await?;
+            tracing::info!(
+                data_set = latest_unprocessed_data_set.filename,
+                "Data set download complete"
+            );
+        }
+
+        // Now that we've downloaded the file, load it into the data set
+        data_set.update(Path::new(&path), latest_unprocessed_data_set.time_to_use)?;
```
michaeldjeffrey, line 182: Should we be checking `time_to_use` here? We could have downloaded a file that's not meant to go live yet.

mobile_verifier/src/boosting_oracles/data_sets.rs
```diff
@@ -0,0 +1,482 @@
...
+
+    fn update(&mut self, path: &Path, time_to_use: DateTime<Utc>) -> anyhow::Result<()>;
+
+    fn is_ready(&self) -> bool;
+}
+
+pub struct DataSetDownloaderDaemon<T, A, B, C> {
+    pool: PgPool,
+    data_set: Arc<Mutex<T>>,
+    data_sets: HexBoostData<A, B, C>,
```
michaeldjeffrey, line 35: Taking a `HexBoostData` _and_ a handle to internal data of that structure to get at `T` seems like a lot of indirection.

mobile_verifier/src/boosting_oracles/mod.rs
michaeldjeffrey: I only wanted to grab the addition side of the diff, sorry.

mobile_verifier/src/boosting_oracles/data_sets.rs
maplant: Nice catch

mobile_verifier/src/rewarder.rs
maplant: as in, keep the `db::` prefix? How qualified are we talking?

mobile_verifier/src/boosting_oracles/data_sets.rs
maplant: the downloader needs to know it's own data set (so that it can replace it when a new one is downloaded), but it also needs all three so that it can re-assign the hexes upon a new download.

mobile_verifier/src/boosting_oracles/assignment.rs
```diff
@@ -58,6 +60,19 @@ impl fmt::Display for Assignment {
...
+        footfall: &impl HexAssignment,
+        landtype: &impl HexAssignment,
+        urbanized: &impl HexAssignment,
+    ) -> anyhow::Result<Self> {
+        Ok(Self {
+            footfall: footfall.assignment(cell)?,
+            landtype: landtype.assignment(cell)?,
+            urbanized: urbanized.assignment(cell)?,
+        })
+    }
```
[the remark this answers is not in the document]
maplant: Agreed. I removed this function, along with `HexBoostData::new` and replaced them with builder patterns. That should hopefully make any bugs more obvious.

mobile_verifier/src/settings.rs
maplant: When you say closer together, do you mean physically fewer lines between them or bundled together in one sub-section?

file_store/src/file_info.rs
```diff
@@ -200,6 +203,9 @@ pub enum FileType {
     RadioThresholdReq,
     RadioThresholdIngestReport,
     VerifiedRadioThresholdIngestReport,
+    UrbanizationDataSet,
```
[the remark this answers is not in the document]
maplant: They have the same file pattern as the others, so I figure best to keep them in there

mobile_verifier/src/boosting_oracles/data_sets.rs
maplant: Unfortunately that's rather difficult right now. We don't really have any way to mock file stores, and a lot of this relies on the file store. 

I will do my best to test the things that are testable.

mobile_verifier/src/coverage.rs
bbalser: Should we just make the mobile-verifier check that it has one of each requried file type and not start the TaskManager until it has the required files ready?

mobile_verifier/src/rewarder.rs
michaeldjeffrey: Hmm. Let's see. 
If there was a `db::` prefixed call here, I would think it was local `rewarder::db::`.

I think my confusion is more around `data_sets`. Nothing about it hints that the rewarder is checking something related to hex boosting. 

I would have a similar confusion if I was looking at the database and saw a plain `data_sets` table.

mobile_verifier/src/settings.rs
michaeldjeffrey: I think both would achieve the same goal. :+1:

mobile_verifier/src/boosting_oracles/data_sets.rs
```diff
@@ -0,0 +1,492 @@
...
+    pub fn to_prefix(self) -> &'static str {
+        match self {
+            Self::Urbanization => "urbanization",
+            Self::Footfall => "footfall",
+            Self::Landtype => "landtype",
+        }
+    }
+}
+
+pub mod db {
```
michaeldjeffrey, line 287: The migration missing from this module. 

Considering the purpose of the table, perhaps a more descriptive name than 'data_sets' could provide clearer context.

maplant: `The migration missing from this module.`
what does this mean?

mobile_verifier/src/boosting_oracles/mod.rs
```diff
@@ -1,162 +1,94 @@
...
-            60 => Landtype::Bare,
-            70 => Landtype::Frozen,
-            80 => Landtype::Water,
-            90 => Landtype::Wet,
-            95 => Landtype::Mangrove,
-            100 => Landtype::Moss,
-            other => anyhow::bail!("unexpected landtype disktree value: {other:?}"),
-        };
-        Ok(val)
+pub struct MockDiskTree;
```
michaeldjeffrey, line 87: This is still unused.

mobile_verifier/src/boosting_oracles/mod.rs
```diff
@@ -1,162 +1,94 @@
...
-            Landtype::Crop => "Cropland",
-            Landtype::Built => "BuiltUp",
-            Landtype::Bare => "BareOrSparseVeg",
-            Landtype::Frozen => "SnowAndIce",
-            Landtype::Water => "Water",
-            Landtype::Wet => "HerbaceousWetland",
-            Landtype::Mangrove => "Mangroves",
-            Landtype::Moss => "MossAndLichen",
-        }
+impl DiskTreeLike for std::collections::HashSet<hextree::Cell> {
```
michaeldjeffrey, line 81: There are no `HashSet`s being used as DiskTrees. 
`DiskTreeLike` can be replaced with `DiskTreeMap` wherever it's used.

mobile_verifier/src/boosting_oracles/mod.rs
```diff
@@ -1,162 +1,94 @@
...
+    Foot: HexAssignment,
+    Land: HexAssignment,
+    Urban: HexAssignment,
+{
+    pub async fn assignments(&self, cell: hextree::Cell) -> anyhow::Result<HexAssignments> {
+        HexAssignments::builder(cell)
+            .footfall(&self.footfall)
+            .landtype(&self.landtype)
+            .urbanized(&self.urbanization)
+            .build()
```
michaeldjeffrey, line 67: Since the locks are gone from the underlying data, this function no longer needs to be `async`, and can be constructed in place removing the need for the a Builder.

```suggestion
    pub fn assignments(&self, cell: hextree::Cell) -> anyhow::Result<HexAssignments> {
        let footfall = self.footfall.assignment(cell)?;
        let landtype = self.landtype.assignment(cell)?;
        let urbanized = self.urbanization.assignment(cell)?;
        Ok(HexAssignments {
            footfall,
            landtype,
            urbanized,
        })
```

mobile_verifier/src/boosting_oracles/data_sets.rs
```diff
@@ -0,0 +1,632 @@
...
+    while let Some(hexes) = unassigned_hexes.try_next().await? {
+        let hexes: anyhow::Result<Vec<_>> = hexes
+            .into_iter()
+            .map(|hex| {
+                let cell = hextree::Cell::try_from(hex.hex)?;
+                let assignments = HexAssignments::builder(cell)
+                    .footfall(&data_sets.footfall)
+                    .landtype(&data_sets.landtype)
+                    .urbanized(&data_sets.urbanization)
+                    .build()?;
```
michaeldjeffrey, line 573:
```suggestion
                let assignments = data_sets.assignments(cell)?;
```

mobile_verifier/src/boosting_oracles/data_sets.rs
```diff
@@ -0,0 +1,632 @@
...
+    A: DataSet,
+    B: DataSet,
+    C: DataSet,
+{
+    fn start_task(
+        self: Box<Self>,
+        shutdown: triggered::Listener,
+    ) -> futures::prelude::future::LocalBoxFuture<'static, anyhow::Result<()>> {
+        let handle = tokio::spawn(async move {
+            #[rustfmt::skip]
```
michaeldjeffrey, line 155: Is there a reason to skip all these `tokio::select!`s?

mobile_verifier/src/boosting_oracles/data_sets.rs
```diff
@@ -0,0 +1,632 @@
...
+            pool,
+            data_sets,
+            store,
+            oracle_boosting_sink,
+            data_set_directory,
+            new_coverage_object_signal,
+        }
+    }
+
+    async fn check_for_new_data_sets(&mut self) -> anyhow::Result<()> {
```
michaeldjeffrey, line 240: I find this iteration of `DataSetDownloaderDaemon` much easier to read through. This is great!

mobile_verifier/src/cli/server.rs
```diff
@@ -101,6 +101,8 @@ impl Cmd {
             settings.usa_and_mexico_fencing_resolution()?,
         )?;
 
+        let (new_coverage_obj_signal_tx, new_coverage_obj_signal_rx) = channel(1);
```
michaeldjeffrey, line 104: This is good. I think it does a good job of clarifying which daemon is responsible for what, and why they care about each other. :+1:

mobile_verifier/src/coverage.rs
```diff
@@ -219,138 +177,13 @@ impl CoverageDaemon {
...
-        let unassigned_hexes = UnassignedHex::fetch(&self.pool);
-        let boosting_reports =
-            set_oracle_boosting_assignments(unassigned_hexes, &self.hex_boost_data, &self.pool)
-                .await?;
-        self.oracle_boosting_sink
-            .write_all(boosting_reports)
-            .await?;
-        self.oracle_boosting_sink.commit().await?;
+        // Tell the data set manager to update the assignments.
+        self.new_coverage_object_signal.try_send(())?;
```
michaeldjeffrey, line 181: Making sure I understand unwrapping a `try_send` here.
If for some reason multiple coverage objects are received while the downloader is downloading, or processing, this will crash the coverage server. Then we'll unwind all the processes through the task manager. But the downloader won't die until it's done with whatever it's downloading/processing, so when we come up, we're good?

mobile_verifier/tests/integrations/common/mod.rs
```diff
@@ -179,35 +180,27 @@ pub fn seconds(s: u64) -> std::time::Duration {
...
+        .landtype(Assignment::A)
+        .build()
+        .unwrap()
+}
+
 type MockAssignmentMap = HashMap<hextree::Cell, Assignment>;
 
-#[derive(Default)]
-pub struct MockHexAssignments {
+#[allow(dead_code)]
```
michaeldjeffrey, line 194: Since https://github.com/helium/oracles/pull/804 landed, you can remove `#[allow(dead_code)]` from this file. 
This function is unused. 
`MockHexBoostingClient` is used.

mobile_verifier/src/geofence.rs
```diff
@@ -1,18 +1,12 @@
...
 
-pub trait GeofenceValidator<T>: Clone + Send + Sync + 'static {
-    fn in_valid_region(&self, t: &T) -> bool;
-}
-
-impl GeofenceValidator<hextree::Cell> for HashSet<hextree::Cell> {
-    fn in_valid_region(&self, cell: &hextree::Cell) -> bool {
-        self.contains(cell)
-    }
+pub trait GeofenceValidator: Clone + Send + Sync + 'static {
```
michaeldjeffrey, line 8: yltsi 💃

mobile_verifier/migrations/33_data_sets.sql
```diff
@@ -0,0 +1,18 @@
...
+       'processed'
+);
+
+CREATE TYPE data_set_type AS enum (
+       'urbanization',
+       'footfall',
+       'landtype'
+);
+
+CREATE TABLE data_sets (
```
michaeldjeffrey, line 13: Can we use a more descriptive name than `data_sets`?
If you're looking only at the database it tells you very little about which modules would use this table, and what type of interactions might be happening to it.

Maybe `hex_boosting_data_set_file_downloading`?

It tells you what type of data it's concerned about. That it's not storing an actual data set, but the status of a file. And the file is being downloaded from somewhere else.

mobile_verifier/src/coverage.rs
maplant: this is incorrect and a remnant of me using `send`, this should just `try_send` and discard any errors.
The reasoning here is that since we've already updated the database, if the channel is full then we will process the new hexes anyway

mobile_verifier/src/boosting_oracles/data_sets.rs
maplant: rustfmt butchers them. it's appalling. It might be fixed at this point tho

mobile_verifier/migrations/33_data_sets.sql
bbalser: maybe `hex_assignment_data_set_status`?? or maybe plural?

andymck (commented):
whilst i do need to take another pass at this...it generally looks good to me...one point I want to note is that the datasets module is somewhat difficult to read through...there is a lot going on in there  I think breaking out its concerns to distinct files would help with this.

mobile_verifier/migrations/33_data_sets.sql
michaeldjeffrey: I like that better :+1:

mobile_verifier/src/boosting_oracles/footfall.rs
```diff
@@ -0,0 +1,70 @@
...
+impl<Foot> HexAssignment for Footfall<Foot>
+where
+    Foot: DiskTreeLike + Send + Sync + 'static,
+{
+    fn assignment(&self, cell: hextree::Cell) -> anyhow::Result<Assignment> {
+        let Some(ref footfall) = self.footfall else {
+            anyhow::bail!("No footfall data set has been loaded");
+        };
+
+        match footfall.get(cell)? {
```
andymck, line 63: some comments around what exactly is being matched here would be good for easy long term reference

mobile_verifier/src/boosting_oracles/assignment.rs
```diff
@@ -97,6 +109,47 @@ impl HexAssignments {
     }
 }
 
+pub struct HexAssignmentsBuilder {
+    cell: hextree::Cell,
+    footfall: Option<Result<Assignment>>,
```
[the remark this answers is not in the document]
maplant: it allows us to carry the errors through to the `build`, rather than earlier. I think it's just slightly cleaner

mobile_verifier/src/boosting_oracles/footfall.rs
maplant: these are called provided functions and can be  overridden, but in this case are not

bbalser (approved):

mobile_verifier/src/boosting_oracles/data_sets.rs
```diff
@@ -0,0 +1,672 @@
...
+                new_landtype.time_to_use,
+            )
+            .await?;
+        }
+
+        Ok(())
+    }
+
+    pub async fn run(mut self) -> anyhow::Result<()> {
+        let poll_duration = Duration::minutes(5);
```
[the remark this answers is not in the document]
maplant: To be honest, I don't really think so. I don't see a situation in which we would want to change whatever value we decide on

mobile_verifier/src/boosting_oracles/data_sets.rs
```diff
@@ -0,0 +1,672 @@
...
+        let Some(first_data_set) = db::fetch_latest_processed_data_set(pool, Self::TYPE).await?
+        else {
+            return Ok(());
+        };
+        let path = get_data_set_path(data_set_directory, Self::TYPE, first_data_set.time_to_use);
+        self.update(Path::new(&path), first_data_set.time_to_use)?;
+        Ok(())
+    }
+
+    async fn check_for_available_data_sets(
```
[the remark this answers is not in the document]
maplant: these are called provided functions and can be overridden, but in this case are not
(accidentally put this on the wrong conversation)

mobile_verifier/src/boosting_oracles/footfall.rs
maplant: :+1:

mobile_verifier/migrations/33_data_sets.sql
maplant: done

andymck (approved):

PR #762, opened 2024-03-13
Center heartbeat latlng to center of containing res12 hex

bbalser:


madninja (approved):

mobile_verifier/src/heartbeats/mod.rs
```diff
@@ -225,10 +225,10 @@ impl Heartbeat {
         }
     }
 
-    pub fn asserted_distance(&self, asserted_location: u64) -> anyhow::Result<i64> {
```
andymck, line 228: wondering why there was no change to the caller to `asserted_distance` but seems there isnt one

andymck (approved):

mobile_verifier/src/heartbeats/mod.rs
bbalser: yeah, function wasn't being used, so I killed it

PR #770, opened 2024-03-20
handle blockhash-not-found errors for mobile packet verifier

andymck:
When submitting a txn via solana, if we hit a blockhash not found error it likely means the blockhash included in the txn is stale...older than 150 blocks.  In this scenario we will now re-sign the txn ( including specifying a current blockhash ) and resubmit immediately.

As the current setup of the calling applications need to take actions when a blockhash not found error is encountered ( such as updating a DB as part of a 2 stage commit ) this retry logic is handled at the application level.  Further then, so as to avoid having retry logic existing at both the application side and then again at the solana lib side, I have dropped the solana lib retry logic and centralised the logic for all retries within each application.

The intention is to revisit this and attempt to centralise the retries within solana but that will require more comprehensive changes.  This interim solution permits the resigning of the stale txns to be deployed and hopefully help to increase the number of successful txn submit attempts during periods of chain congestion.

### DIFF
--- a/boost_manager/src/updater.rs
+++ b/boost_manager/src/updater.rs
@@ -1,7 +1,7 @@
 use crate::db::{self, TxnRow};
 use anyhow::Result;
 use futures::{future::LocalBoxFuture, TryFutureExt};
-use solana::{start_boost::SolanaNetwork, GetSignature};
+use solana::{start_boost::SolanaNetwork, GetSignature, IsErrorBlockhashNotFound};
 use sqlx::{Pool, Postgres};
 use std::time::Duration;
 use task_manager::ManagedTask;
@@ -133,10 +133,7 @@ where
                             .await?;
                         break;
                     }
-                    Err(err)
-                        if self.solana.check_for_blockhash_not_found_error(&err).await
-                            && attempt < MAX_ATTEMPTS =>
-                    {
+                    Err(err) if err.is_error_blockhash_not_found() && attempt < MAX_ATTEMPTS => {
                         tracing::error!("block hash not found..possibly stale block hash, resigning txn and retrying");
                         db::revert_saved_batch_txn_id(&self.pool, &ids).await?;
                         attempt += 1;
--- a/boost_manager/src/updater.rs
+++ b/boost_manager/src/updater.rs
@@ -117,6 +117,7 @@ where
             // get a list of all the activations DB ids which form part of the batch
             let ids: Vec<u64> = batch.iter().map(|sp| sp.location).collect();
             let txn = self.solana.make_start_boost_transaction(batch).await?;
+            let mut signed_txn = self.solana.sign_transaction(&txn).await?;
 
             // if activations were processed successfully then
             // update their status in the DB to success
@@ -126,11 +127,9 @@ where
 
             // retry the sign and submit if we encounter a blockhash not found error
             // all other errors will be returned and exit the retry loop
-            // if we dont have a successful burn by the end of the loop, return an error
             let mut attempt = 1;
             const MAX_ATTEMPTS: u64 = 10;
             loop {
-                let signed_txn = self.solana.sign_transaction(&txn).await?;
                 let transaction_id = signed_txn.get_signature().to_string();
                 db::save_batch_txn_id(&self.pool, &transaction_id, &ids).await?;
                 match self.solana.submit_transaction(&signed_txn).await {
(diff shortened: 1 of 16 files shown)


solana/src/burn.rs
```diff
@@ -45,9 +42,16 @@ pub trait SolanaNetwork: Send + Sync + 'static {
...
+    async fn sign_transaction(
+        &self,
+        transaction: &Self::Transaction,
+    ) -> Result<Self::Transaction, Self::Error>;
+
     async fn submit_transaction(&self, transaction: &Self::Transaction) -> Result<(), Self::Error>;
 
     async fn confirm_transaction(&self, txn: &Signature) -> Result<bool, Self::Error>;
+
+    async fn check_for_blockhash_not_found_error(&self, err: &Self::Error) -> bool;
```
maplant, line 54: You should have the `Error` type satisfy a separate trait. This is a strange way of doing things, for one because self is completely useless. Also, this function does not need to be async

andymck: aha, clearly i misunderstood your prior suggestions.   Addressed better here: 5e532bb4f4477c2566991def9436a2680e846d6f

boost_manager/tests/updater_tests.rs
```diff
@@ -124,14 +138,16 @@ async fn test_process_activations_failure(pool: PgPool) -> anyhow::Result<()> {
...
     )?;
 
     let mut txn = pool.begin().await?;
     seed_activations(&mut txn, now).await?;
     txn.commit().await?;
 
-    // ensure the activations are processed at least 10 times
+    // process the activations
     // submit_txn will bork each time
+    // a failing txn will by default be retried 10 times
```
michaeldjeffrey, line 150: Making sure I understand.
An activation will be processed 10 times before being marked as failed.
And a txn for an activation will be retried 10 times by default.
So this test is "submitting" 110 txns to get the activations into a failed state?

boost_manager/src/updater.rs
```diff
@@ -116,31 +119,49 @@ where
...
+            const MAX_ATTEMPTS: u32 = 10;
+            loop {
+                match self.solana.submit_transaction(&signed_txn).await {
+                    Ok(()) => {
+                        self.handle_submit_txn_success(&ids, batch_size, activations_count)
+                            .await?;
+                        break;
+                    }
+                    Err(err) if err.is_error_blockhash_not_found() && attempt < MAX_ATTEMPTS => {
+                        tracing::error!("block hash not found..possibly stale block hash, resigning txn and retrying");
```
michaeldjeffrey, line 137: Is there any information we can add here about why the blockhash went stale?

boost_manager/tests/updater_tests.rs
michaeldjeffrey: There's a few things I don't know. 
- How long does a blockhash stay valid?
- How often activations are checked for submission (default looks to be 15mins)

Is it possible for an activation to get stuck in a loop to where it never gets marked as Failed in the DB?
It gets processed, but errors, and before the max retries in the db the blockhash cannot be found, so that set of retries gets reset. 
Rinse, repeat.

mobile_packet_verifier/src/burner.rs
```diff
@@ -130,17 +130,55 @@ where
...
+                        .sign_transaction(&txn)
+                        .await
+                        .map_err(BurnError::SolanaError)?;
+
+                    attempt += 1;
+                    continue;
+                }
+                Err(_) if attempt < MAX_ATTEMPTS => {
+                    attempt += 1;
+                    tokio::time::sleep(Duration::from_secs(attempt)).await;
```
michaeldjeffrey, line 174: The boost_manager backs off of a retry_delay with the attempt. Should that also be happening here and in the iot_packet_verifier?

boost_manager/tests/updater_tests.rs
andymck: Yes, an activation will be processed 10 times before being marked as failed.  Each activation attempt the solana txn can be retried up to 10 times.

A blockhash is valid for up to 150 blocks

I believe in mainnet the default activation tick is 15 mins.

I dont see potential for an activation to get stuck in a loop but I will give it a fresh look.  Is there something in particular that suggested to you the potential for that to occur ?

boost_manager/src/updater.rs
andymck: We look for a specific error which is blockhash not found.  This can only mean the blockhash submitted in the txn was incorrect from the first instance ( unlikely ) or it is now stale and it can only go stale if that blockhash is older than 150 blocks ( afaik )

mobile_packet_verifier/src/burner.rs
andymck: yeah it should have been, i will add that in

andymck: addressed here: cdd5a3263e48e0b03ec08c7bdd438babda374fc2

michaeldjeffrey (approved):
I think the blockhash not found retry logic would be a good candidate to move into the solana crate.

andymck:
> I think the blockhash not found retry logic would be a good candidate to move into the solana crate.

yip, as per the pr desc that is the intention but is beyond the scope of this pr

andymck:
closing as no longer on the agenda

PR #775, opened 2024-03-27
Feature/ora 309 mobile packet verifier burn retry

macpie:


mobile_packet_verifier/pkg/settings-template.toml
```diff
@@ -10,6 +10,10 @@ cache = "/var/data/verified-reports"
 # Default value is 1 hour.
 burn_period = 1
 
+# In case burn fails we will accelarate and burn in `min_burn_period` minutes instead of `burn_period` hours.
```
michaeldjeffrey, line 13:
```suggestion
# In case burn fails we will accelerate and burn in `min_burn_period` minutes instead of `burn_period` hours.
```

It feels a bit weird to have 2 different units of time used this close together and only mentioned in the comments. I'm not sure how easy it would be to change both `burn_period_hrs` and `min_burn_period_mins`.

mobile_packet_verifier/src/daemon.rs
```diff
@@ -92,8 +94,15 @@ where
...
-                    self.burner.burn(&self.pool).await?;
-                    burn_time = Instant::now() + self.burn_period;
+                    match self.burner.burn(&self.pool).await {
+                        Ok(_) => {
+                            burn_time = Instant::now() + self.burn_period;
+                        }
+                        Err(e) => {
+                            burn_time = Instant::now() + self.min_burn_period;
+                            tracing::warn!("failed to burn {e:?}, re running burn in {:?} min", self.min_burn_period);
+                        }
```
maplant, line 104: I think this is fine and not really needing of a test, but if you want to you can extract the burn time logic into a function and add a unit test

maplant (approved):

mobile_packet_verifier/pkg/settings-template.toml
```diff
@@ -10,6 +10,10 @@ cache = "/var/data/verified-reports"
 # Default value is 1 hour.
 burn_period = 1
 
+# In case burn fails we will accelerate and burn in `min_burn_period` minutes instead of `burn_period` hours.
```
andymck, line 13: these templates files are no longer in use

mobile_packet_verifier/src/daemon.rs
```diff
@@ -46,6 +47,7 @@ impl<S, GIR, AV> Daemon<S, GIR, AV> {
             burner,
             reports,
             burn_period: Duration::from_secs(60 * 60 * settings.burn_period as u64),
+            min_burn_period: Duration::from_secs(60 * settings.min_burn_period as u64),
```
andymck, line 50: the Duration usage could be moved out to the settings.rs via a helper function to return the final duration.  See the existing helper function `purger_interval` in settings.rs  as an example

michaeldjeffrey (approved):

PR #778, opened 2024-03-28
Use previous location if timestamp is none

maplant:


mobile_verifier/src/heartbeats/mod.rs
```diff
@@ -516,8 +518,43 @@ impl ValidatedHeartbeat {
...
             GatewayResolution::AssertedLocation(location) if heartbeat.hb_type == HbType::Wifi => {
                 let asserted_latlng: LatLng = CellIndex::try_from(location)?.into();
+                let is_valid = match heartbeat.location_validation_timestamp {
+                    None => {
+                        if let Some(last_location) = last_location_cache
+                            .fetch_last_location(&heartbeat.hotspot_key)
+                            .await?
+                        {
+                            if last_location.lat != 0.0 || last_location.lon != 0.0 {
+                                heartbeat.lat = last_location.lat;
```
andymck, line 528: I get the intent with this but the fact that we are rewriting an ingested heartbeat is not ideal.  Typically we should be outputting ingested reports without modification.  Additional data is usually included as part of a wrapper/envelope.    I get that that using an envelope adds additional complexity and maybe not the best way here, so as an alternative should we consider adding a new field/flag to the heartbeat itself to indicate it has been modified ?

mobile_verifier/src/heartbeats/mod.rs
```diff
@@ -741,6 +780,116 @@ pub async fn clear_heartbeats(
...
+            r#"
+            SELECT location_validation_timestamp, lat, lon
+            FROM wifi_heartbeats
+            WHERE location_validation_timestamp IS NOT NULL
+                AND location_validation_timestamp >= $1
+            ORDER BY location_validation_timestamp DESC
+            LIMIT 1
+            "#,
+        )
+        .bind(Utc::now() - Duration::hours(12))
```
andymck, line 819: so the intent here is if the radio has not have a valid location_validation_timestamp within last 12 hours then default it.  But why 12 hours and not 24 or other?

maplant: I don't know, that was the requirement given to me

mobile_verifier/src/heartbeats/mod.rs
```diff
@@ -654,8 +690,8 @@ impl ValidatedHeartbeat {
         let truncated_timestamp = self.truncated_timestamp()?;
         sqlx::query(
             r#"
-            INSERT INTO wifi_heartbeats (hotspot_key, cell_type, latest_timestamp, truncated_timestamp, coverage_object, location_trust_score_multiplier, distance_to_asserted)
-            VALUES ($1, $2, $3, $4, $5, $6, $7)
+            INSERT INTO wifi_heartbeats (hotspot_key, cell_type, latest_timestamp, truncated_timestamp, coverage_object, location_trust_score_multiplier, distance_to_asserted, location_validation_timestamp, lat, lon)
```
andymck, line 693: glad to see location_validation_timestamp getting populated in the DB

mobile_verifier/src/heartbeats/mod.rs
andymck: thinking further on this, i do think a new proto msg might after all be a good option...could be straight forward to use something like 

modified_heartbeat {
  heartbeat original_report = 1,
  heartbeat modified_report = 2,
}

bbalser (approved):

andymck (approved):
Looks good to me, my only concern as previously noted is that we are rewriting an ingested heartbeat to S3.  I get the reasoning

PR #781, opened 2024-04-01
record processing latency for file info streams

jeffgrunewald:
Adds a latency metric to the file info stream recording the amount of time in seconds since a file was created/uploaded to the cloud store to the time it is converted into a stream for processing/ingest by the receiving service.

### DIFF
--- a/file_store/src/file_info_poller.rs
+++ b/file_store/src/file_info_poller.rs
@@ -60,10 +60,10 @@ where
         T: 'static,
     {
         let latency = Utc::now() - self.file_info.timestamp;
-        metrics::histogram!(
+        metrics::gauge!(
             "file-processing-latency",
             latency.num_seconds() as f64,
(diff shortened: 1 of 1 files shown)


Vagabond (approved):
I approve this in spirit but not in implementation

file_store/src/file_info_poller.rs
```diff
@@ -59,6 +59,13 @@ where
     where
         T: 'static,
     {
+        let latency = Utc::now() - self.file_info.timestamp;
+        metrics::histogram!(
+            "file-processing-latency",
+            latency.num_seconds() as f64,
+            "file-type" => self.file_info.prefix.clone()
```
bbalser, line 66: Add process_name as label to the metric

bbalser: and you are sure historgram makes the most sense for this

bbalser (approved):

PR #789, opened 2024-04-13
Landtype Boosting

michaeldjeffrey:
https://github.com/helium/proto/pull/398

Add Landtype to Hex Boosting.

---

### `struct HexAssignments`

`Assignments` are never considered individually, nor should they be. This type packs them together, allowing any code that requires specific use of `Assignment` order to attach itself here.

It also allows `boostring_oracles/mod.rs` to reduce it's reliance on traits. There is now a single exported `BoostedHexAssignments` that allows for getting a `HexAssignments`.

---

### `HexBoostData` and `CoverageDaemon`

With only exporting a single trait from `boosting_oracles/mod.rs`, the `CoverageDaemon` now takes the concrete struct `HexBoostData` with no generic arguments.

Downstream from the `CoverageDaemon`, functions will take `impl BoostedHexAssignments` so they can be used in tests.

---

### `verify-disktree --path <PATH> --type <TYPE>` CLI Command

A CLI command was added to `mobile-verifier`.
It verifies a `.h3tree` file by making sure that every cell can be turned into an `Assignment`.

Current only `Landtype` is supported.

NOTE: In release mode, this function took about 8 minutes to run on my 2.3GHz Intel MacBook Pro for a 28gb file.

```sh
cargo run --release --features "file-store/local" -p mobile-verifier -- -c <CONFIG> verify-disktree --help
```

---

#### Other

- Replaced `Geofence::new` with `Geofence::from_paths` so `new` could used without touching the filesystem.

### DIFF
--- a/mobile_verifier/src/boosting_oracles/assignment.rs
+++ b/mobile_verifier/src/boosting_oracles/assignment.rs
@@ -28,6 +28,19 @@ impl From<Assignment> for i32 {
     }
 }
 
+impl TryFrom<i32> for Assignment {
+    type Error = anyhow::Error;
+
+    fn try_from(value: i32) -> Result<Self, Self::Error> {
+        match value {
+            0 => Ok(Assignment::A),
+            1 => Ok(Assignment::B),
+            2 => Ok(Assignment::C),
+            other => Err(anyhow::anyhow!("could not make Assignment from {other}")),
+        }
+    }
+}
+
 impl fmt::Display for Assignment {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         use Assignment::*;
@@ -59,7 +72,6 @@ pub fn footfall_and_urbanization_multiplier(
     }
 }
 
-#[allow(dead_code)]
 pub fn boosting_oracles_multiplier(
     footfall: Assignment,
     landtype: Assignment,
--- a/mobile_verifier/src/boosting_oracles/assignment.rs
+++ b/mobile_verifier/src/boosting_oracles/assignment.rs
@@ -3,6 +3,13 @@ use rust_decimal::Decimal;
 use rust_decimal_macros::dec;
 use std::fmt;
 
+#[derive(Debug, Clone, PartialEq, Eq, sqlx::FromRow)]
+pub struct HexAssignments {
+    pub footfall: Assignment,
+    pub urbanized: Assignment,
+    pub landtype: Assignment,
+}
+
 #[derive(Copy, Clone, PartialEq, Eq, Debug, sqlx::Type)]
 #[sqlx(type_name = "oracle_assignment")]
 #[sqlx(rename_all = "lowercase")]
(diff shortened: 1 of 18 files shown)


mobile_verifier/src/boosting_oracles/mod.rs
```diff
@@ -1,136 +1,418 @@
...
-        Self {
-            urbanized,
-            usa_geofence,
-        }
+    fn landtype_assignment(&self, cell: hextree::Cell) -> anyhow::Result<Assignment> {
+        let Some((_, vals)) = self.landtype.get(cell)? else {
+            return Ok(Assignment::C);
+        };
+
+        anyhow::ensure!(
```
bbalser, line 81: nice!

mobile_verifier/src/boosting_oracles/assignment.rs
```diff
@@ -28,6 +35,19 @@ impl From<Assignment> for i32 {
     }
 }
 
+impl TryFrom<i32> for Assignment {
```
bbalser, line 38: I would prefer we not match on the i32 value but the enum value
```
impl From<oracle_boosting_hex_assignment::Assignment> for Assignment {
    fn from(value: oracle_boosting_hex_assignment::Assignment) -> Self {
        match value {
            oracle_boosting_hex_assignment::Assignment::A => Self::A,
            oracle_boosting_hex_assignment::Assignment::B => Self::B,
            oracle_boosting_hex_assignment::Assignment::C => Self::C,
        }
    }
}

```
Then we can use the built-in functions to the proto enum to get the enum value ex. `hex.urbanized()` rather than `hex.urbanized`

mobile_verifier/src/boosting_oracles/assignment.rs
```diff
@@ -40,59 +60,53 @@ impl fmt::Display for Assignment {
...
-        (C, A, A) => dec!(0.40),
-        (C, B, A) => dec!(0.30),
-        (C, C, A) => dec!(0.05),
-        // No POI Not Urbanized
-        (C, A, B) => dec!(0.20),
-        (C, B, B) => dec!(0.15),
-        (C, C, B) => dec!(0.03),
-        // Outside of USA
-        (_, _, C) => dec!(0.00),
+#[cfg(test)]
```
andymck, line 103: would somewhere in the tests folder not be a better home for this impl?

mobile_verifier/src/boosting_oracles/mod.rs
```diff
@@ -1,136 +1,418 @@
...
+            Landtype::Mangrove => Assignment::C,
+            Landtype::Moss => Assignment::C,
+        }
     }
 }
 
-trait DiskTreeLike: Send + Sync {
-    fn get(&self, cell: hextree::Cell) -> hextree::Result<Option<(hextree::Cell, &[u8])>>;
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub enum Landtype {
```
andymck, line 112: Nit: LandType rather than Landtype

mobile_verifier/src/boosting_oracles/mod.rs
```diff
@@ -1,136 +1,418 @@
...
+            poi_water_outside_us,
+            poi_no_data_built_outside_us,
+            poi_no_data_grass_outside_us,
+            poi_no_data_water_outside_us,
+            no_poi_built_outside_us,
+            no_poi_grass_outside_us,
+            no_poi_water_outside_us,
+        ];
+        for cell in all_cells {
+            println!("Cell: {cell:?}");
```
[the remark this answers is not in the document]
michaeldjeffrey: It was not. Removed!

mobile_verifier/src/boosting_oracles/assignment.rs
michaeldjeffrey: Yes! That's much nicer, thank you.

mobile_verifier/src/boosting_oracles/assignment.rs
michaeldjeffrey: This method is currently only used in unit tests.
We wouldn't be able to add an implementation to `HexAssignments` without a trait, and right now I can't think of a way to justify it all.

mobile_verifier/src/boosting_oracles/mod.rs
michaeldjeffrey: I ended up with `Landtype` for a few reasons. 

I my opinion, as a member, `landtype` reads nicer to me than `land_type`. 
Especially in places where you're qualifying the data
```rs
let landtype_disktree = DiskTreeMap::open(&settings.landtype_data_set)?;
```
And it felt like it fit in more with `footfall` and `urbanized`. 

When I say them out loud with the inflection of capitalization, both `Landtype` and `LandType` feel more or less the same. So I went with what I thought led to more aesthetically pleasing code.

bbalser (approved):

andymck (approved):

maplant (approved):

mobile_verifier/src/boosting_oracles/assignment.rs
```diff
@@ -40,59 +57,53 @@ impl fmt::Display for Assignment {
...
+            // light blue - No POI Urbanized
+            (C, A, A) => dec!(0.40),
+            (C, B, A) => dec!(0.30),
+            (C, C, A) => dec!(0.05),
+            // dark blue - No POI Not Urbanized
+            (C, A, B) => dec!(0.20),
+            (C, B, B) => dec!(0.15),
+            (C, C, B) => dec!(0.03),
+            // gray - Outside of USA
+            (_, _, C) => dec!(0.00),
```
kurotych, line 95: Does it mean that all radios outside of the USA won't receive coverage points?
What about Mexico?

PR #791, opened 2024-04-19
Macpie/custom tracing

macpie:
- Add a `tracing_subscriber` that is re-loadable via a file (takes the content of the file and turn it into an [EnvFilter](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/filter/struct.EnvFilter.html))
    - File can be created
    - File delete will go back to original filter from settings (same for empty file)
- Gives negative feedback when file content cannot be parsed as an `EnvFilter`
- Functions to create tracing spans around GRPC (tonic) or HTTP (axum) including latency
- GRPC or HTTP are lib features 
- Include one example of use in `ingest/server_mobile`
- Filter ex `custom_tracing=info,mapper_ingestor[tracing{uri=/session_init}]`


custom_tracing/src/lib.rs
```diff
@@ -0,0 +1,131 @@
...
+
+    Ok(())
+}
+
+pub fn record<T>(field: &str, value: T)
+where
+    T: ToString,
+{
+    Span::current().record(field, &tracing::field::display(value.to_string()));
+    ()
```
maplant:
```suggestion
```

ingest/Cargo.toml
```diff
@@ -4,24 +4,24 @@ version = "0.1.0"
 description = "PoC Ingest Server for the Helium Network"
 authors.workspace = true
 edition.workspace = true
-license.workspace  = true
+license.workspace = true
 
 [dependencies]
-anyhow = {workspace = true}
```
maplant, line 10: Formatting these cargo files should probably be done in a separate PR

custom_tracing/src/grpc_layer.rs
```diff
@@ -0,0 +1,26 @@
+use helium_proto::services::Body;
+use http::request::Request;
+use tower_http::{
+    classify::{GrpcErrorsAsFailures, SharedClassifier},
+    trace::{DefaultOnFailure, DefaultOnResponse, TraceLayer},
+    LatencyUnit,
+};
+use tracing::{Level, Span};
+
+type GrpcLayer =
```
maplant, line 10: I think the function name should probably be a little bit more verbose than new

custom_tracing/src/lib.rs
```diff
@@ -0,0 +1,130 @@
...
+        self.reload_handle
+            .modify(|filter| *filter = tracing_subscriber::EnvFilter::new(new_filter.clone()))?;
+
+        tracing::info!(filter = new_filter, "deleted tracing file, updated");
+        Ok(())
+    }
+}
+
+async fn watch_file(state: State, file: String) -> Result<()> {
+    let (mut tx, mut rx) = channel(1);
```
maplant, line 82: It doesn't really look like this function needs to be async? You can just move the match statement into the watcher closure, then call this function in a spawn blocking

macpie: done

ingest/src/settings.rs
```diff
@@ -13,6 +13,9 @@ pub struct Settings {
     /// "ingest=debug,poc_store=info"
     #[serde(default = "default_log")]
     pub log: String,
+    /// "tracing"
```
michaeldjeffrey, line 16: Can this be bit more descriptive?

ingest/src/server_mobile.rs
```diff
@@ -194,6 +211,16 @@ impl poc_mobile::PocMobile for GrpcServer {
         let subscriber_id = event.subscriber_id.clone();
         let timestamp_millis = event.timestamp;
 
+        custom_tracing::record(
+            "pub_key",
+            bs58::encode(event.carrier_pub_key.clone()).into_string(),
+        );
```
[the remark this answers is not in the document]
macpie: good question I  ll remove from debug line

macpie:
> Since this custom tracing is spawning a process off is tokio respecting the shutdown signals from the thing that instantiates these being shutdown or do these need to be run underneath the task manager in order to shut down as expected with a ctrl-c?

Yes I made sure it does

ingest/src/server_mobile.rs
```diff
@@ -194,6 +211,16 @@ impl poc_mobile::PocMobile for GrpcServer {
...
+
+        custom_tracing::record(
+            "subscriber_id",
+            bs58::encode(subscriber_id.clone()).into_string(),
```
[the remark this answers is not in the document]
macpie: yeah just realized that in another PR

custom_tracing/src/lib.rs
```diff
@@ -0,0 +1,132 @@
...
+    });
+
+    tracing::info!("custom tracing installed");
+
+    Ok(())
+}
+
+pub fn record<T>(field: &str, value: T)
+where
+    T: ToString,
```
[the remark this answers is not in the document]
macpie: - I'll fix to something that implement `fmt::Display`
- We can only record on a span so we need the call to get the current one.

ingest/src/settings.rs
```diff
@@ -13,6 +13,9 @@ pub struct Settings {
     /// "ingest=debug,poc_store=info"
     #[serde(default = "default_log")]
     pub log: String,
+    /// File name to be watched by custom tracing
+    #[serde(default = "default_tracing")]
```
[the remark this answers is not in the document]
macpie: done

michaeldjeffrey (approved):

PR #792, opened 2024-04-19
Gateway service info perms

michaeldjeffrey:
Update the Mobile Config Gateway Service to allow gateways to request information about themselves.

ORA-350

### DIFF
--- a/mobile_config/src/gateway_service.rs
+++ b/mobile_config/src/gateway_service.rs
@@ -47,21 +47,16 @@ impl GatewayService {
         Err(Status::permission_denied("unauthorized request signature"))
     }
 
-    fn verify_request_signature_for_info<R>(
-        &self,
-        address: &PublicKey,
-        signer: &PublicKey,
-        request: &R,
-    ) -> Result<(), Status>
-    where
-        R: MsgVerify,
-    {
-        if address == signer && request.verify(signer).is_ok() {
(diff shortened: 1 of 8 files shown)


mobile_config/src/gateway_service.rs
```diff
@@ -47,6 +47,23 @@ impl GatewayService {
...
+        address: &PublicKey,
+        signer: &PublicKey,
+        request: &R,
+    ) -> Result<(), Status>
+    where
+        R: MsgVerify,
+    {
+        if address == signer && request.verify(signer).is_ok() {
+            tracing::debug!(%signer, "self authorized");
+            return Ok(());
```
madninja, line 61: Don’t use debug print for signer?

mobile_config/tests/gateway_service.rs
```diff
@@ -0,0 +1,99 @@
...
+        Code::PermissionDenied,
+        "signature must match signer"
+    );
+
+    Ok(())
+}
+
+fn make_keypair() -> Keypair {
+    Keypair::generate(
+        KeyTag {
```
madninja, line 83: You can us KeyTag::default for the default instance here

bbalser (approved):

mobile_config/src/gateway_service.rs
madninja: Oh I was confused between `%` and `?`.. my bad

PR #797, opened 2024-04-23
Refactor: Hex location use `hextree::Cell` type

michaeldjeffrey:
This is something that I was planning on doing as part of HIP-109 changes I was going to be making. 
So I thought I'd push it early to make it easier to review.

---
Got some advice from Jay to only use a `h3o::CellIndex` when you need to
convert to/from a LatLng. So we went with `hextree::Cell` for the type.

`hextree::Cell` is a loose wrapper around a u64. Parsing the value as
early as possible tightens the gaurantee that we can make valid types
that need use `Cell`s.

It is also an attempt to remove discrepancies where some types carry a
`i64` hex straight from a database, and other types carry a `u64` that
were converted out of a database, even though they're goal is the same.

(see previous commit where `#[sqlx(try_from = "i64")]` was enabled by
the update to hextree).

### DIFF
--- a/mobile_config/src/boosted_hex_info.rs
+++ b/mobile_config/src/boosted_hex_info.rs
@@ -2,7 +2,9 @@ use crate::client::{hex_boosting_client::HexBoostingInfoResolver, ClientError};
 use chrono::{DateTime, Duration, Utc};
 use file_store::traits::TimestampDecode;
 use futures::stream::{BoxStream, StreamExt};
+use helium_proto::services::poc_mobile::BoostedHex as BoostedHexProto;
 use helium_proto::BoostedHexInfoV1 as BoostedHexInfoProto;
+use hextree::Cell;
 use solana_sdk::pubkey::Pubkey;
 use std::{collections::HashMap, convert::TryFrom, num::NonZeroU32};
 
@@ -14,7 +16,7 @@ lazy_static::lazy_static! {
 
 #[derive(Clone, Debug)]
 pub struct BoostedHexInfo {
-    pub location: u64,
+    pub location: Cell,
     pub start_ts: Option<DateTime<Utc>>,
     pub end_ts: Option<DateTime<Utc>>,
     pub period_length: Duration,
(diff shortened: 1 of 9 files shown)


mobile_config/src/boosted_hex_info.rs
```diff
@@ -295,7 +316,7 @@ mod tests {
         };
 
         let msg = BoostedHexInfo::try_from(proto)?;
-        assert_eq!(631252734740306943, msg.location);
+        assert_eq!(hextree::Cell::from_raw(631252734740306943)?, msg.location);
```
andymck, line 319: why not import hextree:Cell in the above crate and remove the need to fully qualify each usage of Cell

andymck (commented):
I think this is fine in itself and it is better to have a consistent representation of a location.  The only point I would make is that maybe it doesnt go wide enough.  We now have a location for the purposes of boosted hexes defined as a Cell but in all other uses of location we still have it as a u64.  This inconsistency niggles me.  Most non boosted hex sources of location data originates from mobile config's gateway info, so perhaps that should be changed to a Cell and span out from there.

mobile_config/src/boosted_hex_info.rs
michaeldjeffrey: I think this file was one of the early ones in the change, before I felt like `Cell` was used enough to live on it's own. I've updated the occurrences in this file. :+1:

michaeldjeffrey:
> The only point I would make is that maybe it doesnt go wide enough. We now have a location for the purposes of boosted hexes defined as a Cell but in all other uses of location we still have it as a u64. This inconsistency niggles me. 

That's where I'd like to be. Considering all this change fanned out from the `BoostedHex::location` field, it seemed more responsible to switch over only the things related to what I plan on touching and get that change in. Rather than starting from the source and me having to make decisions for parts of the code I haven't interacted with yet, and landing a behemoth of a refactor PR.

andymck (approved):

bbalser (approved):

PR #798, opened 2024-04-26
HIP 109 boost by device type

michaeldjeffrey:
https://github.com/helium/proto/pull/401

---

### `BoostedHexInfo.device_type` added

Because legacy boosts exists (`NULL` in the database mapping to `BoostedHexDeviceType::All`), a single hex can now be considered boosted more than a single time. `BoostedHexes` carries `n` number of boosts per cell. 

When asking for the current multiplier of a hex, the device type in question must be provided. All boosts that apply are combined.

### `parse_boosted_hex_info_from_database` test

Because this project does not control the metadata db, I wanted to make sure `BoostedHexDeviceType` would correctly serialize as a part of `BoostedHexInfo`.

Looking at the [helium-program-library implementation](https://github.com/helium/helium-program-library/pull/634/files) it wasn't clear how `DeviceTypeV0` would make it's way into the database. There is a similar field `DeviceType` in `mobile_hotspot_infos`  that has also [uses a rust enum](https://github.com/helium/helium-program-library/blob/master/programs/helium-entity-manager/src/state.rs#L234) that is present as a `jsonb` column.

If that assumption turns out to be false, updating the test should lead us to the glory of correctly serializing `device_type`.

### `BoostedHexes` and "expired boosts"

@bbalser and I talked about `BoostedHexes` only ever containing non-expired boosts. 

Ideally, that would occur at the level of mobile-config when boosted hexes are being streamed out of the db. However, boosted hexes do not have a `end_ts` field that is easily queryable, and I'm not comfortable at this point in time putting that much math into a sql query.

I tried at each next level. In the end, I decided to give up on that goal for this PR. Tests using `BoostedHexes` directly would not have a way to enforce the contract of only holding actively boosted hexes, and now didn't feel like the correct time to dive into those tests to find a way they could.


michaeldjeffrey:
```rs
let (_, rewards) = tokio::join!(
    // run rewards for poc and dc
    rewarder::reward_poc_and_dc(
        &pool,
        &hex_boosting_client,
        &mobile_rewards_client,
        &speedtest_avg_client,
        &epoch,
        dec!(0.0001)
    ),
    receive_expected_rewards(&mut mobile_rewards)
);
```

When I tried to run `rewarder::reward_poc_and_dc` by itself, _then_ receive the rewards, the test would never finish. I do not understand why.

mobile_verifier/src/coverage.rs
```diff
@@ -549,6 +549,13 @@ impl IndoorCoverageLevel {
             _ => dec!(0),
         }
     }
+
+    fn device_type(&self) -> BoostedHexDeviceType {
+        match self.radio_key {
+            OwnedKeyType::Cbrs(_) => BoostedHexDeviceType::CbrsIndoor,
```
macpie, line 555: I am guessing the fact that we default to indoor for both is from the HIP?

andymck:
> ```rust
> let (_, rewards) = tokio::join!(
>     // run rewards for poc and dc
>     rewarder::reward_poc_and_dc(
>         &pool,
>         &hex_boosting_client,
>         &mobile_rewards_client,
>         &speedtest_avg_client,
>         &epoch,
>         dec!(0.0001)
>     ),
>     receive_expected_rewards(&mut mobile_rewards)
> );
> ```
> 
> When I tried to run `rewarder::reward_poc_and_dc` by itself, _then_ receive the rewards, the test would never finish. I do not understand why.

This will be because the filesink writer awaits a oneshot response to indicate the file/s have been written to disk.  If you do not call `receive_expected_rewards` then this oneshot never gets returned and thus the rewarder will wait indefinitely.  See link below for an example of the oneshot response being sent from the tests receive handler: https://github.com/helium/oracles/blob/main/mobile_verifier/tests/common/mod.rs#L35

andymck:
Other than clippy not being happy, this all looks good to me.

mobile_verifier/tests/integrations/hex_boosting.rs
```diff
@@ -990,6 +1007,163 @@ async fn test_poc_with_cbrs_and_multi_coverage_boosted_hexes(pool: PgPool) -> an
...
+#[sqlx::test]
+async fn test_poc_boosted_hex_stack_multiplier(pool: PgPool) -> anyhow::Result<()> {
+    // This test seeds the database with 2 identically performing radios on
+    // neighboring hexes. Both are eligible for boosted rewards.
+    //
+    // In the first pass, no hexes are boosted, to ensure the radios are equal.
+    // In the second pass, 1 radio is boosted with 2 boosts.
+    //  - legacy boost,   BoostType::All
+    //  - specific boost, BoostType::CbrsIndoor
+    // Rewards are checked to ensure the 2 boosts accumulate correctly.
```
andymck, line 1019: can we get a test  to include boosted hex entries that cover each of the device types

michaeldjeffrey: What's the goal?

We can add a boosted hex of each type, to make sure only `::All` boosted hexes are added to individual radios types.

If you're looking for every radio type getting rewarded with different boosts applied... The only way I can think to make any meaningful assertions against those rewards would be to:
1. hardcode the reward values, which we're trying to stay/get away from
2. do the reward math in the test, which would need to change as often as the actual reward calculations.

There's also an option of doing a second pass for each type of radio. But if all radio types are involved in each pass I think that would put us back in the situation of, the only meaningful assertions we could make are against concrete numbers, or doing complex math in the test.

We could also flesh out the unit test `boosted_hex_info::tests::boosted_hexes_accumulate_multipliers` to go through all device types.

Thoughts?

andymck: The goal would be to ensure that when a hex is boosted for each specific device type only that type gets boosted rewards.  If you want to avoid hardcoding reward amounts you can assert on relative values, you could setup two hexes each with a cbrs and a wifi radio.  Hex 1 is boosted for CBRS with a multiplier of 5x, hex 2 has no boosts.  The cbrs radio in hex1 should have 5x the reward amount of the cbrs radio in hex2.  The wifi radios in both hexes should have the same reward

michaeldjeffrey: Gotcha, that sounds like a good test. 
Added here https://github.com/helium/oracles/pull/798/commits/dc181f87ff99a5bd6c88699690864a08b5fdaa9e

boost_manager/src/activator.rs
```diff
@@ -96,7 +96,7 @@ where
         manifest: RewardManifest,
     ) -> Result<()> {
         // get latest boosted hexes info from mobile config
```
macpie, line 98:
```suggestion
        // get active boosted hexes info from mobile config
```

coverage_map/src/indoor.rs
```diff
@@ -172,10 +179,14 @@ mod test {
...
             insert_indoor_coverage_object(&mut indoor_coverage, cov_obj);
         }
-        let ranked: HashMap<_, _> =
-            into_indoor_coverage_map(indoor_coverage, &NoBoostedHexes, Utc::now())
-                .map(|x| (x.cbsd_id.clone().unwrap(), x))
-                .collect();
+        let ranked: HashMap<_, _> = into_indoor_coverage_map(
+            indoor_coverage,
+            &NoBoostedHexes,
+            DeviceType::CbrsIndoor,
```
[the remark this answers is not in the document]
macpie: This is a test so I am guessing that it defaults to cbrs indoor

boost_manager/src/activator.rs
michaeldjeffrey: I think the update to to the function name renders the comment unnecessary. 
Good looking out.

mobile_config/src/boosted_hex_info.rs
```diff
@@ -128,27 +143,39 @@ impl TryFrom<BoostedHexProto> for BoostedHex {
...
+}
+
 impl BoostedHexes {
-    pub fn new(hexes: Vec<BoostedHexInfo>) -> Self {
-        let hexes = hexes
-            .into_iter()
-            .map(|info| (info.location, info))
-            .collect();
-        Self { hexes }
+    pub fn test_new_active(hexes: Vec<BoostedHexInfo>) -> anyhow::Result<Self> {
```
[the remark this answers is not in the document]
michaeldjeffrey: Unfortunately, it does not appear that it can be. 
Being defined in `mobile_config`, and depended on in `boost_manager` and `mobile_verifier`. 
`boost_manager` cannot use `#[cfg(test)]` code from another crate.

So I opted for an explicit `test_` prefix.

PR #803, opened 2024-05-03
Fix clippy

macpie:
It seems like this version of Rust does not like to have `struct` and `impl` in different files.


macpie:
I have tried and clippy is not happy:
```
error: associated function `new` is never used
  --> mobile_verifier/tests/common/mod.rs:26:12
   |
25 | impl MockHexBoostingClient {
   | -------------------------- associated function in this implementation
26 |     pub fn new(boosted_hexes: Vec<BoostedHexInfo>) -> Self {
   |            ^^^
   |
   = note: `-D dead-code` implied by `-D warnings`
   = help: to override `-D warnings` add `#[allow(dead_code)]`
```
But `new` is used...

michaeldjeffrey (approved):

PR #805, opened 2024-05-03
Update metrics

michaeldjeffrey:
This is pointing into `mj/telemetry-histograms` branch to ease reviewing. This update changes a lot of files.

Also bumps metrics-exporter-prometheus.

The biggest change from 0.21 -> 0.22 is in this PR https://github.com/metrics-rs/metrics/pull/394

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -71,7 +71,7 @@ spl-token = "3.5.0"
 reqwest = {version = "0", default-features=false, features = ["gzip", "json", "rustls-tls"]}
 beacon = { git = "https://github.com/helium/proto", branch = "master" }
 humantime = "2"
-metrics = "0.22"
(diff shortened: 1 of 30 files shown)


Cargo.toml
```diff
@@ -71,7 +71,7 @@ spl-token = "3.5.0"
 reqwest = {version = "0", default-features=false, features = ["gzip", "json", "rustls-tls"]}
 beacon = { git = "https://github.com/helium/proto", branch = "master" }
 humantime = "2"
-metrics = "0.21"
+metrics = "0.22"
```
bbalser, line 74: should this be `>=0.22` ?

michaeldjeffrey: Indeed it does, I forgot cargo treats pre `1.0` versions differently.

PR #806, opened 2024-05-06
Attempt to make CI faster

macpie:
The goals here are:
- Make test faster by running most items in parallel
- Make it more readable by having each test being a single step so we can get faster feedback on what fails (fmt might fail but would be nice to still know if the actual tests failed or not)

@maplant  confirmed that running test in release mode or not makes no difference.


maplant (approved):

.github/workflows/Tests.yml
```diff
@@ -0,0 +1,157 @@
...
+      - name: Clippy
+        run: cargo clippy --all-targets -- -Dclippy::all -D warnings
+
+  tests:
+    needs: build
+    runs-on: oracles-20.04
+    strategy:
+      fail-fast: false
+      matrix:
+        package: [boost-manager,file-store,ingest,iot-config,iot-packet-verifier,iot-verifier,mobile-config,mobile-packet-verifier,mobile-verifier,reward-scheduler,task-manager]
```
michaeldjeffrey, line 117: We'll have to remember to call out if a PR adds a test to a package not in this list.

michaeldjeffrey (approved):

.github/workflows/Tests.yml
macpie: yes indeed

PR #807, opened 2024-05-08
Make hex_assignments (boosting_oracles) as separate create

kurotych:
It makes it possible to reuse `hex_assignments` code without bloated `mobile-verifier` dependencies (which causes dependency conflict issues)

### DIFF
--- a/iot_verifier/src/runner.rs
+++ b/iot_verifier/src/runner.rs
@@ -10,7 +10,7 @@ use crate::{
     witness_updater::WitnessUpdater,
     Settings,
 };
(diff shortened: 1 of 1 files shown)


bbalser:
I think we should rename the crate `hex_assignments` rather than `boosting_oracles`.  I didn't like that name before, but I really don't like it as a crate.

maplant:
@kurotych do you want to use the `DataSetDownloader`, or would you rather we keep that in `mobile_verifier`? If you plan on using it for updating the data sets we'd be happy to include it, but it would bring in a rather large set of dependencies with it

PR #810, opened 2024-05-10
Macpie/custom tracing everywhere

macpie:
Adding custom tracing library in all the packages

### DIFF
--- a/boost_manager/src/settings.rs
+++ b/boost_manager/src/settings.rs
@@ -10,7 +10,7 @@ pub struct Settings {
     /// "poc_entropy=debug,poc_store=info"
     #[serde(default = "default_log")]
     pub log: String,
-    #[serde(default = "custom_tracing::Settings::serde_default")]
+    #[serde(default)]
     pub custom_tracing: custom_tracing::Settings,
     /// Cache location for generated verified reports
     pub cache: String,
--- a/boost_manager/src/settings.rs
+++ b/boost_manager/src/settings.rs
@@ -10,6 +10,7 @@ pub struct Settings {
     /// "poc_entropy=debug,poc_store=info"
     #[serde(default = "default_log")]
     pub log: String,
+    #[serde(default = "custom_tracing::Settings::serde_default")]
     pub custom_tracing: custom_tracing::Settings,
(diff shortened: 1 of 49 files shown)


boost_manager/src/settings.rs
```diff
@@ -9,6 +9,9 @@ pub struct Settings {
     /// "poc_entropy=debug,poc_store=info"
     #[serde(default = "default_log")]
     pub log: String,
+    /// File name to be watched by custom tracing
+    #[serde(default = "default_tracing_cfg_file")]
+    pub tracing_cfg_file: String,
```
michaeldjeffrey, line 14: Can we follow the poc-metrics and file-store pattern of using a libs settings struct.
```suggestion
    pub custom_tracing: custom_tracing::Settings,
```

macpie: Yeah I was thinking of doing that. 👍🏻

custom_tracing/src/settings.rs
```diff
@@ -0,0 +1,20 @@
...
+    #[serde(default = "default_tracing_cfg_file")]
+    pub tracing_cfg_file: String,
+}
+
+pub fn default_tracing_cfg_file() -> String {
+    "tracing.cfg".to_string()
+}
+
+impl Settings {
+    pub fn serde_default() -> Self {
```
michaeldjeffrey, line 15: Because you have a default provider for `tracing_cfg_file`, `Settings` can derive `Default`.

Then where it's used can annotate the field with `#[serde(default)]` and don't need to point to an explicit default function.

ingest/src/server_iot.rs
```diff
@@ -397,3 +409,7 @@ pub async fn grpc_server(settings: &Settings) -> Result<()> {
         .start()
         .await
 }
+
+fn pub_key_to_b58(pub_key: &[u8]) -> String {
```
michaeldjeffrey, line 413: It looks like it would be useful to provide a `custom_tracing::record_b58/2`, or export `pub_key_to_b58` from custom_tracing.

michaeldjeffrey (approved):

andymck (approved):

PR #811, opened 2024-05-10
humantime serde

michaeldjeffrey:
Bring Durations and DateTimes into alignment across settings.
Companion Ops PR: https://github.com/novalabsxyz/ops/pull/1348

## A Bit about `Duration`

When possible `std::time::Duration` was preferred in place of `chrono::Duration`. 

If there was a duration or timestamp comparison that looked like negative Durations could be in play `chrono::Duration` was left. (In future chrono versions they've renamed `chrono::Duration` to `chrono::TimeDelta` with the thinking that `Duration` implies only positive time.)

There are also some tests with `chrono::Duration` is left in because the constructors added readability.
There's a current nightly only feature that will add some of these constructors to `std::time::Duration` https://github.com/rust-lang/rust/issues/120301

## `.u(hhhh)nwrap()`???

In an attempt to make default duration settings easier to understand, I used `humantime::format_duration` which returns a result. These are only used in serde default functions.


boost_manager/src/main.rs
```diff
@@ -85,16 +85,18 @@ impl Server {
...
         let (manifest_receiver, manifest_server) =
             file_source::continuous_source::<RewardManifest, _>()
                 .state(pool.clone())
                 .store(file_store)
                 .prefix(FileType::RewardManifest.to_string())
-                .lookback(LookbackBehavior::StartAfter(settings.start_after()))
-                .poll_duration(settings.reward_check_interval())
-                .offset(settings.reward_check_interval() * 2)
+                .lookback(LookbackBehavior::StartAfter(settings.start_after))
+                .poll_duration(reward_check_interval)
```
bbalser, line 98: The file_info_poller takes a chrono Duration, just to turn it into a std Duration, maybe we should just change it to accept a std Duration?

bbalser (approved):

boost_manager/src/main.rs
michaeldjeffrey: That makes good sense to me. Also because it doesn't make sense for file info poller to need to support negative time deltas.

maplant (approved):

PR #817, opened 2024-05-24
Create hex_assignments crate 

kurotych:
## Motivation
Make it possible to reuse some parts of boosting Oracle code without bloated dependencies from `mobile_verifier` and the tided sqlx version.

## Changes
- Create hex_assignments crate
- [Add HexBoostDataAssignments trait](https://github.com/helium/oracles/pull/817/commits/3054ebdeb2fc98d8dc00e86b1a1476341a1cd329) to comfortable pass `HexBoostData` as argument without repeatable specifying template arguments
- Remove the `new_mock` function instead use the optional parameter in `new`, since I wanna use it without "data set downloader" but the "_mock" postfix signaling that it is functional for "testing only" which is not.


michaeldjeffrey (approved):
Can we get the `Cargo.toml` files auto-formatted to match everything?

maplant (approved):

PR #823, opened 2024-06-04
Fix custom tracing b58 printing

macpie:


custom_tracing/Cargo.toml
```diff
@@ -6,22 +6,23 @@ license.workspace = true
 edition.workspace = true
 
 [dependencies]
-serde = { version = "1", features = ["derive"] }
-notify = { version = "6", default-features = false }
 anyhow = "1"
+axum = { version = "0.7", features = ["tracing"], optional = true }
+bs58 = { version = "0.4", features = ["check"] }
```
madninja, line 11: don'tt be this precise in a lib this far upstream? Make version = "0" or version = ">=..." if it needs a minimum version

custom_tracing/src/lib.rs
```diff
@@ -54,7 +55,7 @@ where
 }
 
 pub fn record_b58(key: &str, pub_key: &[u8]) {
-    let b58 = bs58::encode(pub_key).into_string();
+    let b58 = PublicKeyBinary::from(pub_key).to_string();
```
madninja, line 58: huh? what format was that key string thing there?

macpie: Just wanted to print b58 my old stuff was bad did not realized that `to_string` on a `PublicKeyBinary` or `PublicKey` gives you the b58 string version

bbalser (approved):

michaeldjeffrey (approved):

michaeldjeffrey:
Can we get a description of why this was the wrong way to print a b58?

madninja:
> Can we get a description of why this was the wrong way to print a b58?

in helium crypto we use b58 *check encoding* (with a version 0 prefix) not a plain b58 encoding

PR #825, opened 2024-06-06
Change coverage map to treat hotspot_key as bytes rather than a PublicKeyBinary

bbalser:
In an effort to keep this crate simple and the deps small, refactored to treat the hotspot_key as just a bag of bytes rather than a PublicKeyBinary.

### DIFF
--- a/coverage_map/Cargo.toml
+++ b/coverage_map/Cargo.toml
@@ -10,7 +10,6 @@ edition.workspace = true
 [dependencies]
 chrono = { workspace = true }
 h3o = { workspace = true }
-helium-crypto = { workspace = true }
 helium-proto = { workspace = true }
 hex-assignments = { path = "../hex_assignments" }
 hextree = { workspace = true }
--- a/coverage_map/Cargo.toml
+++ b/coverage_map/Cargo.toml
@@ -9,8 +9,5 @@ edition.workspace = true
 
 [dependencies]
 chrono = { workspace = true }
-h3o = { workspace = true }
-helium-proto = { workspace = true }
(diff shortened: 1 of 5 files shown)


coverage_map/src/lib.rs
```diff
@@ -382,12 +358,9 @@ mod test {
     }
 
     fn outdoor_cbrs_coverage(cbsd_id: &str, hex: u64, signal_power: i32) -> CoverageObject {
-        let owner: PublicKeyBinary = "112NqN2WWMwtK29PMzRby62fDydBJfsCLkCAf392stdok48ovNT6"
-            .parse()
-            .expect("failed owner parse");
         CoverageObject {
             indoor: false,
-            hotspot_key: owner,
+            hotspot_key: vec![0, 0],
```
maplant, line 363: it's not very important, but these don't even need to be length 2. They could just be `vec![0]`, `vec![1]`, etc.

maplant (approved):

coverage_map/Cargo.toml
```diff
@@ -10,7 +10,6 @@ edition.workspace = true
 [dependencies]
 chrono = { workspace = true }
 h3o = { workspace = true }
```
michaeldjeffrey, line 12: `h30`, `helium-proto`, and `uuid` are also candidates for removal.

michaeldjeffrey (approved):

coverage_map/Cargo.toml
bbalser: nice catch!, removed 530cb59

PR #826, opened 2024-06-06
Abstract data set processing to a trait

maplant:
This PR abstracts out the data set processing part of the `DataSetDownloader` into a separate trait. 
There are two implementers, `FileSinkClient` and `NopDataSetProcessor`. 
This is so that the `DataSetDownloader` can be used without processing the data. 
The implementer of this trait being `FileSinkClient` is because you would only want to supply a file sink if you are processing the data.

### DIFF
--- a/mobile_verifier/src/boosting_oracles/data_sets.rs
+++ b/mobile_verifier/src/boosting_oracles/data_sets.rs
@@ -173,11 +173,11 @@ where
     h.urbanization.is_ready() && h.footfall.is_ready() && h.landtype.is_ready()
 }
 
-pub struct DataSetDownloaderDaemon<A, B, C> {
+pub struct DataSetDownloaderDaemon<A, B, C, T> {
     pool: PgPool,
     data_sets: HexBoostData<A, B, C>,
     store: FileStore,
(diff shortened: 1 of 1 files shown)


mobile_verifier/src/boosting_oracles/data_sets.rs
```diff
@@ -507,6 +502,73 @@ impl DataSetType {
...
+    async fn set_unassigned_oracle_boosting_assignments(
+        &self,
+        pool: &PgPool,
+        data_sets: &HexBoostData<impl HexAssignment, impl HexAssignment, impl HexAssignment>,
+    ) -> anyhow::Result<()>;
+}
+
+#[async_trait::async_trait]
+impl DataSetProcessor for FileSinkClient {
+    async fn set_all_oracle_boosting_assignments(
```
michaeldjeffrey, line 522: The functions this trait is replacing are unused, can they be removed?

maplant: oop yup, my bad

michaeldjeffrey (approved):

bbalser (approved):

PR #827, opened 2024-06-07
Include share values in rewards manifest

andymck:
This PR adds the `bones_per_coverage_point` among other information to the rewards manifest.


maplant:
I think that this should be bones per coverage point, not shares.

andymck (commented):
lgtm, but as a committer i cant approve

bbalser (approved):

file_store/src/reward_manifest.rs
```diff
@@ -1,12 +1,27 @@
...
     pub start_timestamp: DateTime<Utc>,
     pub end_timestamp: DateTime<Utc>,
+    pub reward_data: Option<RewardData>,
+}
+
+#[derive(Clone, Debug)]
+pub enum RewardData {
+    MobileRewardData {
+        poc_bones_per_coverage_point: Decimal,
+        boosted_poc_bones_per_coverage_point: Decimal,
```
michaeldjeffrey, line 18:
```suggestion
        poc_bones_per_reward_share: Decimal,
        boosted_poc_bones_per_reward_share: Decimal,
```
These should be representing portions of a reward share like the IoT reward data.

"Coverage Points" has covered many meaning in the past. Ideally it refers to points rewarded _only_ for covering hexes. When combined with Multipliers they become a radios "Reward Shares".

This change is also taking place over in [radio_reward_v2 proto](https://github.com/helium/proto/pull/407/files#diff-41a424c2a5f8e8fc2a265baae2b8a8f277ea1d7d21a1a4bb73a2ab391c971e6eR490-R495).

mobile_verifier/src/rewarder.rs
```diff
@@ -301,12 +305,21 @@ where
         transaction.commit().await?;
 
         // now that the db has been purged, safe to write out the manifest
+        let reward_data = ManifestMobileRewardData {
+            poc_bones_per_coverage_point: Some(helium_proto::Decimal {
+                value: poc_dc_shares.normal.to_string(),
```
michaeldjeffrey, line 310: Do we want to support a maximum amount precision?

The [rust_decimal docs](https://docs.rs/rust_decimal/latest/rust_decimal/index.html) says trailing zeros may be exposed when in string form, do we want to remove those?

maplant: I don't think that's an issue personally

maplant: To more precisely address this, here is what the google decimal specification says (which we are adhering to):
```
 Additionally, services **may** preserve trailing zeroes in the fraction to indicate increased precision, but are not required to do so.
 ```
so either option is fine

michaeldjeffrey (approved):

PR #828, opened 2024-06-13
Add mock speedtest constructor

maplant:
Add a mock speedtest constructor. This is so that the mobile rewards estimator can easily construct a collection of speedtests that are guaranteed to be assigned `SpeedtestTier::Good` when averaged. 

I'm open to better ways of doing this, this just seemed like the easiest way


michaeldjeffrey (approved):
My alternative to creating speedtests was this.
https://github.com/helium/oracles/blob/aecb7bcd1051e2060794145e6dd54ec03793da51/coverage_point_calculator/tests/coverage_point_calculator.rs#L406-L431

PR #831, opened 2024-06-18
Update to use pyth sponsored feeds

bbalser:
Current Pyth feeds are being deprecated and we have to move to using new sponsored feeds.

Sponsored feeds are not backwards compatible.

### DIFF
--- a/price/src/price_generator.rs
+++ b/price/src/price_generator.rs
@@ -3,7 +3,9 @@ use anyhow::{anyhow, bail, Error, Result};
 use chrono::{DateTime, TimeZone, Utc};
 use file_store::file_sink;
 use futures::{future::LocalBoxFuture, TryFutureExt};
+use helium_anchor_gen::anchor_lang::AccountDeserialize;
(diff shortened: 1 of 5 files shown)


price/src/price_generator.rs
```diff
@@ -233,36 +235,45 @@ impl PriceGenerator {
...
+        let PriceUpdateV2 {
+            price_message:
+                PriceFeedMessage {
+                    ema_price: price,
+                    ema_conf: confidence,
+                    exponent,
+                    publish_time,
+                    ..
+                },
+            ..
```
maplant, line 248: it's probably fewer lines to destructure this one level deep and then dot reference the fields: 
```suggestion
        let PriceUpdateV2 {
            price_message,
            ..
```

maplant (approved):

michaeldjeffrey (approved):

abhay (approved):

ChewingGlass (approved):

PR #832, opened 2024-06-22
Update price service to use helium-lib

bbalser:
price now uses helium-lib to get sponsored pyth price off chain

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -103,7 +103,6 @@ triggered = "0"
 futures = "*"
 futures-util = "*"
 prost = "*"
-pyth-solana-receiver-sdk = "0"
 once_cell = "1"
 lazy_static = "1"
 config = { version = "0", default-features = false, features = ["toml"] }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,7 +35,7 @@ license = "Apache-2.0"
 edition = "2021"
 
 [workspace.dependencies]
-anchor-client = "0.29.0"
+anchor-client = "0.30.0"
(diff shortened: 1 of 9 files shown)


Cargo.toml
```diff
@@ -66,15 +66,16 @@ sqlx = { version = "0", features = [
 ] }
 helium-anchor-gen = { git = "https://github.com/helium/helium-anchor-gen.git" }
 helium-crypto = { version = "0.8.4", features = ["sqlx-postgres", "multisig"] }
+helium-lib = { git = "https://github.com/helium/helium-wallet-rs.git", branch = "master" }
```
abhay, line 69: time to extract `helium-lib` out of wallet? maybe push into `helium-program-library`?

madninja: probably not? The whole point is that compiling all of HPL is very brittle and complicated.. helium-lib only relies on the IDLs for the HPL defined programs

price/src/settings.rs
```diff
@@ -59,19 +32,13 @@ pub struct Settings {
     /// Tick interval (secs). Default = 60s.
     #[serde(with = "humantime_serde", default = "default_interval")]
     pub interval: Duration,
-    /// Cluster Configuration
-    #[serde(default)]
-    pub cluster: ClusterConfig,
+    pub tokens: Vec<TokenSetting>,
+    // /// Cluster Configuration
+    // #[serde(default)]
+    // pub cluster: ClusterConfig,
```
maplant, line 38: remove this 
```suggestion
```

bbalser: good catch, removed

PR #835, opened 2024-06-27
HIP-125

bbalser:
https://github.com/helium/HIP/blob/main/0125-temporary-anti-gaming-measures-for-boosted-hexes.md

Proto changes: https://github.com/helium/proto/pull/405


coverage_point_calculator/src/lib.rs
```diff
@@ -262,16 +270,23 @@ impl RadioType {
...
+// pub enum RadioThreshold {
+//     Verified,
+//     Unverified,
+// }
+
+// impl RadioThreshold {
+//     fn is_met(&self) -> bool {
+//         matches!(self, Self::Verified)
+//     }
+// }
```
maplant, line 289:
```suggestion
```

bbalser: good catch, removed!

mobile_verifier/src/sp_boosted_rewards_bans.rs
```diff
@@ -0,0 +1,654 @@
...
+    ) -> anyhow::Result<()> {
+        let report = BannedRadioReport::try_from(ingest.clone())?;
+        let is_authorized = self.is_authorized(&report.pubkey).await?;
+
+        if is_authorized {
+            db::update_report(transaction, &report).await?;
+            self.update_seniority(transaction, &report).await?;
+        }
+
+        let status = match is_authorized {
```
maplant, line 247: I do think that an if statement is cleaner here

mobile_verifier/src/sp_boosted_rewards_bans.rs
```diff
@@ -0,0 +1,654 @@
...
+impl BannedRadios {
+    pub fn insert_wifi(&mut self, pubkey: PublicKeyBinary) {
+        self.wifi.insert(pubkey);
+    }
+
+    pub fn insert_cbrs(&mut self, cbsd_id: String) {
+        self.cbrs.insert(cbsd_id);
+    }
+
+    pub fn contains(&self, pubkey: &PublicKeyBinary, cbsd_id_opt: Option<&str>) -> bool {
```
maplant, line 108: maybe make this RadioKey

mobile_verifier/src/sp_boosted_rewards_bans.rs
```diff
@@ -0,0 +1,654 @@
...
+    use std::str::FromStr;
+
+    use chrono::Duration;
+    use sqlx::Row;
+
+    use super::*;
+
+    pub async fn get_banned_radios(
+        pool: &PgPool,
+        date_time: DateTime<Utc>,
```
maplant, line 315: This should be constrained to an epoch, otherwise the behavior could be non-deterministic. 
(i.e. a radio banned after an epoch could cause it to lose out on rewards in the current epoch). 
I don't think that behavior is bad, but I think we should consider avoiding non-deterministic behavior.

mobile_verifier/src/sp_boosted_rewards_bans.rs
```diff
@@ -0,0 +1,654 @@
...
+    ) -> anyhow::Result<BannedRadios> {
+        sqlx::query(
+            r#"
+                SELECT radio_type, radio_key
+                FROM sp_boosted_rewards_bans
+                WHERE until > $1 
+                    AND (invalidated_at > $1 OR invalidated_at IS NULL)
```
maplant, line 322: you can use coalesce to simplify this: 
```suggestion
                    AND COALESCE(invalidated_at > $1, TRUE)
```

mobile_verifier/src/seniority.rs
```diff
@@ -0,0 +1,206 @@
...
+    ) -> Self {
+        Self {
+            key,
+            heartbeat_ts,
+            uuid,
+            action,
+        }
+    }
+
+    pub fn from(
```
maplant, line 67: you shouldn't have a method named from. That's confusing

mobile_verifier/src/sp_boosted_rewards_bans.rs
bbalser: Addressed 15874cb

mobile_verifier/src/sp_boosted_rewards_bans.rs
bbalser: Addressed 15874cb

mobile_verifier/src/sp_boosted_rewards_bans.rs
bbalser: addressed 15874cb

mobile_verifier/src/seniority.rs
bbalser: fixed 15874cb

coverage_point_calculator/src/lib.rs
```diff
@@ -115,7 +115,7 @@ pub struct CoveragePoints {
     /// Input Radio Type
     pub radio_type: RadioType,
     /// Input RadioThreshold
```
michaeldjeffrey, line 117:
```suggestion
    /// Input ServiceProviderBoostedRewardEligibility
```

ingest/src/server_mobile.rs
```diff
@@ -337,6 +340,34 @@ impl poc_mobile::PocMobile for GrpcServer {
...
+        let event = request.into_inner();
+
+        custom_tracing::record_b58("pub_key", &event.pub_key);
+
+        let report = self
+            .verify_public_key(event.pub_key.as_ref())
+            .and_then(|public_key| self.verify_network(public_key))
+            .and_then(|public_key| self.verify_signature(public_key, event))
+            .map(
+                |(_, event)| ServiceProviderBoostedRewardsBannedRadioIngestReportV1 {
```
michaeldjeffrey, line 358: I know this didn't change in this PR.
`public_key` is always ignored when returned from `verify_signature/2`.
Do we want to remove it altogether?

michaeldjeffrey (approved):

ingest/src/server_mobile.rs
bbalser: good idea, but I will leave it for another PR

file_store/src/file_info.rs
```diff
@@ -209,6 +212,8 @@ pub enum FileType {
     InvalidatedRadioThresholdReq,
     InvalidatedRadioThresholdIngestReport,
     VerifiedInvalidatedRadioThresholdIngestReport,
+    ServiceProviderBoostedRewardsBannedRadioIngestReport,
+    VerifiedServiceProviderBoostedRewardsBannedRadioIngestReport,
```
andymck, line 216: as above drop ServiceProvider in favour of SP?

coverage_point_calculator/src/lib.rs
```diff
@@ -114,8 +114,8 @@ pub struct CoveragePoints {
     pub speedtest_multiplier: Decimal,
     /// Input Radio Type
     pub radio_type: RadioType,
-    /// Input RadioThreshold
-    pub radio_threshold: RadioThreshold,
+    /// Input ServiceProviderBoostedRewardEligibility
+    pub service_provider_boosted_reward_eligibility: ServiceProviderBoostedRewardEligibility,
```
andymck, line 118: This naming convention is very verbose, how about replacing the ServiceProvider chunk with SP ?

andymck (commented):
On a first pass looks good other than a couple of naming convention nits. Will take another pass

coverage_point_calculator/src/lib.rs
bbalser: I actually think the naming is helpful in the coverage point calculator

PR #839, opened 2024-07-09
Add verification_mapping proto

macpie:


mobile_verifier/src/reward_shares.rs
```diff
@@ -236,7 +247,32 @@ impl MapperShares {
...
                     },
                 )
+            });
+
+        let verified_rewards = self
+            .verified_mapping_event_shares
+            .into_iter()
+            .map(move |verified_share| proto::SubscriberReward {
+                subscriber_id: verified_share.subscriber_id,
+                discovery_location_amount: (Decimal::from(verified_share.total_reward_points)
```
andymck, line 257: We should have a separate entry for verification mapping rather than including all in discover_location_amount.  Ie we should have a second entry in subscriber reward such as verification_mapping_amount for which the value is the subscribers reward for verification mapping for the epoch

file_store/src/verified_mapping_event.rs
```diff
@@ -0,0 +1,63 @@
...
+use sqlx::Row;
+
+#[derive(Clone, Deserialize, Serialize, Debug, PartialEq)]
+pub struct VerifiedSubscriberMappingEvent {
+    pub subscriber_id: Vec<u8>,
+    pub total_reward_points: u64,
+    pub timestamp: DateTime<Utc>,
+}
+
+impl MsgDecode for VerifiedSubscriberMappingEvent {
```
andymck, line 17: I dont see any filestore support for the ingest report ( which is a wrapper around the ingestor received request ) ie this msg in my original proto branch
https://github.com/helium/proto/blob/andymck/subscriber-mapping-support/src/service/poc_mobile.proto#L228

ingest/src/server_mobile.rs
```diff
@@ -484,6 +548,17 @@ pub async fn grpc_server(settings: &Settings) -> Result<()> {
         .create()
         .await?;
 
+    let (subscriber_mapping_event_sink, subscriber_mapping_event_server) =
+        file_sink::FileSinkBuilder::new(
+            FileType::VerifiedSubscriberMappingEvent,
```
andymck, line 553: The filetype written to this sink should be the ingest report ( the wrapper around the original request )

mobile_verifier/src/reward_shares.rs
```diff
@@ -201,11 +207,15 @@ impl MapperShares {
...
         let discovery_mappers_count = Decimal::from(self.discovery_mapping_shares.len());
 
+        let verified_mapping_event_count = Decimal::from(self.verified_mapping_event_shares.len());
+
         // calculate the total eligible mapping shares for the epoch
         // this could be simplified as every subscriber is awarded the same share
         // however the function is setup to allow the verification mapper shares to be easily
         // added without impacting code structure ( the per share value for those will be different )
-        let total_mapper_shares = discovery_mappers_count * DISCOVERY_MAPPING_SHARES;
+        let total_mapper_shares =
```
andymck, line 216: I believe this is incorrect.  Prior to these changes this code dealt solely with discovery mapping. As per Hip79 each discovery mapping subscriber is allocated 30 reward points.  You would need to get a sum of total shares first for discovery mapping ( as per orig code ) and then a sum of total shares for verification mapping ( sum of all verification mapping rewards for the epoch ). total_mapper_shares then would equal: total_disovery_mapping_shares + total_verification_mapping_shares

file_store/src/verified_mapping_event.rs
```diff
@@ -0,0 +1,63 @@
...
+use chrono::{DateTime, Utc};
+use helium_proto::services::poc_mobile::VerifiedSubscriberMappingEventV1;
+use serde::{Deserialize, Serialize};
+use sqlx::Row;
+
+#[derive(Clone, Deserialize, Serialize, Debug, PartialEq)]
+pub struct VerifiedSubscriberMappingEvent {
+    pub subscriber_id: Vec<u8>,
+    pub total_reward_points: u64,
+    pub timestamp: DateTime<Utc>,
```
andymck, line 14: for completeness & transparency you should carry the pubkey through to this msg.  That fits with existing convention of ingestor received requests where the only field dropped tends to be the signature

file_store/src/verified_mapping_event.rs
```diff
@@ -0,0 +1,63 @@
...
+    traits::{MsgDecode, MsgTimestamp, TimestampDecode, TimestampEncode},
+    Error, Result,
+};
+use chrono::{DateTime, Utc};
+use helium_proto::services::poc_mobile::VerifiedSubscriberMappingEventV1;
+use serde::{Deserialize, Serialize};
+use sqlx::Row;
+
+#[derive(Clone, Deserialize, Serialize, Debug, PartialEq)]
+pub struct VerifiedSubscriberMappingEvent {
```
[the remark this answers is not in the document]
macpie: Will be matching that branch more closely

ingest/src/server_mobile.rs
```diff
@@ -81,6 +67,57 @@ fn make_span(_request: &http::request::Request<helium_proto::services::Body>) ->
...
+            address,
+            api_token,
+        }
+    }
+
+    pub async fn run(self, shutdown: triggered::Listener) -> anyhow::Result<()> {
+        let api_token = self.api_token.clone();
+        let address = self.address;
+
+        transport::Server::builder()
```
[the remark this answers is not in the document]
macpie: easier to test to I can just spawn the run function

file_store/src/verified_mapping_event.rs
macpie: Will add

file_store/src/file_info.rs
```diff
@@ -160,6 +160,8 @@ pub const LANDTYPE_DATA_SET: &str = "landtype";
 pub const SP_BOOSTED_REWARDS_BANNED_RADIO: &str = "service_provider_boosted_rewards_banned_radio";
 pub const VERIFIED_SP_BOOSTED_REWARDS_BANNED_RADIO: &str =
     "verified_service_provider_boosted_rewards_banned_radio";
+pub const VERIFIED_SUBSCRIBER_MAPPING_EVENT_REPORT: &str =
```
andymck, line 163: Looks like you are missing a msg type.   There should be a definition for the report received by the ingestor ( which is your VERIFIED_SUBSCRIBER_MAPPING_EVENT_REPORT type ) and then there should be a definition for the report produced by the ingestor and consumed by the mobile verifier ( which is defined as VERIFIED_SUBSCRIBER_MAPPING_EVENT_INGEST_REPORT in my 846 pr )

file_store/src/verified_subscriber_mapping_event.rs
```diff
@@ -0,0 +1,70 @@
...
+        Ok(Self {
+            subscriber_id: v.subscriber_id,
+            total_reward_points: v.total_reward_points,
+            timestamp,
+            verification_mapping_pubkey: v.verification_mapping_pubkey.into(),
+        })
+    }
+}
+
+impl sqlx::FromRow<'_, sqlx::postgres::PgRow> for VerifiedSubscriberMappingEvent {
```
andymck, line 61: for better or worse we have tended not to include db/sqlx requirements in the filestore.  Sql support in filestore is an optional feature and thus baking these may lead to issues

mobile_verifier/src/reward_shares.rs
```diff
@@ -206,9 +212,19 @@ impl MapperShares {
         // however the function is setup to allow the verification mapper shares to be easily
         // added without impacting code structure ( the per share value for those will be different )
         let total_mapper_shares = discovery_mappers_count * DISCOVERY_MAPPING_SHARES;
```
andymck, line 214:
```suggestion
        let total_disco_mapper_shares = discovery_mappers_count * DISCOVERY_MAPPING_SHARES;
```

file_store/src/verified_subscriber_mapping_event.rs
macpie: removed

file_store/src/file_info.rs
macpie: I think the name is wrong but one type is enough?

andymck: Having looked at this again you are correct, we only need one type which is the VERIFIED_SUBSCRIBER_MAPPING_EVENT_INGEST_REPORT ( which is the wrapped mapping event report produced by the mobile ingestor and consumed by the mobile verifier )

In the past we have added a type for the report submitted to the mobile ingestor but as we never consume those reports over the info poller they are actually redundant and never used.  So the convention we have used in the past is actually incorrect in this instance.

Good with one type

mobile_verifier/src/reward_shares.rs
andymck: I think you may have misunderstood my suggestion above.  Rather than producing two SubscriberRewards ( one for disco mapping and then one for verification mapping ) the two should be combined into a single SubscriberReward but with distinct fields within the proto.  The subscriber reward proto itself would be updated to something like:

message subscriber_reward {
  // id of the subscriber to which the reward will be credited
  bytes subscriber_id = 1;
  /// Amount in bones credited to the subscriber for location sharing
  uint64 discovery_location_amount = 2;
  /// Amount in bones credited to the subscriber for verification mapping 
  uint64 verification_mapping_amount = 3;
}

Cargo.toml
```diff
@@ -133,3 +133,6 @@ sqlx = { git = "https://github.com/helium/sqlx.git", rev = "92a2268f02e0cac6fccb
 # [patch.'https://github.com/helium/proto']
 # helium-proto = { path = "../proto" }
 # beacon = { path = "../proto/beacon" }
+# Temporary patch
```
andymck, line 136: remember to remove/comment out before merging

ingest/tests/common/mod.rs
```diff
@@ -0,0 +1,131 @@
...
+use std::{net::SocketAddr, sync::Arc};
+use tokio::{net::TcpListener, sync::mpsc::Receiver};
+use tonic::{
+    metadata::{Ascii, MetadataValue},
+    transport::Channel,
+    Request,
+};
+use triggered::Trigger;
+
+pub async fn setup_mobile() -> anyhow::Result<(
```
andymck, line 21: any hard reason not to follow the existing test harness setup for the iot ingest ?  It would be preferable to have a common approach across both iot and mobile

mobile_verifier/src/reward_shares.rs
andymck: Still think this is relevant, line 217 is not calculating the total_mapper_shares ( ie total shares for mapping ) but instead is calculating the total discovery mapping shares.  Keeping it as total_mapper_shares is a misleading var name

ingest/tests/common/mod.rs
macpie: It seemed cleaner, especially that the mobile stuff needs a lot of file sinks

mobile_verifier/src/reward_shares.rs
macpie: Sorry I think it got lost in my rebase

mobile_verifier/src/reward_shares.rs
macpie: I think this https://github.com/helium/oracles/pull/839/commits/63dad9aa2d02cdf74220663e53e38f95482611fa does it?

mobile_verifier/src/reward_shares.rs
```diff
@@ -220,19 +236,63 @@ impl MapperShares {
...
+                * reward_per_share)
+                .round_dp_with_strategy(0, RoundingStrategy::ToZero)
+                .to_u64()
+                .unwrap_or_default();
+
+            if verification_mapping_amount > 0 {
+                subscriber_rewards
+                    .entry(verified_share.subscriber_id.clone())
+                    .and_modify(|reward| {
+                        reward.verification_mapping_amount += verification_mapping_amount;
```
andymck, line 274: Should this replace the default value rather than adding to it ? I get that the default value will be zero but logically it doesnt seem right to add instead of replace

Other than that this latest change looks good

macpie: Idk to be honest what should be the expected behavior at this point. I guess it depends if we can receive more than one but I think I remember that normally it's not possible.

andymck: verified_mapping_event_shares will contain only a single entry per subscriber as per https://github.com/helium/oracles/blob/63dad9aa2d02cdf74220663e53e38f95482611fa/mobile_verifier/src/verified_subscriber_mapping_event.rs#L126

So whilst the ingestor should only receive a single verified mapping event per epoch, should for whatever reason more than one be received they are aggregated per subscriber in the above into a single entry

mobile_verifier/src/reward_shares.rs
```diff
@@ -924,11 +992,17 @@ mod test {
         // get the summed rewards allocated to subscribers for discovery location
         let mut allocated_mapper_rewards = 0_u64;
         for (reward_amount, subscriber_share) in
-            mapping_shares.into_subscriber_rewards(&epoch, rewards_per_share)
+            shares.into_subscriber_rewards(&epoch, rewards_per_share)
         {
             if let Some(MobileReward::SubscriberReward(r)) = subscriber_share.reward {
-                assert_eq!(expected_reward_per_subscriber, r.discovery_location_amount);
-                assert_eq!(reward_amount, r.discovery_location_amount);
+                assert_eq!(
```
andymck, line 998: can we extend this test so that it also asserts on the individual discovery_location_amount and the verification_mapping_amount along with the existing reward totals

andymck (commented):
I think this looks good and is close, just a couple of nits.  One thing to note is we should not merge this until the appropriate time as it changes existing reward outputs

Cargo.toml
```diff
@@ -133,3 +133,6 @@ sqlx = { git = "https://github.com/helium/sqlx.git", rev = "92a2268f02e0cac6fccb
...
+[patch."https://github.com/helium/proto"]
```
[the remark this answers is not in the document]
macpie: I can try again but last time I set to a specific branch it would not update properly, now that the proto is more settled it might be better. I will attempt.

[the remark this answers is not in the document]
michaeldjeffrey: I added this comment. The intent was to patch the proto lib while still developing the definitions. 
To reduce the `update proto -> push proto -> update oracles -> repeat` cycle. 
Then when things are in a good spot, you replace the patch with a pointer to the proto branch.

It might be a good idea to put a comment towards that end in here to help ward off this exact scenario.

ingest/tests/common/mod.rs
```diff
@@ -0,0 +1,134 @@
...
+        TestClient {
+            client,
+            key_pair: Arc::new(key_pair),
+            authorization: format!("Bearer {}", api_token).try_into().unwrap(),
+            file_sink_rx,
+        }
+    }
+
+    pub async fn recv(mut self) -> anyhow::Result<Vec<u8>> {
+        match self.file_sink_rx.recv().await {
```
michaeldjeffrey, line 93: May want to add a timeout so the test cannot hang indefinitely if something goes wrong. `mobile_verifier/tests/integrations/common/mod.rs` has an a example.

mobile_verifier/src/reward_shares.rs
```diff
@@ -209,9 +215,19 @@ impl MapperShares {
         // however the function is setup to allow the verification mapper shares to be easily
         // added without impacting code structure ( the per share value for those will be different )
         let total_mapper_shares = discovery_mappers_count * DISCOVERY_MAPPING_SHARES;
+
+        let total_verified_mapping_event_shares: Decimal = self
```
michaeldjeffrey, line 219: I think this code renders some of the above comments stale.

mobile_verifier/src/reward_shares.rs
```diff
@@ -220,19 +236,63 @@ impl MapperShares {
         reward_period: &'_ Range<DateTime<Utc>>,
         reward_per_share: Decimal,
     ) -> impl Iterator<Item = (u64, proto::MobileRewardShare)> + '_ {
-        self.discovery_mapping_shares
+        let mut subscriber_rewards: HashMap<Vec<u8>, proto::SubscriberReward> = HashMap::new();
+
+        // Accumulate rewards from discovery_mapping_shares
+        for subscriber_id in self.discovery_mapping_shares {
+            let discovery_location_amount = (DISCOVERY_MAPPING_SHARES * reward_per_share)
```
michaeldjeffrey, line 243: This can be calculated once?

mobile_verifier/src/verified_subscriber_mapping_event.rs
```diff
@@ -0,0 +1,146 @@
...
+    ) -> anyhow::Result<()> {
+        tracing::info!(
+            "Processing Verified Mapping Event file {}",
+            file.file_info.key
+        );
+        let mut transaction = self.pool.begin().await?;
+        let mut reports = file.into_stream(&mut transaction).await?;
+
+        while let Some(report) = reports.next().await {
+            save_event(&report.report, &mut transaction).await?;
```
bbalser, line 80: We need to verify the `carrier_mapping_key` is a trusted key before saving to database.

bbalser:
Needs to validate the carrier mapping key and subscriber id and write out a report to the output bucket with the report and the status of the report.  Look at subscriber location as an example.

https://github.com/helium/oracles/blob/341b6e151c63e852fe2a64a06125e62d9599f9b6/mobile_verifier/src/subscriber_location.rs#L123

mobile_verifier/src/verified_subscriber_mapping_event.rs
```diff
@@ -0,0 +1,146 @@
...
+    sqlx::query(
+        r#"
+        insert into verified_mapping_event (subscriber_id, total_reward_points, timestamp)
+        values ($1, $2, $3)
+        on conflict (subscriber_id, timestamp) do nothing
+        "#,
+    )
+    .bind(&event.subscriber_id)
+    .bind(event.total_reward_points as i64)
+    .bind(event.timestamp)
```
bbalser, line 110: shouldn't this be the `received_timestamp` of the ingest report?

mobile_verifier/src/rewarder.rs
```diff
@@ -9,7 +9,7 @@ use crate::{
     },
     sp_boosted_rewards_bans, speedtests,
     speedtests_average::SpeedtestAverages,
-    subscriber_location, telemetry, Settings,
+    subscriber_location, telemetry, verified_subscriber_mapping_event, Settings,
```
macpie, line 12: Line 295 we need to cleanup data older than 7 days.

mobile_verifier/src/verified_subscriber_mapping_event.rs
macpie: done

mobile_verifier/src/verified_subscriber_mapping_event.rs
macpie: done

mobile_verifier/src/subscriber_verified_mapping_event.rs
```diff
@@ -0,0 +1,292 @@
...
+        "SELECT 
+            subscriber_id, 
+            SUM(total_reward_points) AS total_reward_points
+        FROM 
+            verified_subscriber_verified_mapping_event
+        WHERE received_timestamp >= $1 AND received_timestamp < $2
+        GROUP BY 
+            subscriber_id;",
+    )
+    .bind(reward_period.end - Duration::days(SUBSCRIBER_REWARD_PERIOD_IN_DAYS))
```
bbalser, line 273: this can be `reward_period.start`

mobile_verifier/migrations/35_verified_subscriber_verified_mapping_event.sql
```diff
@@ -0,0 +1,7 @@
+CREATE TABLE IF NOT EXISTS verified_subscriber_verified_mapping_event (
```
bbalser, line 1: NIT: table doesn't need the `verified` prefix since all records in the table will be verified.

bbalser (approved):

PR #840, opened 2024-07-10
HIP-119: Location Trust Score from maximum asserted distance difference

michaeldjeffrey:
[HIP-119](https://github.com/helium/HIP/blob/main/0119-closing-gaming-loopholes-within-the-mobile-network.md#maximum-asserted-distance-difference) introduces a new way for a radio to be ineligible for receiving Service Provider boosted rewards.

When validating a Heartbeat, a new set of distances for [indoor](indoor-table) and [outdoor](outdoor-table) radios are used to determine a Location Trust Multiplier.

### `BoostedHexStatus` has a new Status!!

`BoostedHexStatus::AverageAssertedDistanceOverLimit(Decimal)` means a Radio is not eligible to receive service provider boosted rewards because the average distance heartbeat locations from the radio's asserted location is greater than 50 meters.

#### Average vs. Maximum

"Maximum Asserted Distance for Boosted Hexes" is the heading used in HIP-119. A followup question was asked in the #hip-119-closing-gaming-loopsholes-within-the-mobile-network Discord channel.

> "[...] is that the average of distance_to_asserted over the day or the max distance_to_asserted?"

And was answered by the HIP author.

> "I’ll leave that up to you guys, whichever you think is the better approach"

This implementation has chosen to go with the average distance from an asserted location being greater than 50 meters, rather than _any_ individual distance from the asserted location being greater than 50 meters.

If a radio spends more time than not within 50 meters of its asserted location, the benefit of the doubt would say it is most likely in that location. A temporary excursion outside of 50 meters through the means of SkyHook or other location reporting services will not be penalized as harshly as possible.

### Service Provider Hex Boosting and Location Trust

It was previously the case that a radio covering a boosted hex would lower the location trust score of any `LocationTrust` that was more then 50 meters from the asserted location.

Now, regardless of covering boosted hexes, `LocationTrust` will not be changed.

A Radio will not receive boosted rewards when greater than 50 meters away from it's asserted location on average. A Wifi Radio will not receive boosted rewards if it's naturally occuring `location_trust_multiplier` falls below `0.75x` as stated in [hip-93][wifi-aps].

### Items and answers

- [valid heartbeats](https://github.com/helium/oracles/pull/840/files#diff-5ded0a45305c5fbd258aba4c38fcbc14b69f0ead2ff5ecebd3c71dd3693fc1deR545), should they completely lower a trust score? `Yes`
- [furthest covered hex](https://github.com/helium/oracles/pull/840/files#diff-5ded0a45305c5fbd258aba4c38fcbc14b69f0ead2ff5ecebd3c71dd3693fc1deR549), should a single hex exceeding the `max_distance_to_coverage` setting completely lower a trust score? `Yes`
- Do you think I could pull off [this jacket](https://a.co/d/0c0iNLNF) `Resounding Yes`

[indoor-table]: https://github.com/helium/HIP/blob/main/0119-closing-gaming-loopholes-within-the-mobile-network.md#indoor-access-points
[outdoor-table]: https://github.com/helium/HIP/blob/main/0119-closing-gaming-loopholes-within-the-mobile-network.md#outdoor-access-points
[wifi-aps]: https://github.com/helium/HIP/blob/main/0093-addition-of-wifi-aps-to-mobile-subdao.md#341-indoor-access-points-rewards


coverage_point_calculator/src/lib.rs
```diff
@@ -219,27 +230,36 @@ impl CoveragePoints {
 pub enum BoostedHexStatus {
     Eligible,
     WifiLocationScoreBelowThreshold(Decimal),
+    AverageAssertedDistanceOverLimit(Decimal),
     RadioThresholdNotMet,
     ServiceProviderBanned,
 }
 
 impl BoostedHexStatus {
     fn new(
```
bbalser, line 239: If the radio is not eligible due to `service_provider_banned` or ` radio_threshold_not_met` should those be returned regardless of `location_trust_multiplier` and `average_asserted_distance_over_limit`?

bbalser:
The answer to all 3 questions is yes.  If no `location_validate_timestamp` is present or the max_distance to coverage_object is > 3000m than the `location_trust_score_multiplier` should be 0.

And that jacket would look great!

coverage_point_calculator/src/lib.rs
michaeldjeffrey: I agree. 
Active reasons for not receiving boosted rewards should take precedence over passive reasons.
updated [#3d7a1cd](https://github.com/helium/oracles/pull/840/commits/3d7a1cd2582d1d44352282930f370c94caa6bb00)

coverage_point_calculator/src/service_provider_boosting.rs
```diff
@@ -0,0 +1,37 @@
...
+// less than the reported location from external services for both indoor
+// and outdoor Access Points.
+//
+// [HIP-119: Gaming Loopholes][gaming-loopholes]
+//
+// [gaming-loopholes]: https://github.com/helium/HIP/blob/main/0119-closing-gaming-loopholes-within-the-mobile-network.md#maximum-asserted-distance-for-boosted-hexes
+pub(crate) const MAX_AVERAGE_DISTANCE: Decimal = dec!(50);
+
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub enum ServiceProviderBoostedRewardEligibility {
```
andymck, line 23: nit: how about replacing with `SPBoostedRewardEligibility` to reduce verbosity

mobile_verifier/src/heartbeats/mod.rs
```diff
@@ -538,17 +537,35 @@ impl ValidatedHeartbeat {
...
-		// The heartbeat location to asserted location must be less than the max_distance_to_asserted value:
-                    && distance_to_asserted <= max_distance_to_asserted as i64
-		// The heartbeat location to every associated coverage hex must be less than max_distance_to_coverage:
-		    && coverage_object.max_distance_m(hb_latlng).round() as u32 <= max_distance_to_coverage
-                {
-                    dec!(1.0)
+                let max_distance = coverage_object.max_distance_m(hb_latlng).round() as u32;
+
+                let location_trust_score_multiplier = if !is_valid {
+                    // QUESTION: Invalid heartbeats dropped the location trust multiplier to 0.25x.
```
andymck, line 545: remember to remove these questions given the answers have been confirmed

coverage_point_calculator/src/location.rs
```diff
@@ -17,20 +11,42 @@ pub struct LocationTrust {
...
+///
+/// [HIP-119: Gaming Loopholes][gaming-loopholes]
+///
+/// [gaming-loopholes]: https://github.com/helium/HIP/blob/main/0119-closing-gaming-loopholes-within-the-mobile-network.md#maximum-asserted-distance-difference
+pub fn asserted_distance_to_trust_multiplier(
+    radio_type: RadioType,
+    meters_to_asserted: Meters,
+) -> Decimal {
+    match radio_type {
+        RadioType::IndoorWifi | RadioType::IndoorCbrs => match meters_to_asserted {
```
andymck, line 24: my reading of hip119 is that the trust score change applies to wifi radios only ( see phase 1 of implementation phases ). Has there been confirmation these trust scores based on specified distances should be applied to cbrs radios?

coverage_point_calculator/src/service_provider_boosting.rs
michaeldjeffrey: Also to match the proto

[861bd8f](https://github.com/helium/oracles/pull/840/commits/861bd8f227b515ff1ebb299c93a3a0d3e26b28c5)

mobile_verifier/src/heartbeats/mod.rs
michaeldjeffrey: Removed, thank you

coverage_point_calculator/src/location.rs
michaeldjeffrey: Indeed it does say that. 
CBRS radios are intercepted earlier when validating heartbeats, but I've updated this function to return `dec!(1.0)` for CBRS if somehow they make it here.
[c3e8194](https://github.com/helium/oracles/pull/840/commits/c3e81944b6d3b1b82253ce426551d794a74a5e7d)

bbalser (approved):

coverage_point_calculator/src/location.rs
andymck: yeah the change also helps to convey the intent better to the reader as well as aligning better with the hip.  better for long term understanding

PR #843, opened 2024-07-16
Integrate mobile rewards v2

michaeldjeffrey:
Proto PR: https://github.com/helium/proto/pull/407

- Both `Reward::RadioReward` and `Reward::RadioRewardV2` will be written out for 90 days.
- Integration tests now use v2 reward structs. The mock file sink returns only the v2 reward struct, internally it pulls both from the stream to ensure `poc_reward` match between the two.
- `poc_mobile::oracle_boosting_hex_assignment::Assignment` has been moved a top level enum at  `poc_mobile::OracleBoostingAssignment` because Assignments are listed in the v2 radio reward.


mobile_verifier/src/rewarder.rs
```diff
@@ -450,13 +450,20 @@ async fn reward_poc(
         {
             // handle poc reward outputs
             let mut allocated_poc_rewards = 0_u64;
-            for (poc_reward_amount, mobile_reward_share) in mobile_reward_shares {
+            for (poc_reward_amount, mobile_reward_share_v1, mobile_reward_share_v2) in
```
andymck, line 453: Throwing this out there for a possible future refactor....it might be a more tidier solution to have coverage_share.into_rewards return only the v2 reward then then have a conversion handler to go from a v2 to a v1.  That conversion would be done below as part of writing out the reward shares.

andymck (approved):

mobile_verifier/src/rewarder.rs
michaeldjeffrey: I spent a bit of time trying to make this happen. Once `rust_decimal::Decimal` goes into `helium_proto::Decimal` it's wrapped in an `Option` and pulling all the numbers back out again to make a v1 mobile reward was very uncomfortable. We would also lose the ability to track down any discrepancies between two if any arose.

It was a good exercise though.

bbalser (approved):

PR #853, opened 2024-08-09
[ORA-426] Add speedtest average to radio_reward_v2

kurotych:
Add `speedtest_average` field to the `radio_reward_v2`
Minor: [Fix dividing by zero in Speedtest:avg if array is empty](https://github.com/helium/oracles/pull/853/commits/8b53dfb3b15892cd660be481432b7a566d20b9e5)


bbalser (approved):
LGTM

coverage_point_calculator/src/lib.rs
```diff
@@ -165,12 +167,13 @@ impl CoveragePoints {
         let hex_coverage_points = hexes::calculated_coverage_points(&covered_hexes);
 
         let speedtests = speedtest::clean_speedtests(speedtests);
-        let speedtest_multiplier = speedtest::multiplier(&speedtests);
+        let (speedtest_multiplier, speedtest_avg) = speedtest::multiplier(&speedtests);
```
andymck, line 170: rather than having a function named `multiplier` now return both the multiplier and the average when there is an existing average function, would it be better to just call the average function here first, then change the multiplier function to access that average as a param?

andymck (approved):
Other than one noted nit re multiplier function, lgtm

coverage_point_calculator/src/lib.rs
kurotych: I was thinking that way. 

> then change the multiplier function to access that average as a param?

I'm not sure that it will give us any benefits because `speedtest_avg` has the same type as `speedtest` (pub struct Speedtest).  I don't see how to make this function non-confusing. (Except making new type `SpeedTestAvg` that `multiplier` will accept) 

Also, there is an important check there and I'm not sure that it is worth moving it somewhere. 

```rust
if speedtests.len() < MIN_REQUIRED_SPEEDTEST_SAMPLES {
        return (dec!(0), avg);
}
```

So it looks safer (but maybe not cleaner) for me, save it as is.

andymck: yeah its all a lil convoluted in there with some of these functions and maybe beyond scope of this PR to address

PR #854, opened 2024-08-09
Fix rare case where the file info poller will re-process old files

maplant:
We found a case where on devnet, the reward indexer would re-process old files after cleaning out the files processed table. This is due to the offset being further back in time than the oldest file in the files processed table due to a burst of files. This PR fixes that by checking for a full files processed table, and if it is full limiting the offset to be at most the oldest timestamp in the table.

### DIFF
--- a/file_store/src/file_info_poller.rs
+++ b/file_store/src/file_info_poller.rs
@@ -25,6 +25,12 @@ pub trait FileInfoPollerState: Send + Sync + 'static {
         file_type: &str,
     ) -> Result<Option<DateTime<Utc>>>;
 
+    async fn oldest_timestamp(
+        &self,
+        process_name: &str,
+        file_type: &str,
+    ) -> Result<Option<DateTime<Utc>>>;
+
     async fn exists(&self, process_name: &str, file_info: &FileInfo) -> Result<bool>;
 
     async fn clean(&self, process_name: &str, file_type: &str) -> Result;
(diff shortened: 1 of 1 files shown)


file_store/src/file_info_poller.rs
```diff
@@ -395,36 +412,58 @@ impl FileInfoPollerState for sqlx::Pool<sqlx::Postgres> {
...
+            WHERE process_name = $1 AND file_type = $2
+            ORDER BY file_timestamp DESC
+            OFFSET $3
+            LIMIT 1
+            "#,
+        )
+        .bind(process_name)
+        .bind(file_type)
+        .bind(MAX_SIZE_OF_FILES_PROCESSED_TABLE - 1)
+        .fetch_one(self)
```
bbalser, line 432: won't this error if there are not 99 entries matching the where clause?

file_store/src/file_info_poller.rs
```diff
@@ -395,36 +412,58 @@ impl FileInfoPollerState for sqlx::Pool<sqlx::Postgres> {
...
+    async fn oldest_timestamp(
+        &self,
+        process_name: &str,
+        file_type: &str,
+    ) -> Result<Option<DateTime<Utc>>> {
+        sqlx::query_scalar(
+            r#"
+            SELECT file_timestamp FROM files_processed
+            WHERE process_name = $1 AND file_type = $2
+            ORDER BY file_timestamp DESC
```
bbalser, line 424: This also needs to be sorted ASC to get the value you want.

file_store/src/file_info_poller.rs
maplant: Ah I meant to do fetch optional then

bbalser (approved):

bbalser:
This PR still exhibits a bug that could occur when the cleaner runs 12 hours after boot.  Closing in favor a solution to only modifies the cleaning algorithm.

PR #855, opened 2024-08-21
`FileInfoPollerServer` lookback exceeds limit leads to reprocessing files

michaeldjeffrey:
## [Some Background](https://www.freepik.com/free-photos-vectors/rust-background)

A `FileInfoPollerServer` keeps track of processed files in a db table. To keep from growing forever, this table is cleaned when   a `FileInfoPollerServer` is started, and every 12 hours afterwards, to the most recent 100 processed files. For most cases, this is well and good.

To fetch files for processing, the timestamp of the latest processed file and an `offset` Duration is used to determine how far back to look.

## [The Situation](https://en.wikipedia.org/wiki/Michael_Sorrentino)

A burst of files arrived on devnet exceeding the 100 row limit. And they were processed. It was good.

For sake of this tale, let's say 150 files arrived in 10 minutes. And the `offset` provided was 1 hour.

Later, the cleanup job came around (or when the service was restarted) and removed the 50 that were over the limit. 

The next time the `FileInfoPollerServer` was asked for some files to process, it took the latest timestamp, added the `offset` and requested those files from s3. It pulled down everything from the last hour.

However, the first 50 files it pulled down that were already processed were no longer in the database because of the cleanup job. So they were dutifully re-processed. It was bad.

## [A Solution](https://www.britannica.com/science/solution-chemistry)

When the cleanup job runs, either on interval or startup, we find the timestamp of the row at the allowed limit (in this case, still 100). 

If the `offset` provided is earlier than the 100th item (`t100`), the cleanup job will remove rows older than the `offset`.
If `t100` is older, rows older than it are removed.

In this way, files that may be fetched for processing should always exist in the database until there is no longer a chance they will be fetched. And we keep around some history for debugging.

### DIFF
--- a/file_store/src/file_info_poller.rs
+++ b/file_store/src/file_info_poller.rs
@@ -5,6 +5,7 @@ use derive_builder::Builder;
 use futures::{future::LocalBoxFuture, stream::BoxStream, StreamExt};
 use futures_util::TryFutureExt;
 use retainer::Cache;
+use sqlx::postgres::PgQueryResult;
 use std::{collections::VecDeque, marker::PhantomData, sync::Arc, time::Duration};
 use task_manager::ManagedTask;
 use tokio::sync::mpsc::{Receiver, Sender};
@@ -27,7 +28,13 @@ pub trait FileInfoPollerState: Send + Sync + 'static {
 
     async fn exists(&self, process_name: &str, file_info: &FileInfo) -> Result<bool>;
 
-    async fn clean(&self, process_name: &str, file_type: &str) -> Result;
+    // Returns number of items cleaned
+    async fn clean(
+        &self,
+        process_name: &str,
+        file_type: &str,
+        offset: DateTime<Utc>,
+    ) -> Result<u64>;
 }
 
 #[async_trait::async_trait]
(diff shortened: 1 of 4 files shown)


file_store/src/file_info_poller.rs
```diff
@@ -408,23 +467,208 @@ impl FileInfoPollerState for sqlx::Pool<sqlx::Postgres> {
...
+        offset: DateTime<Utc>,
+    ) -> Result<u64> {
+        let t100_timestamp: Option<DateTime<Utc>> = sqlx::query_scalar(
+            r#"
+                SELECT file_timestamp
+                FROM files_processed
+                WHERE process_name = $1
+                    AND file_type = $2
+                ORDER BY file_timestamp DESC
+                LIMIT 1 OFFSET 100;
```
andymck, line 483: Do we actually care about the 100 limit now, is finding the above timestamp of any value ?  Why not eliminate that and purge anything outside of the offset window?

michaeldjeffrey: Depending on the how often a given service processes files, keeping around some amount of history past the offset window allows the offset window to be increased while potentially forgetting to provide a new `StartAfter` lookback value. 

I don't think it's a great foot gun shield. But it is an added benefit along with more history during debugging.

andymck (approved):

bbalser (approved):

PR #856, opened 2024-08-23
HIP 114: Service Provider Rewards

maplant:
## Abstract

This HIP allows for service providers to allocate some percent of their dedicated rewards to promotions, which are received via the ingest bucket.

At reward time, any unallocated rewards left in the service provider bucket will be used to match the promotion rewards, up to 100%. If there is not enough unallocated rewards to match all of the promotion rewards, they will be scaled.

## Unallocated Reward Scaling

We scale the unallocated rewards by the amount of total rewards given by the service providers. Therefore, if a service provider rewards their entities a total of half of all promotion rewards given, they will be given half of the unallocated rewards regardless of their percentage. 

### Examples:

| Service provider: | Rewards Given: | Percent: | Promotion Rewards: | % of unallocated rewards given: | 
| -- | -- | -- | -- | -- |
| Sp1 | 12 | 50% | 6 | 2/3 |
| Sp2 | 6 | 50% | 3 | 1/3 | 

| Service provider: | Rewards Given: | Percent: | Promotion Rewards: | % of unallocated rewards given: | 
| -- | -- | -- | -- | -- |
| Sp1 | 12 | 50% | 6 | 1/2 |
| Sp2 | 6 | 100% | 6 | 1/2 | 

| Service provider: | Rewards Given: | Percent: | Promotion Rewards: | % of unallocated rewards given: | 
| -- | -- | -- | -- | -- |
| Sp1 | 10 | 100% | 10 | 1/3 |
| Sp2 | 1,000 | 2% | 20 | 2/3 | 

## Things left to complete in this PR:

- [x] Testing needs to be added and fixed
- [x] Currently, the updates to the Solana chain are in the develop branch and not in master, so we can't deploy this until after that occurs
- [x] Some refactoring needs to be done on the actual rewarding, it's a bit of a mess as it stands right now


bbalser:
I believe you forgot to commit `file_store/lib/promotion_reward.rs`

mobile_verifier/src/promotion_reward.rs
```diff
@@ -0,0 +1,371 @@
...
+    pub fn into_rewards(
+        self,
+        rewards_allocated_for_promotion: HashMap<ServiceProvider, Decimal>,
+        unallocated_sp_rewards: Decimal,
+        reward_period: &'_ Range<DateTime<Utc>>,
+    ) -> impl Iterator<Item = (u64, proto::MobileRewardShare)> + '_ {
+        let total_rewards_allocated: Decimal = rewards_allocated_for_promotion.values().sum();
+        let sp_share_of_unallocated_pool: HashMap<_, _> = rewards_allocated_for_promotion
+            .iter()
+            .map(|(sp, rewards)| (sp, total_rewards_allocated / rewards))
```
bbalser, line 294: shouldn't this be `rewards / total_rewards_allocated`

mobile_verifier/src/promotion_reward.rs
```diff
@@ -0,0 +1,541 @@
...
+
+    async fn save(self, transaction: &mut Transaction<'_, Postgres>) -> anyhow::Result<()> {
+        let (subscriber_id, gateway_key) = match self.promotion_reward.entity {
+            Entity::SubscriberId(sub_id) => (Some(sub_id), None),
+            Entity::GatewayKey(gk) => (None, Some(gk)),
+        };
+        sqlx::query(
+            r#"
+            INSERT INTO promotion_rewards (time_of_reward, subscriber_id, gateway_key, service_provider, shares)
+            VALUES ($1, COALESCE($2, '{}'), COALESCE($3, ''), $4, $5)
```
andymck, line 214: Defining the table with entity_key and entity_type would eliminate the need for the COALESCE usage and make the code that bit more readable and understandable in the future

mobile_verifier/src/promotion_reward.rs
```diff
@@ -0,0 +1,565 @@
...
+        let mut verified_promotion_rewards = pin!(
+            VerifiedPromotionReward::validate_promotion_rewards(reports, &self.carrier_client,)
+        );
+
+        while let Some(promotion_reward) = verified_promotion_rewards.try_next().await? {
+            promotion_reward.write(&self.promotion_rewards_sink).await?;
+            if promotion_reward.is_valid() {
+                promotion_reward.save(&mut transaction).await?;
+            }
+        }
```
michaeldjeffrey, line 125: I'm struggling with "verified" and "validate" in this context. 

If we're getting Verified Promotions by validating promotions it seems weird to ask if `promo.is_valid()` when iterating them.

Maybe `VerifiedPromotionReward::from_promotion_rewards(reports, &self.carrier_client)`? Then reading this usage, it doesn't seem redundant to call `promo.is_valid()` before saving.

mobile_verifier/src/promotion_reward.rs
```diff
@@ -0,0 +1,565 @@
...
+            r#"
+            INSERT INTO promotion_rewards (time_of_reward, entity, entity_type, service_provider, shares)
+            VALUES ($1, $2, $3, $4, $5)
+            ON CONFLICT DO NOTHING
+            "#
+        )
+            .bind(self.promotion_reward.timestamp)
+            .bind(entity)
+            .bind(entity_type)
+            .bind(self.service_provider.map(|x| x as i64))
```
michaeldjeffrey, line 231: `service_provider` is a `NOT NULL` column in the migration. 
I'm not sure if it makes sense to check earlier that the service provider exists, or to remove the option from this struct.

mobile_verifier/tests/integrations/rewarder_sp_rewards.rs
```diff
@@ -70,6 +70,7 @@ async fn test_service_provider_rewards(pool: PgPool) -> anyhow::Result<()> {
     let (_, rewards) = tokio::join!(
```
andymck, line 70: Should add a version of this test with promotion rewards seeded

mobile_verifier/src/reward_shares/mod.rs
```diff
@@ -419,6 +399,66 @@ impl ServiceProviderShares {
...
+    ) -> impl Iterator<Item = proto::MobileRewardShare> + '_ {
+        self.rewards
+            .into_iter()
+            .map(|(service_provider_id, reward)| proto::MobileRewardShare {
+                start_period: reward_period.start.encode_timestamp(),
+                end_period: reward_period.end.encode_timestamp(),
+                reward: Some(ProtoReward::ServiceProviderReward(
+                    proto::ServiceProviderReward {
+                        service_provider_id,
+                        amount: (reward.for_promotions + reward.for_service_provider)
```
andymck, line 446: It is a lil odd to read this and see the SP being rewarded the sum of for_promotions and for_service_provider.  Without being aware of the `take` above it is confusing....I wonder if there is a better way to approach this without things being as obfuscated

mobile_verifier/src/reward_shares/mod.rs
```diff
@@ -419,6 +399,66 @@ impl ServiceProviderShares {
...
+                            .unwrap_or(0),
+                    },
+                )),
+            })
+    }
+}
+
+fn service_provider_id_to_carrier_name(sp: ServiceProvider) -> &'static str {
+    match sp {
+        ServiceProvider::HeliumMobile => "Helium Mobile",
```
andymck, line 458: I do believe "Helium Mobile" is the correct onchain representation it my memory serves me correct but double checked ?

mobile_verifier/src/promotion_reward.rs
maplant: I renamed the struct to "ValidatedPromotionReward" to be more in line with everything else

mobile_verifier/src/reward_shares/mod.rs
```diff
@@ -345,49 +341,33 @@ impl ServiceProviderShares {
...
+            }
+            let percent_for_promotion_rewards = solana
+                .fetch_incentive_escrow_fund_percent(service_provider_id_to_carrier_name(
+                    share.service_provider,
+                ))
+                .await?;
+            rewards.insert(
+                share.service_provider as i32,
+                ServiceProviderReward {
+                    for_promotions: total * percent_for_promotion_rewards,
```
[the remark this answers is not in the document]
maplant: This will get optimized into the same calculation in all likelihood, I think it's fine as is

mobile_verifier/src/promotion_reward.rs
```diff
@@ -0,0 +1,644 @@
...
+    }
+
+    fn validate_promotion_rewards<'a>(
+        promotion_rewards: impl Stream<Item = PromotionReward> + 'a,
+        authorization_verifier: &'a impl AuthorizationVerifier,
+        gateway_info_resolver: &'a impl GatewayResolver,
+        entity_verifier: &'a impl EntityVerifier,
+    ) -> impl Stream<Item = anyhow::Result<Self>> + 'a {
+        promotion_rewards.then(move |promotion_reward| async move {
+            Self::validate(
```
macpie, line 227: I think we do not need the nested function here.

mobile_verifier/src/promotion_reward.rs
```diff
@@ -0,0 +1,644 @@
...
+            pin!(ValidatedPromotionReward::validate_promotion_rewards(
+                reports,
+                &self.authorization_verifier,
+                &self.gateway_info_resolver,
+                &self.entity_verifier
+            ));
+
+        while let Some(promotion_reward) = verified_promotion_rewards.try_next().await? {
+            promotion_reward.write(&self.promotion_rewards_sink).await?;
+            if promotion_reward.is_valid() {
```
macpie, line 132: let see if we can validate here instead of line 122 so it will make code simpler to read

mobile_verifier/src/promotion_reward.rs
```diff
@@ -0,0 +1,644 @@
...
+            .map(|(sp, total_shares)| {
+                if total_promotion_rewards_allocated.is_zero() || total_shares.is_zero() {
+                    (*sp, SpPromotionRewardShares::default())
+                } else {
+                    let rewards_allocated_for_promotion =
+                        sp_rewards.take_rewards_allocated_for_promotion(sp);
+                    let share_of_unallocated_pool =
+                        rewards_allocated_for_promotion / total_promotion_rewards_allocated;
+                    let sp_share = SpPromotionRewardShares {
+                        shares_per_reward: rewards_allocated_for_promotion / total_shares,
```
macpie, line 421:
```suggestion
                        rewards_per_share: rewards_allocated_for_promotion / total_shares,
```

mobile_verifier/src/promotion_reward.rs
```diff
@@ -0,0 +1,644 @@
...
+
+        self.rewards
+            .into_iter()
+            .map(move |reward| {
+                let shares = Decimal::from(reward.shares);
+                let sp_promotion_reward_share = sp_promotion_reward_shares
+                    .get(&reward.service_provider_id)
+                    .copied()
+                    .unwrap_or_default();
+                let service_provider_amount = sp_promotion_reward_share.shares_per_reward * shares;
```
macpie, line 439:
```suggestion
                let service_provider_token = sp_promotion_reward_share.shares_per_reward * shares;
```
Should this be token seems more clear than just `amount`?

mobile_verifier/src/promotion_reward.rs
```diff
@@ -0,0 +1,644 @@
...
+                let matched_amount = (sp_promotion_reward_share.shares_per_matched_reward * shares)
```
macpie, line 440:
```suggestion
                // Goal here is to never match more than the original promotion amount.
                let matched_amount = (sp_promotion_reward_share.shares_per_matched_reward * shares)
```

mobile_verifier/src/promotion_reward.rs
macpie: This looks like to be Mobile Bones.

mobile_verifier/src/promotion_reward.rs
macpie: Fixed

mobile_verifier/src/promotion_reward.rs
macpie: will add

mobile_verifier/src/reward_shares/mod.rs
michaeldjeffrey: I've been trying to think of a way to make this stage clearer as well. So far, nothing straight forward has come to mind. But please let me know if you have any ideas.

PR #858, opened 2024-08-28
Implementation of HIP-131

bbalser:
https://github.com/helium/HIP/blob/main/0131-bridging-gap-between-verification-mappers-and-anti-gaming-measures.md

New Ban_type(PoC) is added and is radio is is banned with type PoC then all assignment_multipliers for it are set to 0.0.


coverage_point_calculator/src/hexes.rs
```diff
@@ -76,7 +77,9 @@ pub(crate) fn clean_covered_hexes(
 
             // hip-103: if a hex is boosted by a service provider >=1x, the oracle
             // multiplier will automatically be 1x, regardless of boosted_hex_status.
-            let assignment_multiplier = if ranked.boosted.is_some() {
```
andymck, line 79: above comment needs an update

andymck (approved):
lgtm

michaeldjeffrey:
`SpBoostedRewardsBannedRadioBanType::BoostedHex` maps directly to `SPBoostedRewardEligibility::ServiceProviderBanned`.

`SpBoostedRewardsBannedRadioBanType::Poc` appears equivalent to not processing the radio for points at all. 

If HIP-131 was more than a stop-gap measure for another HIP, I would like to see it integrated into the calculator as more than a boolean.

But it does introduce the concept of a boosted hex restriction applying to oracle boosting, whereas it was only service provider boosting before. It could be a nice idea to extend `coverage_point_calculator::BoostedHexStatus` to include restriction statuses on oracle boosting.

mobile_verifier/src/reward_shares.rs
```diff
@@ -669,6 +672,7 @@ impl CoverageShares {
         self,
         reward_shares: DataTransferAndPocAllocatedRewardBuckets,
         epoch: &'_ Range<DateTime<Utc>>,
+        banned_radios: BannedRadios,
```
michaeldjeffrey, line 675: It strikes me now that `OracleBoostingStatus` should probably be a part of `RadioInfo` the same way `SPBoostedRewardEligibility` is.

bbalser: That is better 73a0c61

michaeldjeffrey (approved):

PR #864, opened 2024-09-12
hip 118 visualization data requirements

andymck:
Linked proto pr: https://github.com/helium/proto/pull/418

Adds:

-  ingestor support for hex and radio usage messages
-  filestore support for hex and radio usage messages
-  known carrier key verification to ingestor.  Avoids having to consume the usage msgs from mobile verifier when those msgs are not used by the verifier


macpie (approved):

ingest/src/server_mobile.rs
```diff
@@ -150,10 +169,25 @@ impl GrpcServer {
             .map_err(|_| Status::invalid_argument("invalid signature"))?;
         Ok((public_key, event))
     }
+
+    async fn verify_known_carrier_key(&self, public_key: PublicKey) -> VerifyResult<()> {
```
jeffgrunewald, line 173: This is a pretty substantial change for our ingestors as it's effectively performing a verification on the incoming data when we have traditionally done this kind of check entirely in the verifiers with the one exception of signature verification on the incoming request message.

After our experience with the GRPC request/response needed to submit disco mapping verified reports to the ETL this feels like we're setting ourselves up for a similar pain in that _every_ ingested report that needs this validation will require a GRPC request and response to the mobile config service at a minimum and if the carrier key (thankfully a small set at the moment) falls out of the cache it means a db query by the mobile config service in order service the request as well.

What's the drive here behind performing this check in the ingestor?

andymck: The reasoning behind this was to avoid having to push a message to the verifier just to have the verifier carry out this verification when the verifier itself does nothing with the said message...other than consume it, check the key and write it back out with another wrapper.  

The check does not have to be used on every message, I would consider it intent for only those message types which are generated by a non public service but where the message itself is required to hit a public S3 bucket ( which is the case here ).

file_store/src/usage_counts.rs
```diff
@@ -0,0 +1,230 @@
...
+};
+use serde::{Deserialize, Serialize};
+
+#[derive(Clone, Deserialize, Serialize, Debug, PartialEq)]
+pub struct HexUsageCountsReq {
+    pub hex: CellIndex,
+    pub helium_mobile_subscriber_avg_count: u64,
+    pub helium_mobile_disco_mapping_avg_count: u64,
+    pub offload_avg_count: u64,
+    pub tmo_cell_avg_count: u64,
```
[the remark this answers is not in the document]
andymck: The msg types outputted are those originally requested.  Will leave it to @madninja and @bbalser to consider your comment as i do no have any insight into the TMO t&cs

michaeldjeffrey (approved):

PR #866, opened 2024-09-20
Promotion Fund workspace

michaeldjeffrey:
**Wait What?** To keep the PRs from being absolute giant behemoths of reviewing, this PR is setting the stage for a re-shuffling of https://github.com/helium/oracles/pull/856. 

proto: https://github.com/helium/proto/pull/413

### New `promotion_fund` workspace 
This workspace is all about dealing with Service Provider Promotion Fund allocation.

HIP-114
https://github.com/helium/HIP/blob/main/0114-incentive-escrow-fund-for-subscriber-referrals.md

Service Provider Promotions are stored in CarrierV0 on Solana.
To keep the mobile-verifier from talking to a chain, this service will
periodically check Solana and compare Service Providers allocations to
what is stored in S3.

If the values have changed, a new file will be output to a bucket for
the mobile-verifier rewarder to read from.

NOTE: Allocation Values are stored in Bps (Basis Points)
https://www.investopedia.com/terms/b/basispoint.asp

### Supporting updates 

- Add ingesting promotions 
- Dumping reward files with promotions
- Add to reward indexer

### DIFF
--- a/promotion_fund/src/daemon.rs
+++ b/promotion_fund/src/daemon.rs
@@ -58,7 +58,7 @@ impl Daemon {
         settings: &Settings,
         file_sink: FileSinkClient<ServiceProviderPromotionFundV1>,
     ) -> anyhow::Result<Self> {
-        let s3_current = fetch_s3_bps(&settings.file_store_output).await?;
+        let s3_current = fetch_s3_bps(settings).await?;
         let solana_client = SolanaRpc::new(&settings.solana).context("making solana client")?;
         let check_timer = tokio::time::interval(settings.solana_check_interval);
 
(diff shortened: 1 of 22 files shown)


macpie (approved):

promotion_fund/src/daemon.rs
```diff
@@ -0,0 +1,168 @@
...
+        }
+    }
+}
+
+pub async fn fetch_s3_bps(settings: &file_store::Settings) -> anyhow::Result<S3Value> {
+    let file_store = FileStore::from_settings(settings).await?;
+    let mut results = HashMap::new();
+
+    let all = file_store
+        .list_all(FileType::ServiceProviderPromotionFund.to_str(), None, None)
```
bbalser, line 125: anyway to have a date in this statement, so that over time we are pulling a list of thousands of files to just get the last one.  Maybe we could write something to disk?  and if it is not available on disk, we fall ack to just get everything?

reward_index/migrations/11_add_mobile_promotion_reward_type.sql
```diff
@@ -0,0 +1 @@
+ALTER TYPE reward_type ADD VALUE 'mobile_promotion';
```
bbalser, line 1: I don't think we need a new rewad_type, I think this should just be `mobile_subscriber` or `mobile_gateway` depending on key of promotion reward.

bbalser (approved):

reward_index/migrations/11_add_mobile_promotion_reward_type.sql
michaeldjeffrey: This promotion type has been removed in favor of using the already existing `MobileSubscriber` and `MobileGateway` members.

promotion_fund/src/daemon.rs
michaeldjeffrey: A `lookback_start_after` setting was added. It defaults to the unix epoch, but can be updated to a value closer to a recent file when the s3 bucket starts to fill up.

PR #874, opened 2024-10-08
Change mapping rewards so verification mappers can only earn rewards …

bbalser:
…if they also qualified for disco mapping rewards

### DIFF
--- a/mobile_verifier/src/rewarder.rs
+++ b/mobile_verifier/src/rewarder.rs
@@ -527,7 +527,7 @@ pub async fn reward_mappers(
     let location_shares =
         subscriber_location::aggregate_location_shares(pool, reward_period).await?;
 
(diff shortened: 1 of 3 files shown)


mobile_verifier/src/rewarder.rs
```diff
@@ -527,9 +527,14 @@ pub async fn reward_mappers(
     let location_shares =
         subscriber_location::aggregate_location_shares(pool, reward_period).await?;
 
+    // Verification mappers can only earn rewards if the qualified for disco mapping
```
jeffgrunewald, line 530:
```suggestion
    // Verification mappers can only earn rewards if they qualified for disco mapping
```

jeffgrunewald (approved):

andymck (approved):

PR #877, opened 2024-10-17
Service Provider Promotions again

michaeldjeffrey:
Proto PR :: https://github.com/helium/proto/pull/424
Please scrutinize the names.

## Previously

- `incentive_escrow_fund_bps` was stored in Solana, and written to s3.
- Promotions were submitted daily, and could point at a Subscriber or Gateway.

## Now

- `incentive_escrow_fund_bps` and promotions are stored in the `metadata_db`.
- Promotions are pointed to by special NFTs.
- Share logic works the same
- PromotionDaemon has been removed

## Temporary `metadata_db` table creation
I am testing with a local table until incentive programs make it into the metadata_db proper.
Using the [Column mapping](https://github.com/helium/helium-program-library/blob/master/packages/account-postgres-sink-service/src/utils/defineIdlModels.ts#L9) and [Account Structure ](https://github.com/helium/helium-program-library/pull/714/files#diff-73ae32f0b7d51084b1b68e09ada8ec3c24ae73ffa83520dfaa2d5984a1151f87) I have the following table.

```sql
CREATE TABLE incentive_escrow_programs (
    address character varying(255) NOT NULL,
    carrier character varying(255) NULL,
    start_ts numeric NULL,
    stop_ts numeric NULL,
    shares integer NULL,
    bump_seed integer NULL,
    name character varying(255) NULL,
    refreshed_at timestamp with time zone NULL,
    created_at timestamp with time zone NOT NULL
  );
```


mobile_verifier/src/service_provider/promotions.rs
```diff
@@ -0,0 +1,51 @@
...
+
+pub async fn get_promotions(
+    client: &impl CarrierServiceVerifier<Error = ClientError>,
+) -> anyhow::Result<ServiceProviderPromotions> {
+    let promos = client.list_incentive_promotions().await?;
+    Ok(ServiceProviderPromotions(promos))
+}
+
+#[derive(Debug, Default, Clone)]
+pub struct ServiceProviderPromotions(Vec<proto::ServiceProviderPromotions>);
```
andymck, line 19: i kinda like this wrapper and also dont

andymck (approved):

mobile_verifier/src/service_provider/promotions.rs
michaeldjeffrey: Agreed. The plural works at both levels. I was hoping someone might help with a better name for the proto, and the wrapper name would become clearer.

PR #880, opened 2024-10-22
[Mobile Config] Add device_types filter to stream_gateways_info

kurotych:


mobile_config/src/gateway_service.rs
```diff
@@ -156,16 +156,21 @@ impl mobile_config::Gateway for GatewayService {
         let signer = verify_public_key(&request.signer)?;
         self.verify_request_signature(&signer, &request)?;
 
-        tracing::debug!("fetching all gateways' info");
+        tracing::debug!(
+            "fetching all gateways' info. Device types: {:?} ",
+            request.device_types
```
jeffgrunewald, line 161: should this debug statement come after the map below? it would debug to a vec of i32s before that conversion, no?

mobile_config/src/gateway_info.rs
```diff
@@ -73,6 +73,17 @@ pub enum DeviceType {
...
+impl DeviceType {
+    fn to_sql_argument(&self) -> String {
+        match self {
+            DeviceType::Cbrs => "\"cbrs\"".to_string(),
+            DeviceType::WifiIndoor => "\"wifiIndoor\"".to_string(),
+            DeviceType::WifiOutdoor => "\"wifiOutdoor\"".to_string(),
+            DeviceType::WifiDataOnly => "\"wifiDataOnly\"".to_string(),
+        }
+    }
+}
```
jeffgrunewald, line 85: should be able to make this more general and just `impl std::fmt::Display` for this, avoid the escape characters, and it will work just as well

mobile_config/src/gateway_service.rs
kurotych: Good catch! Fixed

mobile_config/src/gateway_info.rs
kurotych: I've made it more generic but still forced to use quotas to make it works

mobile_config/src/gateway_info.rs
```diff
@@ -153,11 +165,24 @@ pub(crate) mod db {
...
+        match device_types.is_empty() {
+            true => sqlx::query_as::<_, GatewayInfo>(GET_METADATA_SQL)
+                .fetch(db)
+                .filter_map(|metadata| async move { metadata.ok() })
+                .boxed(),
+            false => sqlx::query_as::<_, GatewayInfo>(&DEVICE_TYPES_METADATA_SQL)
+                .bind(
+                    device_types
+                        .iter()
+                        .map(|v| format!("\"{}\"", v))
```
andymck, line 179: I assume the escaping here is required ?  A note to explain why would be useful for future readers, especially given the table schema is managed outside of the oracles

andymck (approved):

mobile_config/src/gateway_info.rs
kurotych: The device_types field has a jsonb type but is being used as a string, which forces us adding of the quotes.
image
image


Column in "meta" database: 
image

(Added comment to code)

bbalser: not sure it is a better idea but you can query the field normally like this
`WHERE mhi.device_type #>> '{}' = 'wifiDataOnly'`

jeffgrunewald (approved):

PR #883, opened 2024-10-29
mobile config list programs fix

michaeldjeffrey:
- Add mobile-config-cli carrier service command
- Fix trying to verify the pubkey of the wrong field of a request

### DIFF
--- a/mobile_config_cli/src/client.rs
+++ b/mobile_config_cli/src/client.rs
@@ -171,8 +171,14 @@ impl CarrierClient {
             signature: vec![],
         };
         request.signature = request.sign(keypair)?;
-        let res = self.client.list_incentive_promotions(request).await?;
-        Ok(res.into_inner())
+        match self.client.list_incentive_promotions(request).await {
+            Ok(response) => {
+                let res = response.into_inner();
+                res.verify(&self.server_pubkey)?;
+                Ok(res)
+            }
+            Err(error) => Err(error)?,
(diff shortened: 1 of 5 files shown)


mobile_config_cli/src/client.rs
```diff
@@ -142,6 +152,36 @@ impl AuthClient {
...
+    pub async fn list_incentive_promotions(
+        &mut self,
+        keypair: &Keypair,
+    ) -> Result<CarrierIncentivePromotionListResV1> {
+        let timestamp = SystemTime::now().duration_since(UNIX_EPOCH)?.as_secs();
+        let mut request = CarrierIncentivePromotionListReqV1 {
+            timestamp,
+            signer: keypair.public_key().into(),
+            signature: vec![],
+        };
```
jeffgrunewald, line 172: is this list proprietary such that we need to be signature-gating a request?

bbalser (approved):

mobile_config_cli/src/client.rs
michaeldjeffrey: yes, only the mobile-verifier should be retrieving it.

madninja (approved):

mobile_config_cli/src/client.rs
jeffgrunewald: If that’s the case why did we add it to the cli and not just the client?

PR #884, opened 2024-10-30
fix promotion rewards by encoding timestamp as seconds

bbalser:


mobile_config/src/client/carrier_service_client.rs
```diff
@@ -71,7 +71,7 @@ impl CarrierServiceVerifier for CarrierServiceClient {
         &self,
     ) -> Result<Vec<ServiceProviderPromotions>, Self::Error> {
         let mut request = mobile_config::CarrierIncentivePromotionListReqV1 {
-            timestamp: Utc::now().encode_timestamp_millis(),
+            timestamp: Utc::now().encode_timestamp(),
```
jeffgrunewald, line 74: If we’re generally dealing in timestamps in the millisecond precision range should we adjust the consuming code for this proto request struct to expect millis rather than confirming this scenario to deal in second-level precision?

bbalser: Looking through the mobile-config api, every request that has a timestamp has it in seconds.  So I think that having this be in seconds does keep consistency.
The timestamp in question is also being used in the sql query into the incentive_escrow_programs table which has `start_ts` and `end_ts` that are timestamps in seconds.  Today, the `carrier_service` just trusts that the timestamp was sent in seconds and uses it in the sql query, which is what caused this bug.  I think the best change we should make here is to have the server side convert the timestamp into a `DateTime<Utc>` and then convert it to the timestamp in seconds.  Thoughts?

jeffgrunewald (approved):

michaeldjeffrey (approved):

PR #887, opened 2024-11-01
Fix Service Provider rewards math

michaeldjeffrey:
**TLDR;**
Mixing unit is bad. mmmkay

Previously, for calculating which percentage of rewards a Service provider should receive, the calculation was suing this number as it's base.
`let used_allocation = total_sp_allocation.max(all_transfer);`

`total_sp_allocation` was the maximum number of Bones that could be rewarded in an epoch. `all_transfer` was all the DC transferred in the epoch. Because they were 2 different units, and Bones is at a scale far surpassing DC, the number would always be `total_sp_allocation`.

This is not a problem, until the amount of DC transferred when converted to Bones exceeds the amount allocated.

That leads to a situation where we would assume there is more rewards available than exist. Which in this case lead to matching rewards when there were non unallocated.

---

Now, we convert DC to bones as soon as possible. Since we're always working in the same units, the percentages should hold true.

### DIFF
--- a/mobile_verifier/src/reward_shares.rs
+++ b/mobile_verifier/src/reward_shares.rs
@@ -2275,7 +2275,7 @@ mod test {
             // force the service provider to have spend more DC than total rewardable
             ServiceProviderDCSessions::from([(sp1, total_rewards_value_in_dc * dec!(2.0))]),
             ServiceProviderPromotions::default(),
-            total_rewards_value_in_dc,
+            total_sp_rewards_in_bones,
             mobile_bone_price,
             epoch.clone(),
         );
--- a/mobile_verifier/src/service_provider/reward.rs
+++ b/mobile_verifier/src/service_provider/reward.rs
@@ -422,10 +422,10 @@ mod tests {
         let total_rewards = dec!(1000);
 
         let sp_infos = ServiceProviderRewardInfos::new(
-            ServiceProviderDCSessions::from([(0, total_rewards)]),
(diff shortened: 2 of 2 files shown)


mobile_verifier/src/reward_shares.rs
```diff
@@ -2275,7 +2275,7 @@ mod test {
             // force the service provider to have spend more DC than total rewardable
             ServiceProviderDCSessions::from([(sp1, total_rewards_value_in_dc * dec!(2.0))]),
             ServiceProviderPromotions::default(),
-            total_rewards_value_in_dc,
```
jeffgrunewald, line 2278: this variable doesn't seem to be used anymore but we're still calculating it above?

mobile_verifier/src/service_provider/reward.rs
```diff
@@ -74,28 +56,35 @@ impl ServiceProviderRewardInfos {
     pub fn new(
         dc_sessions: ServiceProviderDCSessions,
         promotions: ServiceProviderPromotions,
-        total_sp_allocation: Decimal,
-        mobile_bone_price: Decimal,
+        total_sp_allocation: Decimal, // Bones
+        mobile_bone_price: Decimal,   // Price in Bones
```
jeffgrunewald, line 60: i didn't think fmt would let you do a line-ending comment like this but great news if it does

bbalser (approved):

jeffgrunewald (approved):

PR #891, opened 2024-11-05
Clear orphan coverage objects

macpie:
If we run, it will delete:
- Last month = 45k
- Last week = 75K

Total current records 120k

Doing last period and keeping the last 10 valid should delete about 62k coverage objects.

### DIFF
--- a/mobile_verifier/migrations/38_coverage_objects_cascade_delete.sql
+++ b/mobile_verifier/migrations/38_coverage_objects_cascade_delete.sql
@@ -1,4 +1,7 @@
 ALTER TABLE
+    hexes DROP CONSTRAINT IF EXISTS hexes_uuid_fkey;
+
+ALTER TABLE
     hexes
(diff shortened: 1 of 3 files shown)


bbalser:
are you proposing that this function is run once a day as part of the cleanup process after rewards have been output?

macpie:
> are you proposing that this function is run once a day as part of the cleanup process after rewards have been output?

I was looking at it more before making a decision but I think yes that would be the goal. We just need to decide if we want to remove month old or week old.

macpie:
@thehardbits Could you confirm that any heartbeat that comes in should always have the latest coverage object uuid?

thehardbits:
No, unfortunately that's not the case. Sometimes time radio misses the last CO and continues to send old.  But radio should not fallback to previously used CO in case it sends new one in HB messages.

bbalser (approved):

PR #893, opened 2024-11-11
Use link to sqlx repo instead of fork

kurotych:
- Use the official SQLx repository link instead of the fork. It makes future updates a little bit easier and more transparent and we don't have to host fork in our repo. 
- Remove the unused `helium-crypto` dependency from `boost_manager`.

Updating `SQLx` further is currently impossible due to package conflicts, mainly because we are using an outdated Solana library.
Upgrading solana-client, solana-sdk, and solana-program to version ^2 could theoretically resolve these conflicts.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1788,7 +1788,6 @@ dependencies = [
  "file-store",
  "futures",
  "futures-util",
- "helium-crypto",
  "helium-proto",
  "http 0.2.11",
  "http-serde",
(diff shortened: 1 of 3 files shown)


bbalser (approved):

jeffgrunewald (approved):
Wow, did we really fork this and not make any changes instead of just pinning the original repo? Wild

kurotych:
> Wow, did we really fork this and not make any changes instead of just pinning the original repo? Wild

It looks like. I think it was for the time we were waiting the changes in https://github.com/launchbadge/sqlx/commit/42dd78fe931df651eac411316ed3eab87c2f79b2 to be accepted into main branch.

PR #894, opened 2024-11-12
[Mobile Config] Add request/response fields to info_stream

kurotych:
- Add `refreshed_at`, `created_at` to `gateway_info`
- Add `min_refreshed_at`  to `gateway_info_req_v1`

related PR: https://github.com/helium/proto/pull/429


mobile_config/src/gateway_service.rs
```diff
@@ -163,14 +163,18 @@ impl mobile_config::Gateway for GatewayService {
         let (tx, rx) = tokio::sync::mpsc::channel(100);
 
         let device_types: Vec<DeviceType> = request.device_types().map(|v| v.into()).collect();
+        let min_refreshed_at = DateTime::<Utc>::from_timestamp(request.min_refreshed_at as i64, 0)
```
jeffgrunewald, line 166: are we relying on this always defaulting to 0 in the proto when omitted, which should always decode to 1970-1-1? we could also `.unwrap_or_else(|| DateTime::UNIX_EPOCH)` which would have the equivalent result but i can see the argument for not silently defaulting to 1970-1-1 if the caller explicitly supplies an i64 timestamp that happens to be invalid

mobile_config/src/gateway_info.rs
```diff
@@ -25,23 +28,47 @@ impl GatewayInfo {
     }
 }
 
+#[derive(thiserror::Error, Debug)]
+pub enum GatewayInfoProtoParseError {
+    #[error("Invalid location string: {0}")]
+    InvalidLocation(String),
```
jeffgrunewald, line 34:
```suggestion
    InvalidLocation(#[from] std::num::ParseIntError),
```
i would personally keep this context from the root error, which should include the unparseable str while still mapping it into a variant of the new unified error type you've got here and then we can drop the extra `map_err` below

mobile_config/src/gateway_info.rs
```diff
@@ -25,23 +28,47 @@ impl GatewayInfo {
...
+                    .map_err(|_| {
+                        GatewayInfoProtoParseError::InvalidLocation(metadata.location.clone())
+                    })?,
             )
         } else {
             None
         };
         let device_type = info.device_type().into();
+
+        let created_at = DateTime::<Utc>::from_timestamp(info.created_at as i64, 0).ok_or(
```
jeffgrunewald, line 58:
```suggestion
        let created_at = Utc.timestamp_opt(info.created_at as i64, 0).single().ok_or(
```

mobile_config/src/gateway_service.rs
kurotych: > are we relying on this always defaulting to 0 in the proto when omitted, which should always decode to 1970-1-1? 

yes we are. I've left comment in [proto file](https://github.com/helium/proto/pull/429/files#diff-0351a6685cffa248fdc4ecd2d7188bf6fff91f11f905ce55fbe1107cec434a15R84) about this. 

> we could also .unwrap_or_else(|| DateTime::UNIX_EPOCH) which would have the equivalent result but i can see the argument for not silently defaulting to 1970-1-1 if the caller explicitly supplies an i64 timestamp that happens to be invalid

I don't see how we can use it in this case. Since `min_refreshed_at` is not an optional field in terms of proto file. 
If 0 was passed this is not an error. So, as for me `unwrap_or_else` doesn't make sense in this case.

bbalser (approved):

jeffgrunewald (approved):

PR #897, opened 2024-11-18
[Mobile Config] Add deployment_info field to GatewayInfo

kurotych:


mobile_config/src/gateway_info.rs
```diff
@@ -2,15 +2,99 @@ use chrono::{DateTime, TimeZone, Utc};
...
+    fn from(v: CbrsRadioDeploymentInfoProto) -> Self {
+        Self {
+            radio_id: v.radio_id,
+            elevation: v.elevation,
+        }
+    }
+}
+
+#[derive(Clone, Debug, Deserialize)]
+pub struct RadioDeploymentInfo {
```
jeffgrunewald, line 66: should probably also prepend this with `CbrsRadioDeploymentInfo` to avoid confusion

jeffgrunewald (approved):

mobile_config/src/gateway_info.rs
```diff
@@ -2,15 +2,99 @@ use chrono::{DateTime, TimeZone, Utc};
...
+    /// CBSD_ID or radio
```
andymck, line 67: what is the radio option here, is it radio_id, radio_serial ?

andymck (approved):

mobile_config/src/gateway_info.rs
kurotych: Fixed

mobile_config/src/gateway_info.rs
kurotych: I've removed confusing part taken from https://github.com/helium/helium-program-library/blob/22a2d00d556bc328f6126c865781d8361610522f/programs/helium-entity-manager/src/state.rs#L259

kurotych: It can't be radio_serial (because of privacy). So I think it is cbsd_id only

PR #898, opened 2024-11-19
hip138 - mobile verifier 

andymck:
linked proto pr: https://github.com/helium/proto/pull/431

This implements the mobile verifier requirements of hip138 and issues rewards in HNT.  The epoch emissions pool is derived from on-chain data.

Given the on-chain epochs are 24hrs in duration beginning at midnight UTC, the mobile verifier will now also use the same epoch.  Previously the mobile rewarder's epoch was 24hours from 1am UTC.  

This change will necessitate a final epoch of 23 hours in which rewards are issued in MOBILE   which will occur on the epoch prior to the switchover to rewarding in HNT.  Thereafter the mobile rewarder will align its 24hr epoch from midnight.


file_store/src/error.rs
```diff
@@ -41,6 +41,8 @@ pub enum Error {
     //Not recommended for internal use!
     #[error("external error")]
     ExternalError(#[from] Box<dyn std::error::Error + Send + Sync>),
+    #[error("general error")]
```
michaeldjeffrey, line 44: This doesn't look to be used anywhere

file_store/src/error.rs
```diff
@@ -41,6 +41,8 @@ pub enum Error {
...
+    Anyhow(#[from] anyhow::Error),
```
michaeldjeffrey, line 45: This doesn't look to be used anywhere

madninja: good.. because that's terribel

mobile_config/src/client/mod.rs
```diff
@@ -29,6 +31,8 @@ pub enum ClientError {
     UnknownServiceProvider(String),
     #[error("Invalid GatewayInfo proto response {0}")]
     InvalidGatewayInfoProto(#[from] crate::gateway_info::GatewayInfoProtoParseError),
+    #[error("error {0}")]
+    AnyhowError(#[from] anyhow::Error),
```
michaeldjeffrey, line 35: This was added to catch errors returned from `file_store::traits::TimestampDecode` which returns a `file_store::Error`. We already provide a mapping to that above with `VerificationError`, though it's not used anywhere. 

Maybe we can rename that 
```rs
    #[error("file_store: {0}")]
    FileStore(#[from] file_store::Error),
```
Then we'll get a full trace of what went wrong. It'll be a bit more verbose, but it removes the option to rely on string errors.

mobile_config/src/sub_dao_epoch_reward_info.rs
```diff
@@ -0,0 +1,113 @@
...
+pub struct RawSubDaoEpochRewardInfo {
+    epoch: u64,
+    epoch_address: String,
+    sub_dao_address: String,
+    rewards_issued: u64,
+    delegation_rewards_issued: u64,
+    rewards_issued_at: DateTime<Utc>,
+}
+
+impl TryFrom<RawSubDaoEpochRewardInfo> for SubDaoEpochRewardInfoProto {
```
michaeldjeffrey, line 28: This cannot fail, can we convert it to a regular `From`?

mobile_config/src/lib.rs
```diff
@@ -94,3 +99,23 @@ impl std::fmt::Display for KeyRole {
...
+
+impl TryFrom<u64> for EpochPeriod {
+    type Error = anyhow::Error;
+
+    fn try_from(next_reward_epoch: u64) -> anyhow::Result<Self> {
+        let start_time = Utc
+            .timestamp_opt(0, 0)
+            .single()
+            .ok_or_else(|| anyhow!("Failed to get Unix epoch start time"))?
+            + Duration::days(next_reward_epoch as i64);
```
michaeldjeffrey, line 115: chrono provides a unix constant.
```suggestion
       let start_time = DateTime::<Utc>::UNIX_EPOCH + Duration::days(next_reward_epoch as i64);
```

Then this can turn into a regular `From` implementation.

mobile_config/src/sub_dao_epoch_reward_info.rs
```diff
@@ -0,0 +1,113 @@
+use crate::EpochPeriod;
+use chrono::{DateTime, Utc};
+use file_store::traits::{TimestampDecode, TimestampEncode};
+use helium_proto::services::sub_dao::SubDaoEpochRewardInfo as SubDaoEpochRewardInfoProto;
+use rust_decimal::prelude::*;
+use std::ops::Range;
+
+#[derive(Clone, Debug)]
+pub struct ResolvedSubDaoEpochRewardInfo {
+    pub epoch: u64,
```
michaeldjeffrey, line 10: Would it be beneficial to name this something like `epoch_day`? Almost everywhere else we use `epoch` it's in reference to a timestamp or Range. 

Or would it make sense to have `EpochPeiod` at this level?

mobile_verifier/src/lib.rs
```diff
@@ -21,8 +21,11 @@ pub mod telemetry;
 pub use settings::Settings;
 
 use async_trait::async_trait;
+use rust_decimal::Decimal;
 use std::error::Error;
 
+pub const MOBILE_SUB_DAO_ONCHAIN_ADDRESS: &str = "39Lw1RH6zt8AJvKn3BTxmUDofzduCM2J3kSaGDZ8L7Sk";
+
```
michaeldjeffrey, line 28: This seems like something that should live in helium-lib? @madninja

madninja: it already -is-.. just not hardcoded like this but iirc derived in dao.rs

mobile_verifier/src/rewarder.rs
```diff
@@ -457,15 +492,15 @@ async fn reward_poc(
         &boosted_hexes,
         &boosted_hex_eligibility,
         &poc_banned_radios,
-        reward_period,
+        &reward_info.epoch_period,
```
michaeldjeffrey, line 495: I guess this is as good a place as any to ask the question.

Here we're passing only the `epoch_period`, and in a handful of other places we're passing the full `ResolvedSubDaoEpochRewardInfo` when we only use the `epoch_period` from it.

Do we want to be consistent and pass the full struct? Or continue to only pass `epoch_period` in places that don't use more information from `reward_info`?

mobile_verifier/src/service_provider/reward.rs
```diff
@@ -57,15 +53,15 @@ impl ServiceProviderRewardInfos {
         dc_sessions: ServiceProviderDCSessions,
         promotions: ServiceProviderPromotions,
         total_sp_allocation: Decimal, // Bones
-        mobile_bone_price: Decimal,   // Price in Bones
-        reward_epoch: Range<DateTime<Utc>>,
+        hnt_bone_price: Decimal,      // Price in Bones
+        reward_info: ResolvedSubDaoEpochRewardInfo,
```
michaeldjeffrey, line 57: This is an example of somewhere that only uses `epoch_period`.

file_store/src/error.rs
andymck: addressed here: e8643f61bafa27c8c1a9680da8f625f068f01393

mobile_config/src/client/mod.rs
andymck: addressed here: e8643f61bafa27c8c1a9680da8f625f068f01393

mobile_config/src/lib.rs
andymck: nice, i had forgotten about that. 
addressed here: e8643f61bafa27c8c1a9680da8f625f068f01393

mobile_config/src/sub_dao_epoch_reward_info.rs
andymck: addressed here: e8643f61bafa27c8c1a9680da8f625f068f01393

mobile_config/src/sub_dao_epoch_reward_info.rs
andymck: addressed here: e8643f61bafa27c8c1a9680da8f625f068f01393

mobile_verifier/src/rewarder.rs
andymck: the approach I have taken is to pass the `ResolvedSubDaoEpochRewardInfo` struct to the top level or primary reward functions.  Thereafter call outs to sub functions use the existing function signature where possible.  There were a couple of places which didnt follow this patten and where the full struct was passed to sub functions unnecessarily, those are now fixed.

addressed here: e8643f61bafa27c8c1a9680da8f625f068f01393

mobile_verifier/src/service_provider/reward.rs
andymck: addressed here: e8643f61bafa27c8c1a9680da8f625f068f01393

mobile_config/src/sub_dao_service.rs
```diff
@@ -0,0 +1,95 @@
...
+
+#[tonic::async_trait]
+impl sub_dao::sub_dao_server::SubDao for SubDaoService {
+    async fn info(
+        &self,
+        request: Request<SubDaoEpochRewardInfoReqV1>,
+    ) -> GrpcResult<SubDaoEpochRewardInfoResV1> {
+        let request = request.into_inner();
+        telemetry::count_request("sub_dao_reward_info", "info");
+        custom_tracing::record("sub_dao", &request.sub_dao_address);
```
jeffgrunewald, line 64: does this need a custom metric? there's only one subdao in the context of the mobile config service, no? and the read context for this is always going to be in the metrics pulled from the mobile config service/host so this shouldn't ever conflict with the metrics from the iot config service?

mobile_config/src/lib.rs
```diff
@@ -94,3 +98,17 @@ impl std::fmt::Display for KeyRole {
...
+pub struct EpochInfo {
+    pub period: Range<DateTime<Utc>>,
+}
+
+impl From<u64> for EpochInfo {
+    fn from(next_reward_epoch: u64) -> Self {
+        let start_time = DateTime::<Utc>::UNIX_EPOCH + Duration::days(next_reward_epoch as i64);
+        let end_time = start_time + Duration::days(1);
+        EpochInfo {
+            period: start_time..end_time,
```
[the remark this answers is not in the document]
bbalser: `next_reward_epoch` is not a unix timestamp, it is the number of days since the unix epoch, `20115` as an example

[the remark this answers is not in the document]
jeffgrunewald: huh. yeah, i see that now but that seems wildly confusing/counter to the prevailing wisdom that unix time is always seconds or milliseconds since the unix epoch.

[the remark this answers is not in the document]
bbalser: that is how the epoch is stored on chain, number of days since 1970

iot_verifier/src/rewarder.rs
```diff
@@ -166,21 +166,24 @@ impl Rewarder {
...
             .write(
                 RewardManifest {
-                    start_timestamp: scheduler.reward_period.start.encode_timestamp(),
-                    end_timestamp: scheduler.reward_period.end.encode_timestamp(),
+                    start_timestamp: scheduler.schedule_period.start.encode_timestamp(),
+                    end_timestamp: scheduler.schedule_period.end.encode_timestamp(),
                     written_files,
                     reward_data: Some(IotRewardData(reward_data)),
+                    epoch: 0, // TODO: replace placeholder value
+                    price: 0, // TODO: replace placeholder value
```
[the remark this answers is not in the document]
bbalser: This is fixed in the IOT PR https://github.com/helium/oracles/pull/900/files#diff-ee1adf87153c54a2596f49162b2b4b0c5585ebca1bee4561dc2376f1b502b951

not sure I understand why this change is here, but mack's plan is to merge the 2 branches together and then merge to main

mobile_verifier/src/telemetry.rs
```diff
@@ -1,14 +1,15 @@
...
-
 const LAST_REWARDED_END_TIME: &str = "last_rewarded_end_time";
 const DATA_TRANSFER_REWARDS_SCALE: &str = "data_transfer_rewards_scale";
 
 pub async fn initialize(db: &Pool<Postgres>) -> anyhow::Result<()> {
-    last_rewarded_end_time(rewarder::last_rewarded_end_time(db).await?);
-
+    let next_reward_epoch = rewarder::next_reward_epoch(db).await?;
+    let epoch_period: EpochInfo = next_reward_epoch.into();
+    last_rewarded_end_time(epoch_period.period.start);
```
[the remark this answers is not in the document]
andymck: In the old world, the epoch was derived from two DB fields last_rewarded_end_time & next_rewarded_end_time. 
 These would be advanced by the reward period with each reward run.  As such it dint really support a gap between epochs without manual intervention.   In the new world, it is pretty much the same process with the exception that you define the epoch day.  I dont see any issue here

iot_verifier/src/rewarder.rs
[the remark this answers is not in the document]
andymck: epoch and price are two new fields added to the manifest.  The iot values  are required to be defaulted on the mobile branch in order to compile.  And as per brians comment, the real implementation is done on the iot branch.  That branch will be merged into the mobile branch

mobile_config/src/sub_dao_service.rs
andymck: correct, it wont conflict with the iot config service metric.

macpie (approved):

bbalser (approved):

mobile_config/src/lib.rs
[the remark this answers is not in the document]
andymck: just noticed the second comment re epoch_period, the epoch_period is a generic struct and impl to take an epoch integer and convert to a range.  It is used by the rewarder ( and tests ) at the point of fetching the next reward epoch from the db ( which is stored as per chain ie number of days since unix epoch ) and converting to a range in order to populate the scheduler.  We only create EpochRewardInfo after the scheduler has determined it is time to reward.  As such there is a distinct use case for the epoch_period outside of the EpochRewardInfo.  There is no unnecessary duplication that I can see, all functions convert from an epoch to a range use the epoch_period impl.

PR #902, opened 2024-11-25
Have mobile-config track when the data for a mobile-radio changes

bbalser:
Add ability for mobile-config to track when the data for a mobile radio changes.

- [x] Add tests


mobile_config/src/mobile_radio_tracker.rs
```diff
@@ -0,0 +1,337 @@
...
+use task_manager::ManagedTask;
+
+type EntityKey = Vec<u8>;
+
+#[derive(Debug, Clone, sqlx::FromRow)]
+struct MobileRadio {
+    entity_key: EntityKey,
+    refreshed_at: DateTime<Utc>,
+    location: Option<i64>,
+    is_full_hotspot: Option<i32>,
```
jeffgrunewald, line 15: this is really just a bool, right?

jeffgrunewald (commented):
one concern i have is that we're converting all these hotspot info fields to integers when we pull them in, presumably for easier hashing, but not clear to me if the resulting in Mobile Info is stored somewhere or if its _exclusively_ used for determining a hash to see when the hotspot changes. is there a followup change to actually wire this "hotspot tracked changes" into the query api to do a mobile info api query filtering based on hotspots updated since X? and if yes would those be pulling the metadata (or cached) versions of the data being hashed by integer representation then?

jeffgrunewald:
does it make any more sense to extend the existing struct that we convert metadata hotspot info into to include the additional fields added in the new `MobileRadio` type added in the tracker module and then conditionally cast fields into their easily hashable formats (i32/i64) in the impl for converting the info struct to a change-tracked radio and reduce some duplication and drift in the different postgres-conversion structs?

jeffgrunewald (approved):

mobile_config/src/mobile_radio_tracker.rs
```diff
@@ -0,0 +1,337 @@
...
+        	AND mhi.refreshed_at IS NOT NULL
+    "#,
+    )
+    .fetch(metadata)
+    .filter_map(|result| async move {
+        if let Err(err) = &result {
+            tracing::error!(?err, "error when reading radio metadata");
+        }
+        result.ok()
+    })
```
kurotych, line 234: I'd prefer not to hide such errors but rather return an error result and stop updating to avoid the possibility of finishing with an inconsistent result.

kurotych (approved):

mobile_config/src/mobile_radio_tracker.rs
madninja: This is indeed a code smell.. why is this masking the error?

jeffgrunewald: we do this in several places where we perform `get_all` operations by services/agents; typically the logic has been to allow the overall system to continue operating in a degraded state, reporting errors as they happen, rather than causing a single bad record in 10s of thousands to cause the entire operation like a cache refresh in the mobile-verifier or rewardable entity distribution to grind to a halt, usually in off hours

madninja: fair.. the downside is that without looking at the logs (which no client does) you'd end up operating on partial data?

PR #904, opened 2024-11-26
HIP-134 unique connections

michaeldjeffrey:
- [HIP-134](https://github.com/helium/HIP/blob/main/0134-reward-mobile-carrier-offload-hotspots.md)
- [proto PR](https://github.com/helium/proto/pull/433)

## Summary

If a wifi hotspot has >25 unique connections of a 7 day rolling period, it automatically gets a `1x` Oracle Boosting multiplier.

It is _are_ exempt from Service Provider Banning.

## Update Coverage Point Calculator

- Added `OracleBoostingStatus::Qualified` variant. 

This variant is constructed when a wifi radios has >25 unique connections over a 7 day rolling period as submitted by the CDR processing service to the ingestor.

## New Ingestor

- Added `UniqueConnectionsIngestor`.

Verifies  a `unique_connections_ingest_report_v1` carrier_key, writes out `verified_unique_connections_ingest_report_v1` and saves unique counts to mobile-verifiers db for next rewards epoch.

## New db table

- Added `unique_connections` table

Stores unique connection reports received from the CDR processing service. Counts represent unique connections for a wifi hotspots over a 7 day rolling period.

The rewarder clears this table after rewarding keeping a days extra worth of data for reprocessing.


file_store/src/file_info.rs
```diff
@@ -237,87 +239,13 @@ pub enum FileType {
...
-            Self::PromotionRewardIngestReport => PROMOTION_REWARD_INGEST_REPORT,
-            Self::VerifiedPromotionReward => VERIFIED_PROMOTION_REWARD,
-            Self::ServiceProviderPromotionFund => SERVICE_PROVIDER_PROMOTION_FUND,
-            Self::HexUsageStatsIngestReport => HEX_USAGE_STATS_INGEST_REPORT,
-            Self::RadioUsageStatsIngestReport => RADIO_USAGE_STATS_INGEST_REPORT,
-            Self::HexUsageStatsReq => HEX_USAGE_STATS_REQ,
-            Self::RadioUsageStatsReq => RADIO_USAGE_STATS_REQ,
-        };
-        f.write_str(s)
+        f.write_str(self.to_str())
```
andymck, line 248: finally! nice...

mobile_verifier/src/unique_connections/db.rs
```diff
@@ -0,0 +1,78 @@
...
+    txn: &mut Transaction<'_, Postgres>,
+    report: &UniqueConnectionsIngestReport,
+) -> Result<(), sqlx::Error> {
+    // TODO: on conflict?
+    sqlx::query(
+        r#"
+        INSERT INTO unique_connections 
+        (hotspot_pubkey, unique_connections, start_timestamp, end_timestamp, received_timestamp)
+        VALUES
+        ($1, $2, $3, $4, $5)
```
andymck, line 50: does this require an on conflict strategy ?

mobile_verifier/src/unique_connections/ingestor.rs
```diff
@@ -0,0 +1,181 @@
...
+
+            self.verified_unique_connections_sink
+                .write(
+                    verified_report_proto.into(),
+                    &[("report_status", verified_report_status.as_str_name())],
+                )
+                .await?;
+        }
+
+        txn.commit().await?;
```
andymck, line 160: without an on conflict on the DB save, there is a risk of inserting the same data multiple times should the same file end up being reprocessed

andymck (commented):
generally looks good but need another pass to fully digest

mobile_verifier/src/unique_connections/db.rs
macpie: The other side runs daily, so there should not be any conflicts I think but might still be interesting to add a do nothing. And also maybe some kind of index for the SELECT above?

macpie:
Got a clippy issue

mobile_verifier/src/unique_connections/db.rs
michaeldjeffrey: On conflict statement added. It only protects us from reprocessing the same file again.

The query for retrieving counts has a `DISTINCT ON` to make sure we only use the latest value provided for any radio.

macpie (approved):

file_store/src/unique_connections.rs
```diff
@@ -0,0 +1,108 @@
...
+    }
+}
+
+impl TryFrom<proto::UniqueConnectionsReqV1> for UniqueConnectionReq {
+    type Error = Error;
+
+    fn try_from(value: proto::UniqueConnectionsReqV1) -> Result<Self, Self::Error> {
+        Ok(Self {
+            pubkey: value.pubkey.into(),
+            start_timestamp: value.start_timestamp.to_timestamp_millis()?,
```
andymck, line 74: sanity check: are the source start and end timestamps serialized with milli precision ?

mobile_verifier/src/unique_connections/db.rs
```diff
@@ -0,0 +1,146 @@
...
+        #[sqlx(try_from = "i64")]
+        unique_connections: u64,
+    }
+
+    let rows = sqlx::query_as::<_, UniqueConnections>(
+        r#"
+            SELECT DISTINCT ON(hotspot_pubkey, received_timestamp)
+                hotspot_pubkey, unique_connections
+            FROM unique_connections
+            WHERE received_timestamp >= $1 AND received_timestamp <= $2
```
andymck, line 27: should this not be `received_timestamp >= $1 AND received_timestamp < $2` ? You will have entries that apply to consecutive periods otherwise

mobile_verifier/src/unique_connections/ingestor.rs
```diff
@@ -0,0 +1,181 @@
...
+
+    async fn process_unique_connections_file(
+        &self,
+        file_info_stream: FileInfoStream<UniqueConnectionsIngestReport>,
+    ) -> anyhow::Result<()> {
+        let mut txn = self.pool.begin().await?;
+
+        let mut stream = file_info_stream.into_stream(&mut txn).await?;
+        while let Some(unique_connections_report) = stream.next().await {
+            let verified_report_status = self
```
andymck, line 135: Noting it here, whether it should apply or not is up for debate...the known carrier key check can now be performed at the ingestor side and if invalid/unknown the UniqueConnectionsReport would be dropped to the floor at that point.  If that is used the verifier would simply act as a consumer of the ingest reports without a need to validate again and without a need to write out Verified versions of the ingest report.

Unless the verifier has to apply its own verifications ( other than the known carrier key check ) I am not convinced that the pattern implemented here and elsewhere provides any real additional value

mobile_verifier/src/reward_shares.rs
```diff
@@ -475,11 +482,14 @@ impl CoverageShares {
...
                 .collect();
 
-            let oracle_boosting_status = if banned_radios.contains(&pubkey, cbsd_id.as_deref()) {
-                OracleBoostingStatus::Banned
-            } else {
-                OracleBoostingStatus::Eligible
-            };
+            let oracle_boosting_status =
+                if unique_connections::is_qualified(unique_connections, &pubkey, &radio_type) {
+                    OracleBoostingStatus::Qualified
```
andymck, line 487: a qualified hotspot cannot be banned ?

mobile_verifier/src/unique_connections/ingestor.rs
bbalser: at the moment it is consistent will all other data that verifier reads.  The only data that is verified by the ingestor is never read by the verifier.  I think consistency here is important

mobile_verifier/src/unique_connections/ingestor.rs
```diff
@@ -0,0 +1,181 @@
...
+                .verify_unique_connection_report(&unique_connections_report.report)
+                .await;
+
+            if matches!(
+                verified_report_status,
+                VerifiedUniqueConnectionsIngestReportStatus::Valid
+            ) {
+                db::save(&mut txn, &unique_connections_report).await?;
```
bbalser, line 143: why not do a batch insert for better performance?

mobile_verifier/src/unique_connections/db.rs
bbalser: Order by should be `ORDER BY hotspot_key, received_timestamp DESC` so we also use the last one sent within the rewardable epoch bounds

file_store/src/unique_connections.rs
andymck: confirmed, radio thresholds currently sets these to non milli timestamps

mobile_verifier/src/unique_connections/ingestor.rs
andymck: the consistency argument is fair

file_store/src/unique_connections.rs
michaeldjeffrey: https://github.com/helium/proto/pull/433/files#diff-41a424c2a5f8e8fc2a265baae2b8a8f277ea1d7d21a1a4bb73a2ab391c971e6eR400

timestamps are milliseconds, Pierre is updating radio-thresholds

mobile_verifier/src/unique_connections/db.rs
michaeldjeffrey: Order by fixed in https://github.com/helium/oracles/pull/904/commits/d22c7261792628ae4312c6f4157585174711e243

exclusive end range fixed in https://github.com/helium/oracles/pull/904/commits/d22c7261792628ae4312c6f4157585174711e243

mobile_verifier/src/reward_shares.rs
michaeldjeffrey: That is correct.

https://github.com/helium/HIP/blob/main/0134-reward-mobile-carrier-offload-hotspots.md#detailed-explanation

> Qualified Hotspots earn full PoC rewards regardless of whether they have a call detail record (CDR) or verification mapping event from a discovery mapper. For example, a qualified Hotspot would not have to meet the requirements of HIP 131, HIP 133, or similiar location based HIPs to earn full PoC. Applies to outdoor and indoor wifi hotspots.

mobile_verifier/src/unique_connections/ingestor.rs
michaeldjeffrey: batch inserting as of [d7a1127](https://github.com/helium/oracles/pull/904/commits/d7a1127d0c57f19794b7cf868a31bafeac75b33c)

andymck (approved):

PR #907, opened 2024-12-04
Have location tracking in mobile-config track

kurotych:


mobile_config/src/mobile_radio_tracker.rs
```diff
@@ -77,6 +77,8 @@ struct TrackedMobileRadio {
     hash: String,
     last_changed_at: DateTime<Utc>,
     last_checked_at: DateTime<Utc>,
+    asserted_location: Option<i64>,
+    asserted_location_change_at: Option<DateTime<Utc>>,
```
bbalser, line 81: can we make this `asserted_location_changed_at` to keep inline with other fields

mobile_config/src/mobile_radio_tracker.rs
```diff
@@ -86,17 +88,23 @@ impl TrackedMobileRadio {
             hash: radio.hash(),
             last_changed_at: radio.refreshed_at,
             last_checked_at: Utc::now(),
+            asserted_location: radio.location,
+            asserted_location_change_at: None,
```
bbalser, line 92: What is the thinking about making this None?  The first time we see a radio, this field will be null and how do you envision clients working with it?

mobile_config/src/mobile_radio_tracker.rs
```diff
@@ -86,17 +88,23 @@ impl TrackedMobileRadio {
...
     fn update_from_radio(mut self, radio: &MobileRadio) -> Self {
         let new_hash = radio.hash();
+        self.last_checked_at = Utc::now();
         if self.hash != new_hash {
             self.hash = new_hash;
             self.last_changed_at = radio.refreshed_at;
         }
+        if self.asserted_location != radio.location {
+            self.asserted_location = radio.location;
+            self.asserted_location_change_at = Some(self.last_checked_at);
```
bbalser, line 105: shouldn't we set this to `radio.refreshed_at` like we do with the `last_changed_at` field?

kurotych: it makes sense, fixed

mobile_config/src/mobile_radio_tracker.rs
kurotych: done

mobile_config/src/mobile_radio_tracker.rs
```diff
@@ -77,6 +77,8 @@ struct TrackedMobileRadio {
     hash: String,
     last_changed_at: DateTime<Utc>,
     last_checked_at: DateTime<Utc>,
+    asserted_location: Option<i64>,
```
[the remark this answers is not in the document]
kurotych: In that case I will have to implement sqlx::FromRow manually instead of using automatically generated.
Is it worth?

michaeldjeffrey (approved):

mobile_config/src/mobile_radio_tracker.rs
```diff
@@ -192,7 +201,9 @@ pub async fn get_tracked_radios(
             entity_key,
             hash,
             last_changed_at,
-            last_checked_at
+            last_checked_at,
+            asserted_location::bigint,
```
michaeldjeffrey, line 205: Can the migration add this column as a `BIGINT` to begin with?

kurotych: Done

PR #909, opened 2024-12-06
Feat/675 on chain ouis devaddrs

bryzettler:
After https://github.com/helium/helium-program-library/pull/745 is merged/deployed and a backfill is ran, creation of orgs and devaddrs will have been moved to solana instructions/structs.

• `payer` is renamed to `escrow_key` and is just a string
• `helium_netids` logic has been copied over to the cli where it will be utilized in constructing instructions
• Test migrations stubbed out with bare minimum fields. Indexer service will alter these with full struct fields
• Updated all read operations to read form new indexer tables

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1611,7 +1611,7 @@ checksum = "8c3c1a368f70d6cf7302d78f8f7093da241fb8e8807c05cc9e51a125895a6d5b"
 [[package]]
 name = "beacon"
 version = "0.1.0"
-source = "git+https://github.com/helium/proto?branch=master#b4c8c8f47dfff38a2ff1b7fe14e1b2a1beea651c"
+source = "git+https://github.com/helium/proto?branch=master#4085e00c6f4d82c3da798ae1bc97324bc9cada2e"
 dependencies = [
  "base64 0.21.7",
  "byteorder",
@@ -1621,7 +1621,7 @@ dependencies = [
  "rand_chacha 0.3.0",
  "rust_decimal",
  "serde",
- "sha2 0.10.8",
+ "sha2 0.9.9",
  "thiserror",
 ]
 
@@ -2125,6 +2125,7 @@ dependencies = [
 [[package]]
 name = "circuit-breaker"
 version = "0.1.0"
+source = "git+https://github.com/helium/helium-anchor-gen.git#3036b33793cfe54b20ab24761677493510d5bd50"
 dependencies = [
  "anchor-gen",
  "anchor-lang 0.29.0",
@@ -2766,6 +2767,7 @@ dependencies = [
 [[package]]
 name = "data-credits"
 version = "0.2.2"
+source = "git+https://github.com/helium/helium-anchor-gen.git#3036b33793cfe54b20ab24761677493510d5bd50"
 dependencies = [
  "anchor-gen",
  "anchor-lang 0.29.0",
@@ -3143,6 +3145,7 @@ checksum = "0206175f82b8d6bf6652ff7d71a1e27fd2e4efde587fd368662814d6ec1d9ce0"
 [[package]]
 name = "fanout"
 version = "0.1.0"
+source = "git+https://github.com/helium/helium-anchor-gen.git#3036b33793cfe54b20ab24761677493510d5bd50"
 dependencies = [
  "anchor-gen",
  "anchor-lang 0.29.0",
@@ -3718,6 +3721,7 @@ checksum = "2304e00983f87ffb38b55b444b5e3b60a884b5d30c0fca7d82fe33449bbe55ea"
 [[package]]
 name = "helium-anchor-gen"
 version = "0.1.0"
+source = "git+https://github.com/helium/helium-anchor-gen.git#3036b33793cfe54b20ab24761677493510d5bd50"
 dependencies = [
  "anchor-gen",
  "anchor-lang 0.29.0",
@@ -3727,7 +3731,6 @@ dependencies = [
  "helium-entity-manager",
  "helium-sub-daos",
  "hexboosting",
- "iot-routing-manager",
  "lazy-distributor",
  "lazy-transactions",
  "mobile-entity-manager",
@@ -3765,6 +3768,7 @@ dependencies = [
 [[package]]
 name = "helium-entity-manager"
 version = "0.2.11"
+source = "git+https://github.com/helium/helium-anchor-gen.git#3036b33793cfe54b20ab24761677493510d5bd50"
 dependencies = [
  "anchor-gen",
  "anchor-lang 0.29.0",
@@ -3773,6 +3777,7 @@ dependencies = [
 [[package]]
 name = "helium-lib"
 version = "0.0.0"
+source = "git+https://github.com/helium/helium-wallet-rs.git?branch=master#d527be0dc3db4641c0a658ed462b53e98864eb73"
 dependencies = [
  "anchor-client",
  "anchor-spl",
@@ -3812,6 +3817,7 @@ dependencies = [
 [[package]]
 name = "helium-proto"
 version = "0.1.0"
+source = "git+https://github.com/helium/proto?branch=master#4085e00c6f4d82c3da798ae1bc97324bc9cada2e"
 dependencies = [
  "bytes",
  "prost",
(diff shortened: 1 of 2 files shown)


iot_config/migrations/20241203190903_solana_net_ids.sql
```diff
@@ -0,0 +1,9 @@
+-- Migration here solely for testing purposes
+-- An instance of account-postgres-sink
+-- Will alter this table depending on an on-chain struct
+CREATE TABLE IF NOT EXISTS solana_net_ids (
+    address TEXT PRIMARY KEY,
+    id INTEGER NOT NULL,
+    authority TEXT NOT NULL,
+    current_addr_offset NUMERIC NOT NULL
```
macpie, line 8: What is this offset?

iot_config/migrations/20241203190903_solana_net_ids.sql
```diff
@@ -0,0 +1,9 @@
+-- Migration here solely for testing purposes
+-- An instance of account-postgres-sink
+-- Will alter this table depending on an on-chain struct
+CREATE TABLE IF NOT EXISTS solana_net_ids (
+    address TEXT PRIMARY KEY,
+    id INTEGER NOT NULL,
+    authority TEXT NOT NULL,
```
macpie, line 7: Who would be authority?

iot_config/migrations/20241203193310_add_organization_locks.sql
```diff
@@ -0,0 +1,23 @@
+create table organization_locks (
+    organization varchar(255) primary key not null,
+    locked bool default false,
+
+    inserted_at timestamptz not null default now(),
+    updated_at timestamptz not null default now()
+);
+
+select trigger_updated_at('organization_locks');
```
macpie, line 9:
```suggestion
    organization VARCHAR(255) PRIMARY KEY NOT NULL,
    locked BOOL DEFAULT false ,
    inserted_at TIMESTAMPTZ NOT NULL DEFAULT now(),
    updated_at TIMESTAMPTZ NOT NULL DEFAULT now()
);

SELECT trigger_updated_at('organization_locks');
```

iot_config/migrations/20241203193310_add_organization_locks.sql
```diff
@@ -0,0 +1,23 @@
...
+DO $$
+BEGIN
+    IF EXISTS (
+        SELECT FROM information_schema.tables
+        WHERE table_name = 'solana_organizations'
+    ) THEN
+        insert into organization_locks (organization, locked)
+            select sol_org.address, org.locked
+            from solana_organizations sol_org
+            left join organizations org on sol_org.oui = org.oui;
```
macpie, line 20:
```suggestion
        INSERT INTO organization_locks (organization, locked)
            SELECT sol_org.address, org.locked
            FROM solana_organizations sol_org
            LEFT JOIN organizations org ON sol_org.oui = org.oui;
```

iot_config/migrations/20241203190910_solana_organizations.sql
```diff
@@ -0,0 +1,39 @@
...
+CREATE TABLE IF NOT EXISTS solana_organizations (
+    address TEXT PRIMARY KEY,
+    net_id TEXT NOT NULL,
+    authority TEXT NOT NULL,
+    oui BIGINT NOT NULL,
+    escrow_key TEXT NOT NULL,
+    approved BOOLEAN NOT NULL
+);
+
+CREATE OR REPLACE FUNCTION delete_routes_on_solana_organizations_delete() RETURNS trigger AS $$
```
macpie, line 13: I think this could be replaced by an `ON DELETE CASCADE`

iot_config/migrations/20241203190910_solana_organizations.sql
```diff
@@ -0,0 +1,39 @@
+-- Migration here solely for testing purposes
+-- An instance of account-postgres-sink
+-- Will alter this table depending on an on-chain struct
+CREATE TABLE IF NOT EXISTS solana_organizations (
+    address TEXT PRIMARY KEY,
+    net_id TEXT NOT NULL,
+    authority TEXT NOT NULL,
+    oui BIGINT NOT NULL,
+    escrow_key TEXT NOT NULL,
```
macpie, line 9: Is this the key where the funds are coming from? aka why not keeping it `payer_key`?

iot_packet_verifier/src/burner.rs
```diff
@@ -125,18 +125,18 @@ where
...
         let mut balance_lock = self.balances.lock().await;
-        let payer_account = balance_lock.get_mut(&payer).unwrap();
+        let payer_account = balance_lock.get_mut(&escrow_key).unwrap();
         // Reduce the pending burn amount and the payer's balance by the amount
         // we've burned.
         payer_account.burned = payer_account.burned.saturating_sub(amount);
         payer_account.balance = payer_account.balance.saturating_sub(amount);
 
-        metrics::counter!("burned", "payer" => payer.to_string()).increment(amount);
+        metrics::counter!("burned", "payer" => escrow_key.to_string()).increment(amount);
```
jeffgrunewald, line 139:
```suggestion
        metrics::counter!("burned", "payer" => %escrow_key).increment(amount);
```
this should already be a string now, no?

iot_config/migrations/20241203193310_add_organization_locks.sql
jeffgrunewald: for that matter, why not make the `organization TEXT PRIMARY KEY NOT NULL` to be in line with the rest of the organization records in the testing/postgres-sink table records?

iot_config/src/org_service.rs
```diff
@@ -207,250 +176,9 @@ impl iot_config::Org for OrgService {
...
-            org: Some(org.into()),
-            net_id: net_id.into(),
-            devaddr_constraints,
-            timestamp: Utc::now().encode_timestamp(),
-            signer: self.signing_key.public_key().into(),
-            signature: vec![],
-        };
         resp.signature = self.sign_response(&resp.encode_to_vec())?;
-
+        println!("Response: {:?}", resp);
```
jeffgrunewald, line 181: remove or replace with a log trace

iot_config/src/org_service.rs
```diff
@@ -53,7 +52,7 @@ impl OrgService {
         })
     }
 
-    fn verify_admin_request_signature<R>(
+    fn _verify_admin_request_signature<R>(
```
jeffgrunewald, line 55: seems like this is no longer needed?

iot_config/migrations/20241203190903_solana_net_ids.sql
bryzettler: Part of the backfill will read all existing devaddrs and set this onchain state to the largest `end_addr` of the existing net_id. Then any additional devaddrs just use this `current_addr_offset` as their `start_addr` and increment it accordingly.

iot_config/migrations/20241203190903_solana_net_ids.sql
bryzettler: We would be the authority due to backfilling them with a wallet. We can also create additional net_ids with any wallet as the authority.

iot_config/migrations/20241203193310_add_organization_locks.sql
bryzettler: We want organizations to initialize unlocked?

iot_config/src/org.rs
```diff
@@ -1,478 +1,167 @@
...
-            .collect();
-        let constraints = row
-            .get::<Vec<(i32, i32)>, &str>("constraints")
-            .into_iter()
-            .map(|(start, end)| {
-                Some(DevAddrConstraint {
-                    start_addr: start.into(),
-                    end_addr: end.into(),
+        let address_str: String = row.get("address");
+        let address = Pubkey::from_str(&address_str).map_err(|e| SqlxError::Decode(Box::new(e)))?;
```
jeffgrunewald, line 30: does pubkey support `parse()` to combine these lines? `row.get("address").parse().map_err(etc)`

iot_config/src/org.rs
```diff
@@ -1,478 +1,167 @@
...
-}
-
-async fn insert_helium_constraints(
-    oui: u64,
-    net_id: NetIdField,
-    devaddr_ranges: &[DevAddrConstraint],
-    db: impl sqlx::PgExecutor<'_>,
-) -> Result<(), sqlx::Error> {
-    let mut query_builder: sqlx::QueryBuilder<sqlx::Postgres> = sqlx::QueryBuilder::new(
+    let netid = sqlx::query_scalar::<_, i64>(
```
jeffgrunewald, line 111: should this be queried as a bigint now? the field for that table is defined in the new migration as an `integer` not a `bigint` and i don't see this anything changing the `netid.into()` to expect an i32 to cast the result into the `NetIdField` type unless I missed it

PR #911, opened 2024-12-10
RetryConfig and TimeoutConfig in file_store 

kurotych:
When a connection problem occurs, downloading a file from AWS hangs.
Add the ability to configure timeout and retry policies in the `file_store`.

### DIFF
--- a/file_store/src/file_store.rs
+++ b/file_store/src/file_store.rs
@@ -3,7 +3,7 @@ use crate::{
     settings::{self, Settings},
     BytesMutStream, Error, FileInfo, FileInfoStream, Result,
 };
-use aws_config::meta::region::RegionProviderChain;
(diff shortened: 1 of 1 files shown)


jeffgrunewald (approved):

michaeldjeffrey:
Can we get these added to `Settings` as well so calls to `FileStore::from_settings` can pick them up?

kurotych:
> Can we get these added to `Settings` as well so calls to `FileStore::from_settings` can pick them up?

I think we can but should we?
I can't just add `TimeoutConfig` and `RetryConfig` structs into `Settings` because the `Settings` type is directly mapped to the config file and I'm not sure how much and what exact config parameters from these two aws configs should I include to `Settings`.  Also, after I include them we should start to support these config parameters (with backward compatibilities). I think we shouldn't hurry in this case.

michaeldjeffrey (approved):

PR #912, opened 2024-12-11
Radios that earn 0 rewards should be removed from coverage map

macpie:
Radio with:
- 0 average location trust score
- failed speedtest average
- banned

Should not be included in the coverage map.

### DIFF
--- a/mobile_verifier/src/reward_shares.rs
+++ b/mobile_verifier/src/reward_shares.rs
@@ -1237,7 +1237,7 @@ mod test {
                 timestamp,
                 upload_speed: bytes_per_s(10),
                 download_speed: bytes_per_s(100),
-                latency: 50,
+                latency: 10,
                 serial: "".to_string(),
             },
         };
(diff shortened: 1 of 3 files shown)


mobile_verifier/src/reward_shares.rs
```diff
@@ -770,6 +772,36 @@ pub fn get_scheduled_tokens_for_oracles(duration: Duration) -> Decimal {
...
+    if oracle_boosting_status == OracleBoostingStatus::Banned {
+        return false;
+    }
+
+    let avg_speedtest = Speedtest::avg(speedtests);
+    if avg_speedtest.tier() == SpeedtestTier::Fail {
+        return false;
+    }
+
+    let avg_trust_score = average_trust_score(trust_scores);
```
michaeldjeffrey, line 789: I think you want to use `coverage_point_calculator::location::multiplier();`, you'll need to make it `pub`. 

It handles CBRS always being trusted regardless of having any LocationTrust. But it may be a good idea to handle empty lists for non-CBRS radios.

macpie: Fixed

michaeldjeffrey (approved):

andymck (approved):

bbalser (approved):

PR #913, opened 2024-12-12
Add dump for UniqueConnectionsReport & VerifiedUniqueConnectionsReport

macpie:


michaeldjeffrey (approved):

file_store/src/cli/dump.rs
```diff
@@ -392,6 +394,31 @@ impl Cmd {
                         "coverage": coverage.coverage,
                     }))?;
                 }
+                FileType::UniqueConnectionsReport => {
+                    let report = UniqueConnectionsIngestReportV1::decode(msg)?;
+                    let req = UniqueConnectionReq::try_from(report.report.unwrap())?;
+                    print_json(&json!({
```
andymck, line 400: nit, do you need to do a selective print ? you could just do `print_json(&req)?;`

bbalser (approved):

file_store/src/cli/dump.rs
macpie: I think we just want to not have the signature

PR #917, opened 2024-12-16
mobile verifier rewarder should validate there are unique connections…

bbalser:
… within rewardable period not next period

### DIFF
--- a/mobile_verifier/src/rewarder/db.rs
+++ b/mobile_verifier/src/rewarder/db.rs
@@ -51,7 +51,7 @@ pub async fn no_unique_connections(
     let count = sqlx::query_scalar::<_, i64>(
         "SELECT COUNT(*) from unique_connections WHERE received_timestamp >= $1",
     )
-    .bind(reward_period.end)
+    .bind(reward_period.start)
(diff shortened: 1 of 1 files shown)


mobile_verifier/src/rewarder/db.rs
```diff
@@ -51,7 +51,7 @@ pub async fn no_unique_connections(
     let count = sqlx::query_scalar::<_, i64>(
         "SELECT COUNT(*) from unique_connections WHERE received_timestamp >= $1",
     )
-    .bind(reward_period.end)
+    .bind(reward_period.start)
```
jeffgrunewald, line 54: this sort of implies to me that it should be explicitly querying the range _between_ `reward_period.start` and `reward_period.end`; if this was ever working before or appearing to work it's running after the next rewardable period has started such that we should exclude those rewards from the result set that happened after the previous one technically ended no?

bbalser: Added b3fdc42

andymck (approved):

jeffgrunewald (approved):

PR #918, opened 2024-12-16
Clarify different queries that can prevent rewarding

michaeldjeffrey:
Unique Connections is the outlying query in that it depends on the start _and_ end of a reward_epoch, while the others only care about the end of the reward_epoch.

The different being, unique connections come in once per day, while the others come in constanyl throughout the day.


bbalser (approved):

macpie (approved):

jeffgrunewald (commented):
👍

PR #921, opened 2024-12-18
[Mobile Config] Add updated_at to GatewayInfoV2 (response)

kurotych:
The updated_at field in the response is required to complete the workflow for fetching only the most recently changed hotspots.

This solution is unoptimized since updated_radios is fetched every time. The performance (for v2 endpoints) might be  impacted.

The `updated_radios` hashmap should be cached in future versions.

https://github.com/helium/proto/pull/436

### DIFF
--- a/mobile_config/src/gateway_info.rs
+++ b/mobile_config/src/gateway_info.rs
@@ -103,8 +103,10 @@ pub struct GatewayInfo {
     pub address: PublicKeyBinary,
     pub metadata: Option<GatewayMetadata>,
     pub device_type: DeviceType,
-    // None for V1
+    // None for GatewayInfoProto (V1)
+    pub updated_at: Option<DateTime<Utc>>,
     pub created_at: Option<DateTime<Utc>>,
+    pub refreshed_at: Option<DateTime<Utc>>,
 }
 
 impl GatewayInfo {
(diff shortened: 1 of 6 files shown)


mobile_config/src/gateway_info.rs
```diff
@@ -103,8 +103,10 @@ pub struct GatewayInfo {
     pub address: PublicKeyBinary,
     pub metadata: Option<GatewayMetadata>,
     pub device_type: DeviceType,
-    // None for V1
+    // None for GatewayInfoProto (V1)
+    pub updated_at: Option<DateTime<Utc>>,
     pub created_at: Option<DateTime<Utc>>,
+    pub refreshed_at: Option<DateTime<Utc>>,
```
dcyoung, line 109: can you elaborate on the difference between `updated_at` and `refreshed_at` -- perhaps consider adding as comments as well.

dcyoung: That's where i got confused - the proto only includes `updated_at` but not `refreshed_at`
https://github.com/helium/proto/blob/b5564eb32d1fb11d7a0570a1c2311b9ed39d297e/src/service/mobile_config.proto#L81-L84

So `refreshed_at` is only related to bookkeeping/internals of this service?

kurotych: > These should be documented in the proto definition

There is no `refreshed_at` in proto definition it is internal variable. 
`updated_at` is documented in proto

> So `refreshed_at` is only related to bookkeeping/internals of this service?

Yes `refreshed_at` is internal thing. I intentionally left only `updated_at` to not confuse clients.
Also I don't see any purposes for clients to have `refreshed_at` field 

Added comments co code.

jeffgrunewald (approved):

dcyoung (approved):

bbalser (approved):

PR #923, opened 2024-12-19
[Mobile Config] Handle entity duplicates in Mobile radio tracker

kurotych:
During development testing, we encountered the following issue:

`ERROR mobile_config::mobile_radio_tracker: error in tracking changes to mobile radios err=error returned from database: ON CONFLICT DO UPDATE command cannot affect row a second time`

The root cause of this problem is the existence of duplicate entity keys in the key_to_assets table (on the dev server).

Solution: Handle `(entity_key, asset)` duplicates by fetching unique entries with the newer `refreshed_at`.

(Theoretically possible) Case where one entity_key maps to several different assets will still result in the error described above. Since I don't see a graceful recovery from such data corruption.

### DIFF
--- a/mobile_config/src/mobile_radio_tracker.rs
+++ b/mobile_config/src/mobile_radio_tracker.rs
@@ -208,32 +208,24 @@ pub async fn get_tracked_radios(
 fn get_all_mobile_radios(metadata: &Pool<Postgres>) -> impl Stream<Item = MobileRadio> + '_ {
     sqlx::query_as::<_, MobileRadio>(
         r#"
-        WITH unique_entity_keys AS (
-            SELECT
-                kta.entity_key as entity_key,
-                kta.asset as asset,
-                MAX(mhi.refreshed_at) AS refreshed_at
-            FROM key_to_assets kta
-            INNER JOIN mobile_hotspot_infos mhi
-                ON kta.asset = mhi.asset
-            WHERE kta.entity_key IS NOT NULL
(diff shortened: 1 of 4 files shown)


mobile_config/src/mobile_radio_tracker.rs
```diff
@@ -208,8 +208,21 @@ async fn get_tracked_radios(
...
+                kta.entity_key as entity_key,
+                kta.asset as asset,
+                MAX(mhi.refreshed_at) AS refreshed_at
+            FROM key_to_assets kta
+            INNER JOIN mobile_hotspot_infos mhi
+                ON kta.asset = mhi.asset
+            WHERE kta.entity_key IS NOT NULL
+              AND mhi.refreshed_at IS NOT NULL
+            GROUP BY kta.entity_key, kta.asset
+        )
```
jeffgrunewald, line 222: can this query be simplified back into something more like the original with a `SELECT DISTINCT ON (entity_key, asset) ... ORDER BY refreshed_at DESC` to limit results to the most recent single entry for that given combo of identifiers?

kurotych: Thank you for the hint. Now it is much better! 

https://www.postgresql.org/docs/current/sql-select.html#SQL-DISTINCT

bbalser (approved):

jeffgrunewald (approved):

PR #928, opened 2025-01-08
Implement tracked radios cache

kurotych:
The cache updates each time mobile tracker runs


mobile_config/src/gateway_service.rs
```diff
@@ -241,18 +240,17 @@ impl mobile_config::Gateway for GatewayService {
 
         let (tx, rx) = tokio::sync::mpsc::channel(100);
 
+        let radios_cache = Arc::clone(&self.tracked_radios_cache);
```
jeffgrunewald, line 243: `self.tracked_radios_cache` is already an `Arc`, you should be able to just `self.tracked_radios_cache.clone()` no?

jeffgrunewald (commented):
unless there's some edge cases around working with the RwLock i believe the calls to clone the arc can be simplified to simply `.clone()` the the cache struct field value since the arc implements the lightweight reference-counted clone

mobile_config/src/gateway_service.rs
kurotych: `Arc::clone(...)` and `self.tracked_radios_cache.clone()` are actually the same under the hood.
When I use `Arc::clone` I just want to emphasize (clarity) in the code that it clones the reference (not the value).


Since it is not crucial for me, I've replaced as you ask

mobile_config/src/mobile_radio_tracker.rs
```diff
@@ -183,6 +203,30 @@ async fn identify_changes(
         .await
 }
 
+const GET_UPDATED_RADIOS: &str =
+    "SELECT entity_key, last_changed_at FROM mobile_radio_tracker WHERE last_changed_at >= $1";
+
+pub async fn get_updated_radios(
+    pool: &Pool<Postgres>,
+    min_updated_at: DateTime<Utc>,
+) -> anyhow::Result<HashMap<PublicKeyBinary, DateTime<Utc>>> {
```
michaeldjeffrey, line 212:
```suggestion
) -> anyhow::Result<TrackedRadiosMap> {
```

mobile_config/src/mobile_radio_tracker.rs
```diff
@@ -1,10 +1,15 @@
...
 use chrono::{DateTime, Utc};
 use futures::{Stream, StreamExt, TryFutureExt, TryStreamExt};
+use helium_crypto::PublicKeyBinary;
+use sqlx::Row;
 use sqlx::{Pool, Postgres, QueryBuilder};
+use std::str::FromStr;
 use task_manager::ManagedTask;
+use tokio::sync::RwLock;
 
+pub type TrackedRadiosMap = HashMap<PublicKeyBinary, DateTime<Utc>>;
```
michaeldjeffrey, line 12: I'm wondering if this would make sense as it's own type
```rs
#[derive(Clone)]
pub struct TrackedRadiosMap(Arc<RwLock<HashMap<PublicKeyBinary, DateTime<Utc>>>);
```

We would get Clone for free.
And we could hang methods like `get_updated_radios` off of it, and it would self update.
And other access methods could manage the read locking instead of the user.

Just thinking out loud.

michaeldjeffrey (approved):

mobile_config/src/mobile_radio_tracker.rs
kurotych: Done, thanks

mobile_config/src/mobile_radio_tracker.rs
kurotych: Is it want you mean?
https://github.com/helium/oracles/commit/48dcfd44a4d32d7673c4536a4ba8ac7f1d32c45d

andymck (approved):

jeffgrunewald (approved):

PR #930, opened 2025-01-14
Helium Lib attempt: 2

michaeldjeffrey:
This is a doozy!

The initial swing at using `helium-lib` in Oracles was intended only for the constructing of transactions.
It then moved into also using `helium-lib` for sending and finalizing transactions, as there were issues with trusting when Solana said a transaction was finalized.

Much of the work in this PR was a result of that attempt. It wasn't quite what was considered useful for `helium-lib`, but it did provide a baseline for shoring up burn transaction in Oracles.

### Use `helium-lib` for constructing transaction
- `hexboosting::start_boost` in `boost_manager`
- `dc::burn_delegated` in `iot_packet_verifier` and `mobile_packet_verifier`

### `iot_packet_verifier` updates
- Convert tests to use `sqlx` instead of mocks
- Use `impl TxnStore` for tracking transactions
- Use `solana` workspace provided `TestSolanaClientMap` for Solana mock instead of local mock

### `mobile_packet_verifier` updates
- Add tests
- Add `pending_txns.sql` migration for tracking pending txns in a similar style to `iot_packet_verifier`
- `confirm_pending_txns` on startup
- Use `impl TxnStore` for tracking transactions
- Use `solana` workspace provided `TestSolanaClientMap` for Solana mock instead of local mock

### `solana` updates
- `SolanaRpc` takes it's `SubDao` as an argument
- Provide `burn::test_client::TestSolanaClientMap` for testing
- Add `sender` module for breaking out how to track burn transactions
- Add `Transaction` wrapper for `(SolanaTransaction, BlockHeight)`
- move `send_with_retry!` macro into `start_boost` until it also uses a `TxnStore` for retrying.

### Thoughts for upcoming PRs
- `PendingTxn` could be consolidated between `iot_packet_verifier` and `mobile_packet_verifier` into `solana`, but they are currently not similar enough to feel comfortable doing that.
- Sending multiple burn transactions in parallel, as is the intent for `helium-lib` handling send/retry/finalize logic for users.


mobile_packet_verifier/src/pending_txns.rs
```diff
@@ -0,0 +1,202 @@
...
+    conn: &PgPool,
+    signature: &Signature,
+) -> Result<(), sqlx::Error> {
+    let mut txn = conn.begin().await?;
+    sqlx::query("DELETE FROM pending_txns WHERE signature = $1")
+        .bind(signature.to_string())
+        .execute(&mut *txn)
+        .await?;
+
+    sqlx::query(
```
bbalser, line 138: given how rare this should happen (hopefully), would it be better to do the delete returning the rows deleted and calling `pending_burns::save` for each row to avoid having to maintain two pending_burn insert queries that need to be the same?

bbalser (approved):

macpie (approved):

mobile_packet_verifier/src/pending_txns.rs
```diff
@@ -0,0 +1,182 @@
...
+    for session in transfer_sessions.iter() {
+        pending_burns::save_data_transfer_session(&mut txn, session).await?;
+    }
+
+    txn.commit().await?;
+
+    Ok(())
+}
+
+pub async fn remove_pending_txn_success(
```
jeffgrunewald, line 159: These two functions are pretty much identical; seems like they could be combined into a single one that switches on a `is_success: bool` or enum and then conditionally performs that transfer session `for` loop on the success case which seems to be the only real difference

solana/src/burn.rs
```diff
@@ -1,35 +1,18 @@
-use crate::{send_with_retry, GetSignature, SolanaRpcError};
-use anchor_client::RequestBuilder;
-use async_trait::async_trait;
-use chrono::{DateTime, Utc};
-use helium_anchor_gen::{
-    anchor_lang::{AccountDeserialize, ToAccountMetas},
-    data_credits::{self, accounts, instruction},
-    helium_sub_daos::{self, DaoV0, SubDaoV0},
+use crate::{
+    read_keypair_from_file, sender, GetSignature, Keypair, Pubkey, SolanaRpcError, SubDao,
```
jeffgrunewald, line 2: these seem to be re-exports from from the helium-lib crate, some of which seem to be re-exports from the solana_sdk crate? personally not a huge fan of re-exports (would rather know right away where to go to see the code if i need to know what it's doing) but if we're going to do that can we be consistent about it?

solana/src/lib.rs
```diff
@@ -1,33 +1,60 @@
...
 use solana_sdk::pubkey::ParsePubkeyError;
-use solana_sdk::signature::Signature;
-use solana_sdk::transaction::Transaction;
-use std::time::SystemTimeError;
+use std::{fs::File, io::Read, path::Path, time::SystemTimeError};
+
+pub use helium_lib::{
+    dao::SubDao,
+    error,
+    keypair::{Keypair, Pubkey, Signature},
```
jeffgrunewald, line 8: another instance where these seem to be inconsistent between re-exports of the solana_client or solana_sdk types and functions. i can see a good argument to abstract _all_ of the solana stuff behind `helium-lib` and re-exporting all the above stuff there, otherwise i'd use the original source types and only import stuff from helium-lib that's specific to the helium solana ecosystem.

jeffgrunewald (commented):
biggest concern is the mixing and matching some exports from solana crates with re-exports from helium-lib. my personal bias is to avoid re-exports to avoid obfuscating code i might want to trace later but i can see the argument behind encapsulating anything solana behind helium-lib.

that being said, i think being inconsistent is going to lead to confusion and have seen that cause issues with crate upgrades later if we go to upgrade `solana_sdk` as an example and suddenly it has the "same type" as an older version defined in `helium-lib` that it can't reconcile

Cargo.toml
```diff
@@ -128,10 +128,13 @@ sqlx = { git = "https://github.com/launchbadge/sqlx.git", rev = "42dd78fe931df65
...
-# 
+#
 # [patch.'https://github.com/helium/proto']
-# helium-proto = { path = "../../proto" }
-# beacon = { path = "../../proto" }
+# helium-proto = { path = "../proto" }
+# beacon = { path = "../proto/beacon" }
+
+# [patch.'https://github.com/helium/helium-wallet-rs.git']
+# helium-lib = { path = "../helium-wallet-rs/helium-lib" }
```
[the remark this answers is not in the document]
michaeldjeffrey: As commented, these commented out patches are for convenience when you need to do work in one of these libs and don't want to continually push changes to those repos.

Not necessary, but have saved me from needing to jump to the patch documentation about 100 times.

solana/src/burn.rs
michaeldjeffrey: I think ideally, all `solana_sdk` things should come from `helium-lib` and we can drop the solana dependency from this workspace. But that is still a work in progress.

madninja:
> biggest concern is the mixing and matching some exports from solana crates with re-exports from helium-lib. my personal bias is to avoid re-exports to avoid obfuscating code i might want to trace later but i can see the argument behind encapsulating anything solana behind helium-lib.
> 
> that being said, i think being inconsistent is going to lead to confusion and have seen that cause issues with crate upgrades later if we go to upgrade `solana_sdk` as an example and suddenly it has the "same type" as an older version defined in `helium-lib` that it can't reconcile

I concur -a lot- here.. there's a reason they're re-exported from helium-lib... to avoid version clashes that we've been dealing with already..

PR #938, opened 2025-02-05
Location Validation Timestamp Cache

michaeldjeffrey:
Heartbeats from radios come with a `location_validation_timestamp`, which is the last time a radio had it's location validated.

A `location_validation_timestamp` is valid for up to 24 hours from the first time it was received in a heartbeat.

Previously, we assumed that a heartbeat with a `location_validation_timestamp` had already passed this check. That has been found to not be the case. So "validated heartbeats" were being output with `location_validation_timestamp`'s far in the past.

Now, we are correctly verifying that a `location_validation_timestamp` is within 24 hours of the first received heartbeat that declared it.

### DIFF
--- a/mobile_verifier/src/heartbeats/last_location.rs
+++ b/mobile_verifier/src/heartbeats/last_location.rs
@@ -252,11 +252,11 @@ mod tests {
             cache.get(&hotspot_one, now).await?,
             "Invalid timestamp current"
         );
-        assert_eq!(
-            Some(test_last_location(now, limit_timestamp)),
-            cache.get(&hotspot_two, now).await?,
-            "Limit timestamp current"
-        );
+        let expect = Some(test_last_location(now, limit_timestamp));
+        let cached = cache.get(&hotspot_two, now).await?;
+        println!("expect: {:?}", expect);
+        println!("cached: {:?}", cached);
+        assert_eq!(expect, cached, "Limit timestamp current");
         assert_eq!(
             Some(test_last_location(now, good_timestamp)),
             cache.get(&hotspot_three, now).await?,
--- a/mobile_verifier/src/heartbeats/last_location.rs
+++ b/mobile_verifier/src/heartbeats/last_location.rs
@@ -258,7 +258,7 @@ mod tests {
             "Good timestamp current"
         );
 
-        // Moving an hour into the future should invalidate all timestamps
+        // Moving an 1 day into the future should invalidate all timestamps
         // regardless of what has already been cached.
         let future = now + Duration::days(1);
         assert_eq!(
--- a/mobile_verifier/src/heartbeats/last_location.rs
+++ b/mobile_verifier/src/heartbeats/last_location.rs
@@ -174,15 +174,20 @@ mod tests {
     use sqlx::PgPool;
     use uuid::Uuid;
 
+    // Make sure test timestamps and DB timestamps have the same granularity
+    fn nanos_trunc(ts: DateTime<Utc>) -> DateTime<Utc> {
+        ts.duration_trunc(Duration::nanoseconds(10)).unwrap()
+    }
+    fn hour_trunc(ts: DateTime<Utc>) -> DateTime<Utc> {
+        ts.duration_trunc(Duration::hours(1)).unwrap()
+    }
+
     async fn insert_heartbeat(
         pool: &PgPool,
         hotspot: &PublicKeyBinary,
         received_timestamp: DateTime<Utc>,
         validation_timestamp: DateTime<Utc>,
     ) -> anyhow::Result<()> {
-        let truncated = received_timestamp
-            .duration_trunc(Duration::hours(1))
-            .unwrap();
         sqlx::query(
             r#"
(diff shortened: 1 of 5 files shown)


mobile_verifier/src/heartbeats/last_location.rs
```diff
@@ -56,9 +80,37 @@ impl LocationCache {
...
+        &self,
+        hotspot: &PublicKeyBinary,
+        heartbeat_timestamp: DateTime<Utc>,
+    ) -> anyhow::Result<Option<LastLocation>> {
+        let location = match self.locations.get(hotspot).await {
+            Some(last_location) => {
+                // The value may still be cached according to the system clock
+                // but not valid based on the time of the heartbeat in question.
+                let last = *last_location;
+                last.filter(|l| l.still_valid(heartbeat_timestamp))
```
bbalser, line 98: What is the scenario that makes this check necessary?

mobile_verifier/src/heartbeats/last_location.rs
```diff
@@ -67,58 +119,197 @@ impl LocationCache {
...
-                    .unwrap_or_else(|| Duration::days(365))
-                    .to_std()?,
-            )
-            .await;
+
         Ok(last_location)
     }
 
-    pub async fn fetch_last_location(
+    async fn maybe_cache_last_location(
```
bbalser, line 136: why does it have `maybe` in the name when it always caches a value for the hotspot?

bbalser:
Test is failing due to different level of precision coming out of db than expected.

mobile_verifier/tests/integrations/last_location.rs
```diff
@@ -192,6 +192,10 @@ async fn heartbeat_does_not_use_last_good_location_when_more_than_12_hours(
         dec!(1.0)
     );
 
+    // NOTE: A new cache will only query backwards up to 12 hours.
```
bbalser, line 195: I don't understand this comment, why will a new cache only query backwards up to 12 hours?

mobile_verifier/src/heartbeats/last_location.rs
michaeldjeffrey: I went back and forth on this for a while. 
The short answer is testing.

This is the only way I could think of to ensure we're keeping the contract of 24 hours from a given heartbeat. 

Time is not mockable in `retainer::Cache`. Using different cache durations didn't seem like something I wanted to do. And writing a test that waited significant amounts of real time is a no go.

Given @kurotych's question about caching None values at all, I am wondering if we should fallback to the database in a different way.

mobile_verifier/tests/integrations/last_location.rs
michaeldjeffrey: This comment was a note for me to remember how to ask you the questions I had.

mobile_verifier/src/heartbeats/last_location.rs
```diff
@@ -67,58 +119,197 @@ impl LocationCache {
...
+        );
+        assert_eq!(
+            Some(test_last_location(now, good_timestamp)),
+            cache.get(&hotspot_three, now).await?,
+            "Good timestamp current"
+        );
+
+        // Moving an hour into the future should invalidate all timestamps
+        // regardless of what has already been cached.
+        let future = now + Duration::days(1);
```
andymck, line 264: comment suggests moving 1 hour into future but it moves forward 1 day

mobile_verifier/src/heartbeats/last_location.rs
```diff
@@ -67,58 +119,197 @@ impl LocationCache {
...
+        match location_with_expiration(last_location) {
+            Some((last, cache_duration)) => {
+                self.locations
+                    .insert(hotspot.clone(), Some(last), cache_duration)
+                    .await;
+            }
+            None => {
+                self.locations
+                    .insert(hotspot.clone(), None, Duration::days(365).to_std().unwrap())
+                    .await;
```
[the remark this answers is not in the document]
andymck: agree, i dont see the point in caching a None value here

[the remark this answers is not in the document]
bbalser: might not be necessary, but originally we did it to avoid hitting the db on every heartbeat without a cached location.

mobile_verifier/src/heartbeats/last_location.rs
michaeldjeffrey: The cache duration is derived from the value being cached. 

We could make the cache bust during a test, but that still requires waiting some amount of time during the test.

The more I think about it, the cache is a convenience, but the actual contract of the values being returned is that they are not more than 24 hours from the time of a given heartbeat.

The query achieves that, this filter line achieves that. 

We will also be rethinking how these values are applicable when we figure out how to process heartbeats in a non-CBRS world.

mobile_verifier/src/heartbeats/last_location.rs
michaeldjeffrey: good eye, thank you

bbalser (approved):

PR #940, opened 2025-02-07
Service Provider Overrrides

andymck:
This adds a fourth dataset to address the Service Provider Overrides requirement as detailed here:

https://github.com/helium/helium-release-proposals/blob/feat/initial-template/releases/20250301-core-devs.md


hex_assignments/src/assignment.rs
```diff
@@ -75,36 +76,40 @@ impl HexAssignments {
...
             urbanized,
+            service_provider_override,
         } = self;
 
         use Assignment::*;
-        match (footfall, landtype, urbanized) {
+        match (footfall, landtype, urbanized, service_provider_override) {
+            // service provider override hex
+            // Overrides other dataset assignments if set
+            (_, _, _, A) => dec!(1.00),
```
kurotych, line 86: Do we need to handle (_, _, _, B) somehow?

kurotych (approved):

hex_assignments/src/service_provider_override.rs
```diff
@@ -0,0 +1,37 @@
...
+}
+
+impl HexAssignment for ServiceProviderOverride {
+    fn assignment(&self, cell: hextree::Cell) -> anyhow::Result<Assignment> {
+        let Some(ref service_provider_override) = self.service_provider_override else {
+            anyhow::bail!("No service provider override hex data set has been loaded");
+        };
+        match service_provider_override.contains(cell) {
+            Ok(true) => Ok(Assignment::A),
+            Ok(false) => Ok(Assignment::C),
```
vihu, line 33: This can probably be just:

```
Ok(true) => Ok(Assignment::A),
_ => Ok(Assignment::C),
```

hex_assignments/src/assignment.rs
andymck: For service provider overrides a hex will only have one of two values ( A or C ).  A means the hex has an override, C means no override.
As such there is no need to handle B explicitly, if the value if not A then we ignore that assignment

mobile_verifier/migrations/40_service_provider_override-up.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE hexes ADD COLUMN IF NOT EXISTS service_provider_override oracle_assignment;
```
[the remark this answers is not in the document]
andymck: addressed here: e758b44846f410ab1961e41bac6f28bc608103ff

hex_assignments/src/service_provider_override.rs
andymck: addressed here: e758b44846f410ab1961e41bac6f28bc608103ff

michaeldjeffrey (approved):

macpie (approved):

PR #944, opened 2025-02-18
Remove CBRS related code

kurotych:
- Disable CBRS in `CoverageMapBuilder.insert_coverage_object` function
- Disable `submit_cell_heartbeat` grpc ednpoint
- Disable (cbrs) `heartbeat_report_sink`
- Remove cbrs from CoverageMap
- Fixed/Reworked related tests 
- etc...

`FIXME-K` -- Will be fixed in next PR since it is not CBRS releated
 
https://github.com/helium/HIP/blob/main/0139-phase-out-cbrs.md


mobile_verifier/src/radio_threshold.rs
```diff
@@ -376,13 +375,13 @@ pub async fn verified_radio_thresholds(
 ) -> Result<VerifiedRadioThresholds, sqlx::Error> {
     let mut rows = sqlx::query_as::<_, RadioThreshold>(
         "SELECT hotspot_pubkey, cbsd_id
```
jeffgrunewald, line 377: do we need to select the cbsd_id field any longer here?

mobile_verifier/src/radio_threshold.rs
```diff
@@ -391,14 +390,15 @@ pub async fn delete(
...
+    let cbsd_id: Option<String> = None;
     sqlx::query(
         r#"
             DELETE FROM radio_threshold
             WHERE hotspot_pubkey = $1 AND (cbsd_id is null or cbsd_id = $2)
         "#,
     )
     .bind(ingest_report.report.hotspot_pubkey.to_string())
-    .bind(ingest_report.report.cbsd_id.clone())
+    .bind(cbsd_id)
```
jeffgrunewald, line 401: should be able to just nix this binding altogether and change the query to _just_ be `cbsd_id is null` if that distinction is important going forward

mobile_verifier/src/reward_shares.rs
```diff
@@ -776,6 +754,9 @@ fn eligible_for_coverage_map(
     radio_type: RadioType,
     trust_scores: &[LocationTrust],
 ) -> bool {
+    if radio_type == RadioType::IndoorCbrs || radio_type == RadioType::OutdoorCbrs {
```
jeffgrunewald, line 757: are we maintaining these variants b/c of potentially processing old data or could they just be deleted? it seems like we're no longer ingesting any data that might generate these types of records unless we're maintaining some degree of backward compatibility for existing records

jeffgrunewald (commented):
(hopefully already planned for a separate PR, as this is admittedly fairly large) it seems like it would be possible and nice to perform some additional cleanup on naming schemes and more complex wrapper enums/structs that are no longer necessary if we don't need to maintain a distinction between CBRS and wifi hotspots

mobile_verifier/src/reward_shares.rs
kurotych: Cbrs is removed from RadioType and HbType

mobile_verifier/src/radio_threshold.rs
kurotych: Done

ingest/src/server_mobile.rs
```diff
@@ -219,37 +213,13 @@ where
 
     async fn submit_cell_heartbeat(
         &self,
-        request: Request<CellHeartbeatReqV1>,
+        _request: Request<CellHeartbeatReqV1>,
```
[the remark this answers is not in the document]
kurotych: > just marking the rpc unimplemented!
The unimplemented! macro causes panic of service since it is a gRPC endpoint, which is not recommended.

BTW we are returning OK to keep backward compatibility with service(s) that use it.

file_store/src/coverage.rs
```diff
@@ -21,14 +21,12 @@ pub struct RadioHexSignalLevel {
 
 #[derive(Serialize, Deserialize, Debug, Clone)]
 pub enum KeyType {
-    CbsdId(String),
     HotspotKey(PublicKeyBinary),
```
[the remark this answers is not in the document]
kurotych: I'd save it (at this stage) since it matches with `CoverageObjectReqV1` (proto file)  which is also has `KeyType`.

kurotych:
> (hopefully already planned for a separate PR, as this is admittedly fairly large) it seems like it would be possible and nice to perform some additional cleanup on naming schemes and more complex wrapper enums/structs that are no longer necessary if we don't need to maintain a distinction between CBRS and wifi hotspots

At this (first) stage of removing CBRS code from the codebase we decided to not touch a database layer. 
After this PR is reviewed and deployed to prod. I think we can start another stage and carefully remove deprecated stuff from db layer.

ingest/src/server_mobile.rs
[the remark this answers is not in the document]
kurotych: > hmm. do we care about services that call this since only unsupported devices should be the ones submitting to this endpoint, right? 

I remember @bbalser was asked (somewhere in threads) to return OK instead of error in this endpoint.
@bbalser What do you think?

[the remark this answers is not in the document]
jeffgrunewald: sounds like we're worried about returning errors to our own service provider backend services so we would _not_ want to do this even if it's likely the most efficient course of action for the ingestor to take.

coverage_point_calculator/src/lib.rs
```diff
@@ -358,10 +339,6 @@ impl RadioType {
     pub fn is_wifi(&self) -> bool {
         matches!(self, Self::IndoorWifi | Self::OutdoorWifi)
     }
-
```
michaeldjeffrey, line 361: Should we also drop `is_wifi()` above. Could clean up some match statements where it's used.

coverage_point_calculator/src/location.rs
```diff
@@ -31,17 +31,11 @@ pub fn asserted_distance_to_trust_multiplier(
...
 }
 
-pub(crate) fn average_distance(radio_type: RadioType, trust_scores: &[LocationTrust]) -> Decimal {
-    // CBRS radios are always trusted because they have internal GPS
-    if radio_type.is_cbrs() {
-        return dec!(0);
-    }
-
+pub(crate) fn average_distance(trust_scores: &[LocationTrust]) -> Decimal {
+    // FIXME-K: if count = 0, division by zero happens
```
michaeldjeffrey, line 38: I have a vague memory of @bbalser telling me that the `valid_radios.sql` query will not return radios that have empty location trust arrays. 

But it was a long time ago, and I could be remembering something totally different.

That being said, when this is fixed, I'm not sure what a good value to return would be, as `0` is a best case scenario.

ingest/tests/mobile_ingest.rs
```diff
@@ -189,43 +182,21 @@ async fn submit_radio_usage_report() -> anyhow::Result<()> {
...
-
-    assert!(client.is_cell_heartbeat_rx_empty()?);
-
-    trigger.trigger();
-    Ok(())
+    // Cell heartbeat is disabled but should return OK
+    client
+        .submit_cell_heartbeat(&keypair, "cbsd-1")
+        .await
+        .unwrap();
```
michaeldjeffrey, line 194: Because we've broken things here before when trying to disable CBRS, can we add a check that the response has an `id` that's parseable to a timestamp?

mobile_packet_verifier/src/accumulate.rs
```diff
@@ -18,14 +18,12 @@ pub async fn accumulate_sessions(
...
-    cbrs_disable_time: DateTime<Utc>,
 ) -> anyhow::Result<()> {
     tokio::pin!(reports);
 
     while let Some(report) = reports.next().await {
-        if report.received_timestamp >= cbrs_disable_time
-            && report.report.data_transfer_usage.radio_access_technology
-                == DataTransferRadioAccessTechnology::Eutran
+        if report.report.data_transfer_usage.radio_access_technology
+            == DataTransferRadioAccessTechnology::Eutran
```
michaeldjeffrey, line 26: Could we put this in a function somewhere, like in the ingester. 
I've had the question before, and I'm sure I'll have it again after I've forgotten of what does `Eutran` mean. And it's not totally clear it means CBRS.

mobile_verifier/src/heartbeats/mod.rs
```diff
@@ -108,48 +85,28 @@ impl<'a> Encode<'a, Postgres> for KeyType<'a> {
...
         }
     }
 
-    pub fn is_cbrs(&self) -> bool {
-        matches!(self, Self::Cbrs(_))
-    }
-
     pub fn is_wifi(&self) -> bool {
         matches!(self, Self::Wifi(_))
     }
```
michaeldjeffrey, line 107: It appears this function is unused and can also be removed.

ingest/tests/mobile_ingest.rs
kurotych: Done

mobile_packet_verifier/src/accumulate.rs
kurotych: This is the only place we use it in such way and I think we don't want to add this to our codebease in near future.
Added comment for this line.

mobile_verifier/src/reward_shares.rs
```diff
@@ -1741,66 +1481,30 @@ mod test {
...
+            .get(&owner1)
+            .expect("Could not fetch owner1 rewards");
+
+        let owner_2_reward = *owner_rewards
+            .get(&owner2)
+            .expect("Could not fetch owner1 rewards");
+
+        let owner_3_reward = *owner_rewards
+            .get(&owner3)
+            .expect("Could not fetch owner1 rewards");
```
michaeldjeffrey, line 1494:
```suggestion
        let owner_1_reward = *owner_rewards
            .get(&owner1)
            .expect("Could not fetch owner1 rewards");

        let owner_2_reward = *owner_rewards
            .get(&owner2)
            .expect("Could not fetch owner2 rewards");

        let owner_3_reward = *owner_rewards
            .get(&owner3)
            .expect("Could not fetch owner3 rewards");
```

michaeldjeffrey (approved):

coverage_point_calculator/src/location.rs
bbalser: given the query is from a single table, and the values in question are not nullable, it is not possible to have rows return that from the table and have an empty array.

kurotych: > That being said, when this is fixed, I'm not sure what a good value to return would be, as 0 is a best case scenario.

If array is empty it doesn't return 0, it panics.

I think the best way is just make returning Result<Decimal> and return error if empty array.
It will be better than **panic** in code

bbalser (approved):

PR #949, opened 2025-02-18
Add skip_preflight setting for solana rpc

michaeldjeffrey:
In Oracle's we have historically had `skip_preflight: true` for all txns.
There have been updates to helium-lib that we cannot debug if preflight checks are skipped. 
This PR turns it into a setting, so it can toggle when we need better errors.

### DIFF
--- a/solana/src/burn.rs
+++ b/solana/src/burn.rs
@@ -5,6 +5,7 @@ use async_trait::async_trait;
 use helium_crypto::PublicKeyBinary;
 use helium_lib::{client, dc, token, TransactionOpts};
 use serde::Deserialize;
+use solana_client::rpc_config::RpcSendTransactionConfig;
 use solana_sdk::{commitment_config::CommitmentConfig, signature::Signature};
 use std::str::FromStr;
 use std::sync::Arc;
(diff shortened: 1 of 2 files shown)


andymck (commented):
so to confirm, the skip_preflight setting will be configured with a value of false and thus the preflight checks will be performed?

And isnt 'false' the default value on the sendTransaction API if not specified ?

bbalser (approved):

jeffgrunewald (approved):

michaeldjeffrey:
> so to confirm, the skip_preflight setting will be configured with a value of false and thus the preflight checks will be performed?
> 
> And isnt 'false' the default value on the sendTransaction API if not specified ?

Good question, I was unclear in my description (has been updated).
Oracles has hardcoded `skip_preflight: true` where txns are sent. 
This was generally safe as we sent those txns often and knew they were constructed properly.

The lastest update from helium-lib changes the Txn type to a `VersionedTransaction`. 
It seems to me that this change should not cause any problems, but something is going wrong, and performing preflight checks is a way that we can get errors from solana that are hopefully better than "The thing didn't happen". 

In short, Solana's default is perform preflight checks. Oracle's default was to skip. 
Now it's a setting.

PR #952, opened 2025-02-20
Move away from Ubuntu 20.04

macpie:
Looks like (except for upload) I got it to release.

https://github.com/macpie/oracles/actions/runs/13441498648


.github/workflows/CI.yml
```diff
@@ -6,22 +6,33 @@ on:
...
     branches: ["main"]
     tags: ["*"]
+  workflow_dispatch:  # Allows manual triggering
 
 env:
   CARGO_INCREMENTAL: 0 # this setting is automatically applied by rust-cache but documented here for explicitness
   CARGO_NET_RETRY: 10
   RUST_BACKTRACE: short
-  RUSTFLAGS: "-D warnings"
+  RUSTFLAGS: "-C debuginfo=0 -D warnings"
```
michaeldjeffrey, line 15: We have debug symbols turned on in release mode in `Cargo.toml`.

If we turn them off here, do we still get good stack traces when tests fail? 
Or if something happens to throw in production?

macpie: Good question, this is in CI so should not impact release but I should remove it anyway now it was an attempt at saving disk space.

bbalser (approved):

michaeldjeffrey (approved):

PR #954, opened 2025-02-22
Disable CBRS Take 2 - disable at ingest layer

bbalser:


ingest/src/server_mobile.rs
```diff
@@ -228,7 +239,7 @@ where
             .and_then(|public_key| self.verify_network(public_key))
             .and_then(|public_key| self.verify_signature(public_key, event))
             .map(|(_, event)| CellHeartbeatIngestReportV1 {
-                received_timestamp: timestamp,
+                received_timestamp: timestamp.timestamp_millis() as u64,
```
andymck, line 242: is this change to millisec precision gonna impact anything downstream?

bbalser: no, the first line of the function used to be `let timestamp: u64 = Utc::now().timestamp_millis() as u64;` and I just changed to keep it in `DateTime<Utc>` until written into the proto

andymck: I see that now

ingest/src/server_mobile.rs
```diff
@@ -218,7 +221,15 @@ where
         &self,
         request: Request<CellHeartbeatReqV1>,
     ) -> GrpcResult<CellHeartbeatRespV1> {
-        let timestamp: u64 = Utc::now().timestamp_millis() as u64;
+        let timestamp = Utc::now();
+
+        if timestamp >= self.cbrs_disable_time {
+            return Err(Status::failed_precondition(format!(
+                "CBRS radios are no longer supported as of {}",
```
madninja, line 228: I thought we’d  drop reports not error outfit avoid clients that haven’t updated yet?

ingest/tests/common/mod.rs
```diff
@@ -164,6 +178,32 @@ impl TestClient {
             hex_usage_stats_file_sink_rx,
             radio_usage_stats_file_sink_rx,
             unique_connections_file_sink_rx,
+            cell_heartbeat_rx,
+            data_transfer_rx,
+        }
+    }
+
+    pub async fn cell_heartbeat_recv(mut self) -> anyhow::Result<CellHeartbeatIngestReportV1> {
+        match timeout(Duration::from_secs(2), self.cell_heartbeat_rx.recv()).await {
```
madninja, line 187: What’s the timeout for?

bbalser: The message should be in the channel very quickly, so didn't want the test to take longer than 2 seconds to fail

ingest/src/server_mobile.rs
bbalser: We did say that when talking through having this code in the verifiers, but when we moved to the ingestor I figured we would just return an error, but I can change to return Ok and then drop on the floor if you prefer.

madninja: I'm ok erroring out but we'd have to make sure that upstream services can handle gettign an error back wihout either breaking or retrying like crazy

macpie (approved):

macpie (approved):

madninja (approved):

michaeldjeffrey (approved):

PR #958, opened 2025-02-26
Mobile Packet Verifier Transaction Recovery

michaeldjeffrey:
This PR is step 1 in getting back to using `helium-lib` for solana.
It does not currently use `helium-lib` for anything, yet.
But adds failure resilience for `mobie-packet-verifier`.

All the changes in `iot-packet-verifier` were aimed towards making the integration tests use more production code.
The only new change here, is we check if there are any in flight transactions before burning. 
There should only ever be 1 for this iteration of the code.

### Mobile Packet Verifier
- Can handle a shutdown with an inflight transaction.
- Pending Transactions are handled with 2 new db tables.
  - `pending_txns` and `pending_data_transfer_sessions`.
  - "burns" and their sessions are moved to the new tables when submitted.
    - If the transaction succeeds, the sessions are written to s3 and the pending txn is removed.
    - If it fails, the pending txn is deleted, and the sessions are moved back to the main data transfer table.
  - Burn values are derived from sessions in `data_transfer_sessions`.
- We always confirm any pending transactions before burning
- There should only ever be a single pending transaction

### IoT Packet Verifier
- Updated tests to use `#[sqlx::test]` to use production code.
- Removed unused Mock structs.
- Make sure we only have 1 in flight transaction at a time.

### Solana
- `TestSolanaClientMap` replaces `Arc<Mutex<HashMap<PublicKeyBinary, u64>>>` impls


iot_packet_verifier/src/burner.rs
```diff
@@ -94,9 +94,17 @@ where
     }
 
     pub async fn burn(&mut self) -> Result<(), BurnError> {
+        // There should only be a single pending txn at a time
+        let pending_txns = self.pending_tables.fetch_all_pending_txns().await?;
+        if !pending_txns.is_empty() {
+            tracing::info!(pending_txns = pending_txns.len(), "skipping burn");
```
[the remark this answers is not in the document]
michaeldjeffrey: I added this for parity with MPV. 
But yes, it would require human intervention and I don't think it should.

I'm going to make this branch return an error that will trigger the txn confirmation function, and add a test for it.

Good 👀

mobile_packet_verifier/src/burner.rs
```diff
@@ -36,7 +36,58 @@ impl<S> Burner<S>
...
+            } else {
+                pending_txns::remove_pending_txn_failure(pool, &signature).await?;
+            }
+        }
+
+        Ok(())
+    }
+
+    pub async fn burn(&self, pool: &PgPool) -> anyhow::Result<()> {
+        let pending_txns = pending_txns::pending_txn_count(pool).await?;
```
bbalser, line 85: Is this still needed here, since you are confirming right before burn is called?

michaeldjeffrey: It's technically not needed. I still kind of like it there for testing purposes and making the expectation clear.
I can remove it if you think it should be though.

michaeldjeffrey: Exposing only the `confirm_and_burn()` functionality is possible, but the changes in tests that would be required to make it clear what functionality is being tested, even when we're willing to use only side effects for confirmation, feel... not juice that isn't worth the squeeze.

In consideration that we're planning on refactoring these Burners in the near future, we've agreed this check is fine where it is, and time is better spent elsewhere at this time.

bbalser (approved):

PR #960, opened 2025-02-27
Add some logging when cbrs events after the cbrs disable time are dro…

bbalser:
…pped

### DIFF
--- a/ingest/src/server_mobile.rs
+++ b/ingest/src/server_mobile.rs
@@ -291,7 +291,7 @@ where
 
             tracing::info!(
                 ?pubkey,
-                "dropping DataTransferSessionReqV1 because cbrs disable time as passed"
+                "dropping DataTransferSessionReqV1 because cbrs disable time has passed"
             );
 
             return Ok(Response::new(DataTransferSessionRespV1 {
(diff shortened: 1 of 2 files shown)


ingest/src/server_mobile.rs
```diff
@@ -222,15 +222,19 @@ where
         request: Request<CellHeartbeatReqV1>,
     ) -> GrpcResult<CellHeartbeatRespV1> {
         let timestamp = Utc::now();
+        let event = request.into_inner();
 
         if timestamp >= self.cbrs_disable_time {
+            let pubkey = PublicKeyBinary::from(event.pub_key);
+            tracing::info!(
+                ?pubkey,
+                "dropping CellHeartbeatReqV1 because cbrs disable time as passed"
```
jeffgrunewald, line 231:
```suggestion
                "dropping CellHeartbeatReqV1 because cbrs disable time has passed"
```

jeffgrunewald: this one is confusing b/c it doesn't look like we're actually dropping it, conditionally or otherwise; we're just logging that message regardless of the source of the heartbeat?

macpie: Might want ton use:
```
custom_tracing::record(
    "pub_key",
    bs58::encode(pubkey).into_string(),
);
```

macpie (approved):

PR #963, opened 2025-03-03
Latest CI updates

macpie:
- Do not require `release_version` for manual trigger. The  release job will check for that or skip if left empty.
- Everything now runs in a container `ubuntu:20.04`, this makes runs more consistent.
- Tests are now run on a matrix generated prior based on `cargo metadata` so that new added tests or packages are not left out.
- Caching is now only for release as it seems to be faster to run test in a matrix individually without waisting 1min to pull cache. When releasing, a job to build the release will start right away and the upload job will wait for `[build-release, fmt, clippy, tests]`  to be completed.

Notes:
- I soon as we move away from `ubuntu:20.04` we should be able to remove the `Install Dependencies` step, saving us around 30s on every job.
- Cache is hardcoded to `ubuntu:20.04` for the moment.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -61,7 +61,10 @@ jobs:
           key: ubuntu:20.04-cargo-${{ hashFiles('**/Cargo.lock') }}
 
       - name: Build
-        run: cargo build
+        run: cargo build --all
+
+      - name: Build tests
+        run: cargo test --no-run
 
   fmt:
     runs-on: ubuntu-latest
@@ -186,7 +189,7 @@ jobs:
           repo-token: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Cache
-        uses: actions/cache@v4
+        uses: actions/cache/restore@v4
         with:
           path: |
             ~/.cargo/bin/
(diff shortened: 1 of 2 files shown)


.github/workflows/CI.yml
```diff
@@ -20,54 +20,24 @@ env:
...
       group: ${{ github.workflow }}-${{ github.ref }}-fmt
       cancel-in-progress: true
     steps:
+      - name: Install Dependencies
+        run: |
+          export DEBIAN_FRONTEND=noninteractive
+          apt-get update
+          apt-get install -y tzdata
+          ln -fs /usr/share/zoneinfo/UTC /etc/localtime
+          dpkg-reconfigure -f noninteractive tzdata
```
jeffgrunewald, line 38: i assume these two steps are required for the subsequent package installs to work or else we could combine our `apt-get` calls into a single one-time invocation?

.github/workflows/CI.yml
```diff
@@ -135,33 +143,34 @@ jobs:
...
       - name: Run unit and integration tests
         env:
-          DATABASE_URL: "postgres://postgres:postgres@localhost:5432/postgres"
+          DATABASE_URL: "postgres://postgres:postgres@postgres:5432/postgres"
         run: cargo test -p ${{ matrix.package }}  -- --include-ignored
-  
-  tests:
-    needs: build
+
+  build-release:
```
jeffgrunewald, line 154: what's the purpose of this new separate "build-release" stage compared to the one that actually packages the built binary into a deb package and pushes it up to the packagecloud repo?

macpie: so to push release you need tests, clippy and all to run so to avoid waiting too long we start building release on start so that by the time test are done we should only have a few min left on build and it can be pulled from cache to upload mostly

.github/workflows/CI.yml
macpie: You are right I can probably combine those

macpie: Fixed

bbalser (approved):

PR #968, opened 2025-03-11
Use docker for CI

macpie:


.github/workflows/Docker.yml
```diff
@@ -0,0 +1,59 @@
...
+        description: "Release Version (e.g., 1.0.0)"
+        required: false
+        type: string
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        package: [ingest, iot-config, iot-packet-verifier, iot-verifier, mobile-config, mobile-packet-verifier, mobile-verifier, reward-index]
```
michaeldjeffrey, line 20: Do we need?

- boost-manager
- poc-entropy
- price

.github/workflows/DockerCI.yml
```diff
@@ -0,0 +1,155 @@
...
+    runs-on: ubuntu-latest
+    concurrency: 
+      group: ${{ github.workflow }}-${{ github.ref }}-fmt
+      cancel-in-progress: true
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v4
+
+      - name: Restore Cached Base Image
+        uses: actions/cache@v3
```
michaeldjeffrey, line 46: Do we want to use `actions/cache@v4`?

.github/workflows/Docker.yml
macpie: I did everything with a "runner" but I don't know if we actually run everything, question for Brian

bbalser: We run those 3 services today, is that the question?

macpie: Yes

michaeldjeffrey (approved):

macpie:
Gonna use local branch instead of fork

PR #972, opened 2025-03-21
Feat/675 on chain ouis devaddrs

bryzettler:


iot_config/migrations/20241203190903_solana_net_ids.sql
```diff
@@ -0,0 +1,9 @@
+-- Migration here solely for testing purposes
```
michaeldjeffrey, line 1: Could we move all the tables for testing into a single migration file?
Or into the fixtures folder somehow so we don't have to wonder about dangling tables in prod that are unused?

iot_config/src/client/org_client.rs
```diff
@@ -6,15 +6,17 @@ use async_trait::async_trait;
...
+    OrgDisableReqV1, OrgEnableReqV1, OrgGetReqV2, OrgListReqV2, OrgResV2, OrgV2,
 };
 
 #[async_trait]
 pub trait Orgs: Send + Sync + 'static {
-    async fn get(&mut self, oui: u64) -> Result<OrgResV1, ClientError>;
-    async fn list(&mut self) -> Result<Vec<OrgV1>, ClientError>;
-    async fn enable(&mut self, oui: u64) -> Result<(), ClientError>;
-    async fn disable(&mut self, oui: u64) -> Result<(), ClientError>;
+    type Error: std::fmt::Debug + std::fmt::Display + Send + Sync + 'static;
```
michaeldjeffrey, line 14: I think this was incorrectly rebased while this PR was laying dormant.

solana/src/burn.rs
```diff
@@ -68,12 +65,8 @@ fn min_priority_fee() -> u64 {
 }
 
 impl Settings {
-    pub fn payers_to_monitor(&self) -> Result<Vec<PublicKeyBinary>, SolanaRpcError> {
-        self.payers_to_monitor
-            .iter()
-            .map(|payer| PublicKeyBinary::from_str(payer))
-            .collect::<Result<_, _>>()
-            .map_err(SolanaRpcError::from)
+    pub fn escrow_keys_to_monitor(&self) -> Result<Vec<String>, SolanaRpcError> {
```
michaeldjeffrey, line 68: This doesn't need to return a Result anymore.

iot_config/Cargo.toml
```diff
@@ -47,6 +48,8 @@ triggered = { workspace = true }
 task-manager = { path = "../task_manager" }
 humantime-serde = { workspace = true }
 custom-tracing = { path = "../custom_tracing", features = ["grpc"] }
+solana = { path = "../solana" }
+solana-sdk = { workspace = true }
```
michaeldjeffrey, line 52: We're trying to keep/move all solana deps into the `solana` workspace. Can we move/re-export whatever is needed from there?

iot_config/Cargo.toml
```diff
@@ -23,6 +23,7 @@ futures-util = { workspace = true }
 helium-crypto = { workspace = true, features = ["sqlx-postgres"] }
 helium-lib = { workspace = true }
 helium-proto = { workspace = true }
+helium-anchor-gen = { workspace = true }
```
michaeldjeffrey, line 26: `iot-config` doesn't appear to use `helium-anchor-gen`

iot_config/src/client/org_client.rs
```diff
@@ -6,15 +6,15 @@ use async_trait::async_trait;
...
 #[async_trait]
 pub trait Orgs: Send + Sync + 'static {
-    async fn get(&mut self, oui: u64) -> Result<OrgResV1, ClientError>;
-    async fn list(&mut self) -> Result<Vec<OrgV1>, ClientError>;
-    async fn enable(&mut self, oui: u64) -> Result<(), ClientError>;
-    async fn disable(&mut self, oui: u64) -> Result<(), ClientError>;
+    async fn get(&mut self, oui: u64) -> Result<OrgResV2, Self::Error>;
+    async fn list(&mut self) -> Result<Vec<OrgV2>, Self::Error>;
+    async fn enable(&mut self, oui: u64) -> Result<(), Self::Error>;
+    async fn disable(&mut self, oui: u64) -> Result<(), Self::Error>;
```
michaeldjeffrey, line 17:
```suggestion
    async fn get(&mut self, oui: u64) -> Result<OrgResV2, ClientError>;
    async fn list(&mut self) -> Result<Vec<OrgV2>, ClientError>;
    async fn enable(&mut self, oui: u64) -> Result<(), ClientError>;
    async fn disable(&mut self, oui: u64) -> Result<(), ClientError>;
```

iot_config/migrations/20241203190903_solana_net_ids.sql
bryzettler: These tables should never be dangling as they are the exact tables that a postgres_sink_instance will be creating/managing. They just need to exist at runtime for the tests also.

PR #973, opened 2025-03-24
Move config to only need CLI so that we don't need it for dump

macpie:


file_store/src/cli/bucket.rs
```diff
@@ -15,14 +15,17 @@ use helium_crypto::PublicKey;
...
     str::FromStr,
 };
 use tokio::fs;
 
 /// Commands on remote buckets
 #[derive(Debug, clap::Args)]
 pub struct Cmd {
+    /// Configuration file to use.
+    #[clap(short = 'c')]
+    config: path::PathBuf,
```
michaeldjeffrey, line 28: You already have `PathBuf` imported

michaeldjeffrey (approved):

bbalser (approved):

PR #976, opened 2025-03-26
Fix rewardable_bytes

macpie:
We should not be using `upload_bytes + download_bytes` as `rewardable_bytes` but the actual `rewardable_bytes` from `ValidDataTransferSession`


mobile_verifier/migrations/13_data_session.sql
```diff
@@ -4,6 +4,6 @@ create table hotspot_data_transfer_sessions (
     upload_bytes BIGINT NOT NULL,
     download_bytes BIGINT NOT NULL,
     num_dcs BIGINT NOT NULL,
-    received_timestamp TIMESTAMPTZ not null,
+    received_timestamp TIMESTAMPTZ NOT NULL,
```
bbalser, line 7: Do not modify existing migrations

mobile_verifier/migrations/13_data_session.sql
```diff
@@ -6,4 +6,4 @@ create table hotspot_data_transfer_sessions (
     num_dcs BIGINT NOT NULL,
     received_timestamp TIMESTAMPTZ not null,
     created_at TIMESTAMPTZ default now()
-);
+);
```
bbalser, line 9: This migration is still being modified

mobile_verifier/migrations/41_alter_data_session.sql
michaeldjeffrey, line 1: can we follow the convention of naming migration files after the change they're adding, and not just what they're changing?

mobile_verifier/migrations/13_data_session.sql
macpie: This is a git nightmare ...

bbalser (approved):

michaeldjeffrey (approved):

PR #977, opened 2025-03-27
feat: Improve reward calculation performance and reliability

seriousfuzzy:


michaeldjeffrey:
Hi seriousfuzzy,

Thanks for submitting this PR and for taking the time to engage with the project. 

Unfortunately, I am unable to review these changes because your branch does not compile.

Some things to note however.
- This project is not currently setup to work with both tokio _and_ rayon
- `#[instrument(skip(self), level = "info")]`
  - This will not work for functions that do not have a `self` argument
  - Be aware of when you're dumping large collections into logs
- Please consider not removing explanatory comments

seriousfuzzy:
@michaeldjeffrey 
Thank you for your reply.
Is there any opportunity to join the helium team as a full-time developer?

madninja:
> @michaeldjeffrey Thank you for your reply. Is there any opportunity to join the helium team as a full-time developer?

No. We are not hiring at this time. 

In the future, I would suggest that you submit working PRs instead of broken builds with no or questionable value add.

PR #981, opened 2025-03-30
Update mobile-verifier to process new subscriber mapping activity reports

bbalser:


mobile_verifier/src/subscriber_mapping_activity.rs
```diff
@@ -0,0 +1,271 @@
...
+where
+    AV: AuthorizationVerifier,
+    EV: EntityVerifier + Send + Sync + 'static,
+{
+    fn start_task(
+        self: Box<Self>,
+        shutdown: triggered::Listener,
+    ) -> futures::future::LocalBoxFuture<'static, anyhow::Result<()>> {
+        Box::pin(self.run(shutdown))
+        // let handle = tokio::spawn(async move { self.run(shutdown).await });
```
bbalser, line 220: @michaeldjeffrey I am hoping to actually spawn this, but when i do I get a weird lifetime issue that I haven't figured out yet, maybe you got some ideas?

bbalser: Addressed the lifetime issues here c68a9fb, not sure if it is the best way, but it seems to work

mobile_verifier/src/subscriber_mapping_activity.rs
```diff
@@ -191,43 +196,37 @@ where
...
-        Ok(None)
-    }
-}
-
 impl<AV, EV> ManagedTask for SubscriberMappingActivityDaemon<AV, EV>
 where
-    AV: AuthorizationVerifier + Clone,
-    EV: EntityVerifier + Clone + Send + Sync + 'static,
+    AV: AuthorizationVerifier,
+    EV: EntityVerifier + Send + Sync + 'static,
```
michaeldjeffrey, line 229: Do we want to move `Send + Sync + 'static` to the definition of `EntityVerifier` as it will be required anywhere we use it?

Cargo.toml
```diff
@@ -135,3 +135,7 @@ sqlx = { git = "https://github.com/launchbadge/sqlx.git", rev = "42dd78fe931df65
 # [patch.'https://github.com/helium/proto']
 # helium-proto = { path = "../../proto" }
 # beacon = { path = "../../proto" }
+
+[patch.'https://github.com/helium/proto']
+helium-proto = { git = "https://www.github.com/helium/proto.git", branch = "jg/disco-shares-v2" }
```
michaeldjeffrey, line 140: Might be nice to leave this as a comment when we move back to main branch. It probably won't be the last time we need to override a proto branch of a dep.

mobile_config/src/client/authorization_client.rs
```diff
@@ -11,7 +11,7 @@ use std::{sync::Arc, time::Duration};
 
 #[async_trait]
 pub trait AuthorizationVerifier: Send + Sync + 'static {
-    type Error;
+    type Error: std::error::Error + Send + Sync + 'static;
```
michaeldjeffrey, line 14: Since these are always `ClientError` can we remove the associated type?

mobile_verifier/src/subscriber_mapping_activity.rs
jeffgrunewald: probably more concise: `EntityVerifier: Send + Sync + 'static { trait functions }`

mobile_verifier/migrations/42_subscriber_mapping_activity.sql
```diff
@@ -0,0 +1,18 @@
+CREATE TABLE IF NOT EXISTS subscriber_mapping_activity (
+    subscriber_id BYTEA NOT NULL,
+    discovery_reward_shares BIGINT NOT NULL,
+    verification_reward_shares BIGINT NOT NULL,
+    received_timestamp TIMESTAMPTZ NOT NULL,
+    inserted_at TIMESTAMPTZ NOT NULL DEFAULT now(),
+    PRIMARY KEY (subscriber_id, received_timestamp)
```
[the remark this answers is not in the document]
bbalser: The `received_timestamp` is set by the mobile-ingestor and if the `mapper-verifier` sends 2 reports for the same subscriber we will ingest and save both reports to the db, but when rewarding we only use the most recent report per subscriber.

mobile_config/src/client/authorization_client.rs
bbalser: removed a223808

jeffgrunewald (approved):

mobile_verifier/migrations/42_subscriber_mapping_activity.sql
```diff
@@ -0,0 +1,18 @@
...
+);
+
+INSERT INTO subscriber_mapping_activity(subscriber_id, discovery_reward_shares, verification_reward_shares, received_timestamp, inserted_at)
+SELECT subscriber_id, 30, 0, received_timestamp, created_at AS inserted_at
+FROM subscriber_loc_verified;
+
+UPDATE subscriber_mapping_activity sma
+SET verification_reward_shares = svme.total_reward_points
+FROM subscriber_verified_mapping_event svme
```
michaeldjeffrey, line 16: Are we going to remove this table at some point in the future?

bbalser: yes, was going to remove it in a future PR

michaeldjeffrey (approved):

PR #982, opened 2025-03-31
Remove CBRS from db layer

kurotych:
Migration info: It has taken **1m48s** to run all migrations in this PR on the local DB setup.
UPD migration info: It takes 44s to run all migrations in this PR on the local labnet DB setup (04.22.25)

Mobile verifier:
- Remove `verify_legacy` function and `grandfathered_radio_threshold` TABLE.
- DELETE `cbsd_id` from `radio_threshold` table (migration) and `radio_threshold.rs` module 
- Remove selecting cbrs_heartbeats from valid_radios.sql
- Remove cell_type from struct HeartbeatReward
- Make distances_to_asserted not optional in HeartbeatReward
- DROP `old_hex_coverage` table
- Add migration `DELETE FROM seniority WHERE radio_type = 'cbrs'` which deletes around 500k rows (75%) in seniority table. 
- Add migration `DELETE FROM coverage_objects WHERE radio_type = 'cbrs';` which deletes around 35k rows (55%) in coverage_objects table.


mobile_verifier/src/radio_threshold.rs
```diff
@@ -293,43 +285,23 @@ where
...
-            cbsd_id,
             bytes_threshold,
             subscriber_threshold,
             threshold_timestamp,
             threshold_met,
             recv_timestamp)
-            VALUES ($1, $2, $3, $4, $5, true, $6)
-            ON CONFLICT (hotspot_pubkey, COALESCE(cbsd_id, ''))
+            VALUES ($1, $2, $3, $4, true, $5)
+            ON CONFLICT (hotspot_pubkey)
```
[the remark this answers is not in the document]
kurotych: > Verify that using only hotspot_pubkey as the unique constraint does not allow duplicate entries

In migration `43_radio_threshold_rework_unique_idx.sql` unique index reworked to only `hotspot_pubkey` field
 
```sql
...
CREATE UNIQUE INDEX radio_threshold_hotspot_pubkey_idx ON radio_threshold (hotspot_pubkey);
```

mobile_verifier/src/radio_threshold.rs
```diff
@@ -255,16 +255,8 @@ where
...
         let report_validity = self.do_report_verifications(report).await;
-        let final_validity = if is_legacy
-            && report_validity == RadioThresholdReportVerificationStatus::ThresholdReportStatusValid
-        {
-            RadioThresholdReportVerificationStatus::ThresholdReportStatusLegacyValid
-        } else {
-            report_validity
-        };
-        Ok(final_validity)
+        Ok(report_validity)
```
michaeldjeffrey, line 259: This function can no longer fail.

mobile_verifier/src/radio_threshold.rs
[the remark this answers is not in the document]
michaeldjeffrey: Do we know that this unique constraint holds in the current mainnet db?

mobile_verifier/src/radio_threshold.rs
```diff
@@ -374,8 +345,7 @@ pub async fn verified_radio_thresholds(
     reward_period: &Range<DateTime<Utc>>,
 ) -> Result<VerifiedRadioThresholds, sqlx::Error> {
     let mut rows = sqlx::query_as::<_, RadioThreshold>(
```
michaeldjeffrey, line 347: We're only grabbing a single field, this query can be simplified

```rs
let gateways = sqlx::query_scalar::<_, PublicKeyBinary>(
        "SELECT hotspot_pubkey
             FROM radio_threshold WHERE threshold_timestamp < $1 and cbsd_id IS NULL",
    )
    .bind(reward_period.end)
    .fetch_all(pool)
    .await?
    .into_iter()
    .collect::<HashSet<_>>();

    Ok(VerifiedRadioThresholds { gateways })
```

And we can get rid of `RadioThreshold`.

mobile_verifier/src/radio_threshold.rs
```diff
@@ -392,12 +362,349 @@ pub async fn delete(
...
+        );
+        assert_eq!(result.get::<i64, _>("bytes_threshold"), 1000);
+        assert_eq!(result.get::<i32, _>("subscriber_threshold"), 50);
+        assert!(result.get::<bool, _>("threshold_met"));
+
+        let stored_threshold_timestamp = result.get::<DateTime<Utc>, _>("threshold_timestamp");
+        assert_eq!(
+            stored_threshold_timestamp.timestamp_millis(),
+            now.timestamp_millis()
+        );
```
michaeldjeffrey, line 441: All these types support sqlx Encode and Decode. This could also be written as
```rs
        #[derive(Debug, FromRow, PartialEq)]
        struct TestRadioThreshold {
            hotspot_pubkey: PublicKeyBinary,
            bytes_threshold: i64,
            subscriber_threshold: i32,
            threshold_met: bool,
            threshold_timestamp: DateTime<Utc>,
        }

        let result = sqlx::query_as::<_, TestRadioThreshold>(
            r#"
            SELECT
                hotspot_pubkey,
                bytes_threshold,
                subscriber_threshold,
                threshold_timestamp,
                threshold_met
            FROM radio_threshold
            WHERE hotspot_pubkey = $1
            "#,
        )
        .bind(&hotspot_pubkey)
        .fetch_one(&pool)
        .await
        .unwrap();

        pub fn nanos_trunc(ts: DateTime<Utc>) -> DateTime<Utc> {
            use chrono::{Duration, DurationRound};
            ts.duration_trunc(Duration::nanoseconds(1000)).unwrap()
        }

        assert_eq!(
            result,
            TestRadioThreshold {
                hotspot_pubkey,
                bytes_threshold: 1000,
                subscriber_threshold: 50,
                threshold_met: true,
                threshold_timestamp: nanos_trunc(now)
            }
        );
```

However, `nanos_trunc` is needed because...
When retreiving a timestamp from DB, depending on the version of postgres
the timestamp may be truncated. When comparing datetimes, to ones generated
in a test with `Utc::now()`, you should truncate it.

michaeldjeffrey (approved):

mobile_verifier/src/radio_threshold.rs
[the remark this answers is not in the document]
kurotych: I've tested migrations on the one of the mobile verifier backups and it is worked.

But I will check it again before merge on the latest mainnet backup.

mobile_verifier/src/radio_threshold.rs
kurotych: As I see it can fail in case the pubkey is unknown (RadioThresholdReportVerificationStatus::ThresholdReportStatusInvalidCarrierKey) 
image

kurotych: Good catch, fixed

kurotych:
The comment I accidentally removed: 

@michaeldjeffrey commented on this pull request.

In [mobile_verifier/src/radio_threshold.rs](https://github.com/helium/oracles/pull/982#discussion_r2025380936):

> @@ -374,8 +345,7 @@ pub async fn verified_radio_thresholds(
     reward_period: &Range<DateTime<Utc>>,
 ) -> Result<VerifiedRadioThresholds, sqlx::Error> {
     let mut rows = sqlx::query_as::<_, RadioThreshold>(

We're only grabbing a single field, this query can be simplified

let gateways = sqlx::query_scalar::<_, PublicKeyBinary>(
        "SELECT hotspot_pubkey
             FROM radio_threshold WHERE threshold_timestamp < $1 and cbsd_id IS NULL",
    )
    .bind(reward_period.end)
    .fetch_all(pool)
    .await?
    .into_iter()
    .collect::<HashSet<_>>();

Ok(VerifiedRadioThresholds { gateways })

And we can get rid of RadioThreshold.

kurotych:
> The comment I accidentally deleted:
> 
> @michaeldjeffrey commented on this pull request.
> 
> In [mobile_verifier/src/radio_threshold.rs](https://github.com/helium/oracles/pull/982#discussion_r2025380936):
> 
> > @@ -374,8 +345,7 @@ pub async fn verified_radio_thresholds(
> > reward_period: &Range<DateTime>,
> > ) -> Result<VerifiedRadioThresholds, sqlx::Error> {
> > let mut rows = sqlx::query_as::<_, RadioThreshold>(
> 
> We're only grabbing a single field, this query can be simplified
> 
> let gateways = sqlx::query_scalar::<_, PublicKeyBinary>( "SELECT hotspot_pubkey FROM radio_threshold WHERE threshold_timestamp < $1 and cbsd_id IS NULL", ) .bind(reward_period.end) .fetch_all(pool) .await? .into_iter() .collect::<HashSet<_>>();
> 
> Ok(VerifiedRadioThresholds { gateways })
> 
> And we can get rid of RadioThreshold.

@michaeldjeffrey  done

mobile_verifier/src/radio_threshold.rs
kurotych: Done

mobile_verifier/migrations/42_drop_grandfathered_radio_threshold.sql
```diff
@@ -0,0 +1 @@
+DROP TABLE grandfathered_radio_threshold;
```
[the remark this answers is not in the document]
kurotych: I prefer to keep migrations as simple as possible.
Also I did this migrations in the way I think about the problem, so I wouldn't like to change them to save 1-2files space in repo

bbalser:
Looks good, need to renumber the migrations

kurotych:
> Looks good, need to renumber the migrations

@bbalser Done

bbalser (approved):

PR #987, opened 2025-04-08
Add new metric for latest file_info_poller file timestamp

bbalser:
Add a new metric to file_info_poller to record the timestamp of the file currently being processed.  Idea is to write alerts in the form of `time() - file_info_poller_timestamp` which will alert if the service has no new files to pick up

the current metric outputs `Utc::now() - file_timestamp` which will not alert until a new value is picked up.  Should we remove current metric?

### DIFF
--- a/file_store/src/file_info_poller.rs
+++ b/file_store/src/file_info_poller.rs
@@ -92,7 +92,7 @@ where
         ).set(latency.num_seconds() as f64);
 
         metrics::gauge!(
-            "file-info-poller-timestamp",
+            "file-processing-timestamp",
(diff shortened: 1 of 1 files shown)


jeffgrunewald (approved):

michaeldjeffrey:
Should this be in `FileInfoPollerServer::run`?

I guess my real question is, are we trying to add a metric to `file_store` about wether or not it's picking up files from S3?
Or a metric to make sure oracle code is processing those files in a timely fashion?

michaeldjeffrey (approved):

PR #992, opened 2025-04-15
Remove Generic Associated Error Types

michaeldjeffrey:
In every occurence where we have a GAT error, the ability to change the Error is not used. As all of the touched traits are for API Clients, and they all return ClienError, we can reduce the amount of generic constraints we need to type by removing them.

In a case where we do want to return an error, it is a simple process to return an applicable ClientError variant.

### DIFF
--- a/boost_manager/src/activator.rs
+++ b/boost_manager/src/activator.rs
@@ -11,7 +11,7 @@ use helium_proto::{
 };
 use mobile_config::{
     boosted_hex_info::{BoostedHex, BoostedHexes},
-    client::{hex_boosting_client::HexBoostingInfoResolver, ClientError},
+    client::hex_boosting_client::HexBoostingInfoResolver,
 };
 use poc_metrics::record_duration;
 use sqlx::{Pool, Postgres, Transaction};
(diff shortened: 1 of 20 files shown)


kurotych (commented):
>  we can reduce the amount of generic constraints 

As I understand it, after changing the error from a generic type (like Error: std::fmt::Debug + Send + Sync + 'static) to a specific type like ClientError, we are not reducing the constraints but actually adding a specific one — that the error can only be ClientError.

> it is a simple process to return an applicable ClientError variant.

Yes but every time when client want to have a new error, the error must be added to `ClientError` enum

michaeldjeffrey:
> As I understand it, after changing the error from a generic type (like Error: std::fmt::Debug + Send + Sync + 'static) to a specific type like ClientError, we are not reducing the constraints but actually adding a specific one — that the error can only be ClientError.

That's correct, we are reducing the need to specify a restraint that is only ever 1 value. In all cases where the generic was changed from a `ClientError` it was for the purpose of ignoring errors outright, which can be achieved by simply not using the `ClientError`.

> Yes but every time when client want to have a new error, the error must be added to ClientError enum

This was already the case. There were no implementations using the ability to provide their own error type. And all new errors were added to `ClientError`. If a good reason to override an error type comes up in the future, we can absolutely implement that use case. But as it stands, these GATs add nothing but visual noise imo.

bbalser (approved):

macpie (approved):

kurotych (approved):

PR #994, opened 2025-04-16
Remove Generics around HexAssignments

michaeldjeffrey:
We can remove some amount of indirection in hex boosting by reifying generics to types that are already supported.


kurotych (approved):

bbalser (approved):
LGTM

PR #1000, opened 2025-05-15
Add localstack integration and DataSetDownloaderDaemon test

kurotych:
AwsLocal is the code I use in 3 different projects to mock proto files in AWS.
So, I think it would be nice to have the core AwsLocal as workspace member in `oracles` repo to have capability to reuse it. 

https://github.com/novalabsxyz/mobile-rewards-estimator/pull/20#issuecomment-2695429042

### DIFF
--- a/aws_local/src/lib.rs
+++ b/aws_local/src/lib.rs
@@ -6,6 +6,7 @@ use file_store::traits::MsgBytes;
 use file_store::{file_sink, file_upload, FileStore, FileType, Settings};
 use std::path::Path;
 use std::{str::FromStr, sync::Arc};
+use tempfile::TempDir;
 use tokio::sync::Mutex;
 use tonic::transport::Uri;
 use uuid::Uuid;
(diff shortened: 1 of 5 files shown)


aws_local/src/lib.rs
```diff
@@ -0,0 +1,159 @@
...
+            let uploaded_files = item_recv.await.unwrap().unwrap();
+            assert!(uploaded_files.len() == 1);
+            let mut val = up_2.lock().await;
+            *val = uploaded_files.first().unwrap().to_string();
+
+            // After files uploaded to aws the must be removed.
+            // So we wait when dir will be empty.
+            // It means all files are uploaded to aws
+            loop {
+                if is_dir_has_files(&dir_path_clone) {
```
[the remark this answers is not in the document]
kurotych: `timeout -= dur;`
Causes panic and breaks loop if timeout happened

aws_local/src/lib.rs
```diff
@@ -0,0 +1,159 @@
...
+    }
+    pub async fn put_proto_to_aws<T: prost::Message + MsgBytes>(
+        &self,
+        items: Vec<T>,
+        file_type: FileType,
+        metric_name: &'static str,
+    ) -> Result<String> {
+        // Uuid uses as random to avoid colisions
+        let uuid: Uuid = Uuid::new_v4();
+        let dir_path = format!("/tmp/{}/{}", uuid, self.fs_settings.bucket);
```
michaeldjeffrey, line 80: Can we use `tempdir` here as well?
We don't need to do any cleanup ourselves. All files are removed when the TempDir is dropped.

In the PR we first encountered `AwsLocal` (https://github.com/novalabsxyz/mobile-rewards-estimator/pull/20), this function was never used. Is it in fact not used in any of your projects and we can drop it?

kurotych: @michaeldjeffrey 
This function is used at least in the CAS. It serves as the basis for functions that want to put a particular type of proto file.

https://github.com/novalabsxyz/mobile-coverage/blob/master/cas/tests/common/aws_mock.rs#L141

> Can we use tempdir here as well?
Yes, changed

bbalser (approved):

PR #1003, opened 2025-06-05
Add reward_override_entity_key to subscriber mapping activity

kurotych:
- Handle `reward_override_entity_key` in SubscriberMappingActivityIngestReportV1 report.
- Add `reward_override_entity_key` field to the `subscriber_mapping_activity` table
- Add `reward_override_entity_key` to SubscriberReward proto
- Handle `reward_override_entity_key` in reward_index (override the key/address in reward_index table)

### DIFF
--- a/mobile_verifier/src/reward_shares.rs
+++ b/mobile_verifier/src/reward_shares.rs
@@ -230,7 +230,9 @@ impl MapperShares {
                         subscriber_id: mas.subscriber_id,
                         discovery_location_amount,
                         verification_mapping_amount,
-                        entity_key: mas.entity_key.unwrap_or("".to_string()),
+                        reward_override_entity_key: mas
+                            .reward_override_entity_key
+                            .unwrap_or("".to_string()),
                     })),
                 },
             )
@@ -832,7 +834,7 @@ mod test {
                 subscriber_id: n.encode_to_vec(),
                 discovery_reward_shares: 30,
                 verification_reward_shares: 30,
-                entity_key: None,
+                reward_override_entity_key: None,
             })
         }
 
--- a/mobile_verifier/src/reward_shares.rs
+++ b/mobile_verifier/src/reward_shares.rs
@@ -230,6 +230,7 @@ impl MapperShares {
                         subscriber_id: mas.subscriber_id,
                         discovery_location_amount,
                         verification_mapping_amount,
+                        entity_key: mas.entity_key.unwrap_or("".to_string()),
                     })),
                 },
             )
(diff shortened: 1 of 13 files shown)


mobile_verifier/src/reward_shares.rs
```diff
@@ -230,6 +230,9 @@ impl MapperShares {
                         subscriber_id: mas.subscriber_id,
                         discovery_location_amount,
                         verification_mapping_amount,
+                        reward_override_entity_key: mas
+                            .reward_override_entity_key
+                            .unwrap_or("".to_string()),
```
jeffgrunewald, line 235:
```suggestion
                            .unwrap_or_default(),
```
will result in an empty string

reward_index/src/extract.rs
```diff
@@ -55,13 +55,21 @@ pub fn mobile_reward(
...
         )),
-        MobileReward::SubscriberReward(r) => Ok((
-            RewardKey {
-                key: bs58::encode(&r.subscriber_id).into_string(),
-                reward_type: RewardType::MobileSubscriber,
-            },
-            r.discovery_location_amount + r.verification_mapping_amount,
-        )),
+        MobileReward::SubscriberReward(r) => {
+            let key = if r.reward_override_entity_key.is_empty() {
```
jeffgrunewald, line 59:
```suggestion
            let key = if r.reward_override_entity_key.trim().is_empty() {
```

reward_index/src/extract.rs
```diff
@@ -55,13 +55,21 @@ pub fn mobile_reward(
...
+                bs58::encode(&r.subscriber_id).into_string()
+            } else {
+                r.reward_override_entity_key
+            };
+
+            Ok((
+                RewardKey {
+                    key,
+                    reward_type: RewardType::MobileSubscriber,
```
jeffgrunewald, line 68: should this be a new type of reward like `RewardType::MobileSubscriberOverride`?

michaeldjeffrey (approved):

kurotych:
> should we validate that the reward_override_entity_key is a valid on-chain entity if set

I think it could be a good defense against typos or a compromised service. 
At which step (service) do you think it would be better to implement it?

bbalser:
> > should we validate that the reward_override_entity_key is a valid on-chain entity if set
> 
> I think it could be a good defense against typos or a compromised service. At which step (service) do you think it would be better to implement it?

https://github.com/helium/oracles/blob/536569981c07fc30f3566a1ca43ad7d1f03aa8bc/mobile_verifier/src/subscriber_mapping_activity.rs#L180

bbalser (approved):

PR #1005, opened 2025-06-10
file store sync parser

michaeldjeffrey:
Make `FileInfoPollerParser` a non-async trait.

Simplify making a Parser by removing the need to know how to deal with a stream, and only have to deal with decoding a single message at a time.

The only required function to implement is `decode_item`, the rest can handle itself.

Default implementations are provided for hooking into custom logging for stream and decode errors. Or if you want to handle the whole stream by yourself.

These methods are instance methods to allow for stateful parsers.

### DIFF
--- a/file_store/tests/file_store_parsing.rs
+++ b/file_store/tests/file_store_parsing.rs
@@ -34,7 +34,7 @@ async fn single_message_decode() -> anyhow::Result<()> {
     };
 
     // Upload a simple file
-    let proto = SpeedtestIngestReportV1::try_from(speedtest_report.clone())?;
+    let proto = SpeedtestIngestReportV1::from(speedtest_report.clone());
     aws.put_proto_to_aws(vec![proto], FileType::CellSpeedtestIngestReport, "test")
         .await?;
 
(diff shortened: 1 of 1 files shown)


michaeldjeffrey:
@bbalser 
> can we use these same parsers when using other file-store functions directly?

Absolutely, I'm working on something to do that right now.

@kurotych 
> Could we make until optional parameter to allow run to be finishable?

I'd like to keep `FileInfoPoller` only for it's current purpose of polling for new files endlessly. I took a look through the mobile-coverage project, I think we could make some of those processes simpler without needing to lift an entire `FileInfoPoller`.

> Is it possible to add a flag/feature to completely interrupt and finish file_info_poller if handle_item() returned error?

I think this is an interesting idea. But I'd like to see a few more uses of how to reliably bail from processing a file before trying to add it as a feature.

michaeldjeffrey:
I stumbled across [cdr-service parsing avro files](https://github.com/novalabsxyz/cdr-service/blob/main/cdr_service/src/cdrs/sessions_loader.rs#L47-L86), and the implementation of avro's `Reader` unfortunately is causing this PR to die on the vine. 

The `Reader` can not own avro data without leaking it, which means we can't have a sync parser over it that doesn't require an unnecessary amount of hand holding.

PR #1006, opened 2025-06-11
Add ability to modify AWSLOCAL_ENDPOINT

macpie:


kurotych (commented):
Could you make default as http://localhost:4566/, please?|
localstack domain is only used on CI

bbalser (approved):

PR #1016, opened 2025-07-10
[Mobile Config] (gateway) info_stream_v3

kurotych:
- If a radio isn’t present in the mobile_radio_tracker table, it won’t be included in the endpoint response.
This change ensures that the location and location_changed_at fields remain consistent.
- Introduced the DeviceTypeV2. CBRS radios are no longer supported (in both request and response payloads).
- Remove `electricalDownTilt` and `mechanicalDownTilt` fields from deployment_info
- Remove `refreshed_at` from response due to uselessness. There are: `created_at`, `updated_at`,  `location_changed_at` fields

### DIFF
--- a/mobile_config/src/mobile_radio_tracker.rs
+++ b/mobile_config/src/mobile_radio_tracker.rs
@@ -227,7 +227,7 @@ fn get_all_mobile_radios(metadata: &Pool<Postgres>) -> impl Stream<Item = Mobile
     sqlx::query_as::<_, MobileRadio>(
         r#"
         SELECT
-            DISTINCT ON (kta.entity_key, mhi.asset)
+            DISTINCT ON (kta.entity_key)
             kta.entity_key,
             mhi.asset,
             mhi.refreshed_at,
@@ -244,7 +244,7 @@ fn get_all_mobile_radios(metadata: &Pool<Postgres>) -> impl Stream<Item = Mobile
             kta.asset = mhi.asset
         WHERE kta.entity_key IS NOT NULL
         	AND mhi.refreshed_at IS NOT NULL
-        ORDER BY kta.entity_key, mhi.asset, refreshed_at DESC
+        ORDER BY kta.entity_key, refreshed_at DESC
     "#,
     )
     .fetch(metadata)


mobile_config/src/gateway_info_v3.rs
```diff
@@ -0,0 +1,282 @@
...
+            DeviceTypeV2::Outdoor => "wifiOutdoor",
+            DeviceTypeV2::DataOnly => "wifiDataOnly",
+        }
+    }
+}
+
+#[derive(Clone, Debug)]
+pub struct GatewayInfoV3 {
+    pub address: PublicKeyBinary,
+    pub metadata: Option<GatewayMetadataV3>,
```
michaeldjeffrey, line 57: Is there a way receive a `GatewayInfoV3` without metadata being present?

mobile_config/src/gateway_info_v3.rs
```diff
@@ -0,0 +1,282 @@
...
+    async fn process_row(
+        row: sqlx::postgres::PgRow,
+        mtim: &MobileTrackerInfoMap,
+    ) -> Option<GatewayInfoV3> {
+        let device_type = DeviceTypeV2::from_str(
+            row.get::<Json<String>, &str>("device_type")
+                .to_string()
+                .as_ref(),
+        )
+        .map_err(|err| sqlx::Error::Decode(Box::new(err)))
```
michaeldjeffrey, line 233: Does we need to map this error? 

`.ok()?` will eat it and return None if it happens.

mobile_config/src/gateway_info_v3.rs
```diff
@@ -0,0 +1,282 @@
...
+
+        let updated_at = mti.last_changed_at;
+
+        let metadata = mti.location.and_then(|loc| {
+            let location_changed_at = mti.asserted_location_changed_at?;
+            let deployment_info = row
+                .try_get::<Option<sqlx::types::Json<DeploymentInfo>>, _>("deployment_info")
+                .ok()
+                .flatten()
+                .and_then(|json| match json.0 {
```
michaeldjeffrey, line 252: `sqlx::types::Json` provides `Deref` to it's inner type. We don't need to access the inner member here with `json.0`.

mobile_config/tests/gateway_service_v3.rs
```diff
@@ -0,0 +1,423 @@
...
+    let resp = stream.next().await.unwrap().unwrap();
+    assert_eq!(resp.gateways.len(), 2);
+
+    // Filter by device type
+    let req = make_gateway_stream_signed_req_v3(&admin_key, &[DeviceTypeV2::Indoor], 0, 0);
+    let mut stream = client.info_stream_v3(req).await.unwrap().into_inner();
+    let resp = stream.next().await.unwrap().unwrap();
+    assert_eq!(resp.gateways.len(), 1);
+    let gateway: &GatewayInfoV3 = resp.gateways.first().unwrap();
+    assert_eq!(gateway.device_type, Into::<i32>::into(DeviceTypeV2::Indoor));
```
michaeldjeffrey, line 76: Rust structs with enums provide a method that returns their reified enum variant.
```suggestion
    assert_eq!(gateway.device_type(), DeviceTypeV2::Indoor);
```

mobile_config/src/gateway_info_v3.rs
kurotych: Yes, there is. Added `gateway_stream_info_v3_no_metadata`  test

mobile_config/src/gateway_info_v3.rs
```diff
@@ -0,0 +1,282 @@
...
+                |mut map: MobileTrackerInfoMap, row| async move {
+                    let entity_key_b = row.get::<&[u8], &str>("entity_key");
+                    let entity_key = bs58::encode(entity_key_b).into_string();
+                    let last_changed_at = row.get::<DateTime<Utc>, &str>("last_changed_at");
+                    let asserted_location_changed_at =
+                        row.get::<Option<DateTime<Utc>>, &str>("asserted_location_changed_at");
+                    let asserted_location = row.get::<Option<i64>, &str>("asserted_location");
+
+                    map.insert(
+                        PublicKeyBinary::from_str(&entity_key)?,
```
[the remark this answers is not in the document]
kurotych: I don't think so. It fails when I'm trying use it. 

To restore PubKey, we first need to encode the bytea to base58, and then convert it from base58 into a native Rust struct using PublicKeyBinary::from_str.

The PublicKeyBinary::from_str itself drops first byte after decoding.



image

michaeldjeffrey (approved):

PR #1019, opened 2025-07-17
Update AWS Dependencies (Welcome to the Jungle)

michaeldjeffrey:
## Version Updates
| crate | old | new |
|---|---|---|
|`aws-config` | `0.51.0` | `1.8.2` |
| `aws-sdk-sts` | `0.21.0` | `1.76.0` |
| `aws-credential-types` | `0.54.1` | `1.2.4` |
| `aws-smithy-http` | `0.51.0` | removed |
| `aws-types` | `0.51.0` | `1.3.7` |
| `aws-sig-auth` | `0.51.0` | removed |
| `aws-sigv4` | `0.51.1` | `1.3.3` 
| `aws-sdk-s3` | `0.21.0` | `1.98.0` |
| `aws-smithy-types-convert` | -- | `0.60.9` |

## Helium Crypto

Updating AWS deps was initially fraught with danger because of crypto libraries used for signing. Mainly the `signature` crate that provided the `Signature` trait. 

See: https://github.com/helium/helium-crypto-rs/pull/74

## Clippy Updates

`aws-sdk-sts` has a msvr of `1.86.0`. 

`rust-toolchain.toml` was updated, along with `coverage-point-calculator` and `custom-tracing` for new clippy warnings. 
Those were `path.next_back()` and `std::iter::repeat_n()`.

## DB Store

`generate_rds_iam_token()` needed to be updated to use `aws-sigv4` instead of `aws-sig-auth`. 

This commit (https://github.com/helium/oracles/commit/64632a3bfa9d65aa85c41ccf79caabce7e340c03) was the point in time where I had both versions running to ensure they produced the same iam token.

### `features=local` and path style enforcement
The `local` feature flag no longer includes `aws-types`.
`Credentials` are now provided by their own `aws-sdk-*` crate.

The `local` feature now enforces path style access to s3 buckets.

Path style access is a slow deprecation being replaced by DNS resolution for buckets.

However, when running against a localstack or minio, unless you want to setup some custom DNS rules on in your dev environment, path style access is still the way to go.

### Boxing AWS Errors

Clippy has started complaining about Errors and Results that have large variants.
The compiler doesn't know what error might be throw so has to allocate enough space for the largest possible error.

One of the ways around this, is to `Box` potentially large errors.

### `into_stream_03x()`

Somewhere around `aws-sdk-s3@1` the paginated version of `ListObjectsV2` lost the ability to be turned into a flat stream. 
`aws-smithy-types-convert` was provided as a way to bring that back.

https://github.com/smithy-lang/smithy-rs/pull/3299 explains the bizarre naming.


custom_tracing/src/lib.rs
```diff
@@ -157,6 +157,6 @@ fn file_match(event_path: &Path, file: String) -> bool {
     event_path
         .to_str()
         .map(|path| path.split('/'))
-        .map(|path| path.last().is_some_and(|file_name| file_name == file))
+        .map(|mut path| path.next_back().is_some_and(|file_name| file_name == file))
```
kurotych, line 160: What is the reason you have changed it? Looks the same

kurotych (approved):

macpie (approved):

michaeldjeffrey:
Given the issue with building a release with the updated aws dependencies (mainly `aws-lc-sys`), this PR is waiting until we've completed the migration to kubernetes.

michaeldjeffrey:
Latest set of updates caused by `helium-proto` https://github.com/helium/proto/pull/459

- update `helium-lib`
- update `helium-proto`
- `tonic` and `prost` versions picked up from `helium-proto`
  - `tonic` no longer exported from `helium-proto`, fix imports
- removed `credential_load_timeout` in `file_store`
  - functionality was removed many versions back in `aws-s3-sdk`, if we still have problems we can attempt to recreate it with the new structure, but it's a bit more involved.
- remove calls and tests to gateway service info v1.

michaeldjeffrey:
In effort to get these changes landed before the full switchover of iot to kubernetes. Builds are now using the `AWS_LC_SYS_CMAKE_BUILDER=1` option so `aws-lc-sys` will use a non-bugged version of `cc` for building in an `unbuntu:20.04` environment.

PR #1031, opened 2025-08-15
Add 300Mbps speedtest validation with bounds checking

kurotych:
Implements SpeedtestValueOutOfBounds validation to not count speedtests exceeding 300Mbps upload/download limits.

### DIFF
--- a/mobile_verifier/tests/integrations/speedtests.rs
+++ b/mobile_verifier/tests/integrations/speedtests.rs
@@ -125,7 +125,9 @@ async fn speedtest_upload_exceeds_300megabits_ps_limit(pool: Pool<Postgres>) -> 
 }
 
 #[sqlx::test]
-async fn speedtest_download_exceeds_300mb_limit(pool: Pool<Postgres>) -> anyhow::Result<()> {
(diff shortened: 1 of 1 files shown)


michaeldjeffrey:
- [HIP-98 speedtest](https://github.com/helium/HIP/blob/main/0098-mobile-subdao-quality-of-service-requirements.md#speed-test-reward-tiers)
- [speedtest proto](https://github.com/helium/proto/blob/0ba63207fa6935861ad2c3649edbb3c74f7f87ec/src/service/poc_mobile.proto#L137-L140)

> [..] we have limits 150Mbit for download and 15Mbit for upload configured on the HMH [..]

@bbalser and I were reviewing this earlier, and believe we have a confusion between `Mbps` and `MBps`. 

We should be checking these values against Megabits/ps.

Probably [this constant in the coverage point calculator](https://github.com/helium/oracles/blob/main/coverage_point_calculator/src/speedtest.rs#L15) should be renamed. The value is correct, but the name is incorrect. Should be `const BYTES_PER_MEGABIT: u64 = 125_000;`.

I think all the tests should be using the `mbps()` helper rather than `BYTES * 1024 * 1024`.

PR #1034, opened 2025-08-19
Update settings to have more default values and be able to to override with env variables

bbalser:


mobile_config/src/settings.rs
```diff
@@ -15,8 +15,9 @@ pub struct Settings {
     /// Listen address. Required. Default to 0.0.0.0::8080
     #[serde(default = "default_listen_addr")]
     pub listen: SocketAddr,
-    /// File from which to load config server signing keypair
-    pub signing_keypair: String,
+    /// Base64 encoded string with bytes of helium keypair
+    #[serde(deserialize_with = "crate::deserialize_keypair")]
+    pub signing_keypair: Arc<Keypair>,
     /// B58 encoded public key of the default admin keypair
     pub admin_pubkey: String,
```
michaeldjeffrey, line 22: Do we want to make this a `PubKey` like `mobile_config/src/client/settings.rs`?

bbalser: Yes that should be a `PubKey`

michaeldjeffrey (approved):

mobile_verifier/src/settings.rs
```diff
@@ -55,11 +59,11 @@ fn default_fencing_resolution() -> u8 {
 
 fn default_max_distance_from_coverage() -> u32 {
     // Default is 2 km
-    2000
+    3000
```
kurotych, line 62: Update (or remove) the comment above

kurotych (approved):

macpie (approved):

PR #1041, opened 2025-09-08
[DRAFT] Remove HIP-84 requirement for hotspots in mexico 

bbalser:


coverage_point_calculator/src/lib.rs
```diff
@@ -247,7 +247,6 @@ impl CoveragePoints {
             SpBoostedHexStatus::Eligible => self.coverage_points.boosted,
             SpBoostedHexStatus::WifiLocationScoreBelowThreshold(_) => dec!(0),
             SpBoostedHexStatus::AverageAssertedDistanceOverLimit(_) => dec!(0),
-            SpBoostedHexStatus::RadioThresholdNotMet => dec!(0),
```
kurotych, line 250: Should we also remove these comments?

- https://github.com/helium/oracles/blob/main/coverage_point_calculator/src/lib.rs#L52
- https://github.com/helium/oracles/blob/main/coverage_point_calculator/src/lib.rs#L24

kurotych (commented):
"... in mexico"
I can't get where we check that hotspot is specific to mexico

PR #1044, opened 2025-09-09
Update Gateway APIs

macpie:
The goal is to stop querying the metadata database directly and instead maintain a copy of that data in the config database, using it as the source.

### DIFF
--- a/mobile_config/src/gateway/tracker.rs
+++ b/mobile_config/src/gateway/tracker.rs
@@ -65,14 +65,14 @@ pub async fn execute(pool: &Pool<Postgres>, metadata: &Pool<Postgres>) -> anyhow
 
     const BATCH_SIZE: usize = 1_000;
 
-    let (_, total): (&Pool<Postgres>, u64) = MobileHotspotInfo::stream(metadata)
+    let total: u64 = MobileHotspotInfo::stream(metadata)
         .map_err(anyhow::Error::from)
         .try_filter_map(|mhi| async move { mhi.to_gateway() })
         .try_chunks(BATCH_SIZE)
         .map_err(|TryChunksError(_gateways, err)| err)
-        .try_fold((pool, 0), |(p, total), batch| async move {
-            let affected = Gateway::insert_bulk(p, &batch).await?;
-            Ok((p, total + affected))
+        .try_fold(0, |total, batch| async move {
+            let affected = Gateway::insert_bulk(pool, &batch).await?;
+            Ok(total + affected)
         })
         .await?;
 
--- a/mobile_config/tests/integrations/gateway_db.rs
+++ b/mobile_config/tests/integrations/gateway_db.rs
@@ -1,4 +1,4 @@
-use chrono::{TimeZone, Timelike, Utc};
+use chrono::{TimeZone, Utc};
 use futures::{pin_mut, StreamExt};
 use helium_crypto::PublicKeyBinary;
 use mobile_config::gateway::db::{Gateway, GatewayType};
@@ -9,7 +9,7 @@ use crate::common;
 #[sqlx::test]
 async fn gateway_insert_and_get_by_address(pool: PgPool) -> anyhow::Result<()> {
     let addr = pk_binary();
-    let now = now_truncated();
+    let now = Utc::now();
 
     let gateway = gw(addr.clone(), GatewayType::WifiIndoor, now);
     gateway.insert(&pool).await?;
@@ -19,10 +19,10 @@ async fn gateway_insert_and_get_by_address(pool: PgPool) -> anyhow::Result<()> {
         .expect("gateway should exist");
 
     assert_eq!(gateway.gateway_type, GatewayType::WifiIndoor);
-    assert_eq!(gateway.created_at, now);
-    assert_eq!(gateway.updated_at, now);
-    assert_eq!(gateway.refreshed_at, now);
-    assert_eq!(gateway.last_changed_at, now); // first insert: equals refreshed_at
+    assert_eq!(gateway.created_at, common::nanos_trunc(now));
+    assert_eq!(gateway.updated_at, common::nanos_trunc(now));
+    assert_eq!(gateway.refreshed_at, common::nanos_trunc(now));
+    assert_eq!(gateway.last_changed_at, common::nanos_trunc(now)); // first insert: equals refreshed_at
     assert_eq!(gateway.location, Some(123));
     assert_eq!(gateway.hash, "h0");
     Ok(())
@@ -81,7 +81,7 @@ async fn gateway_upsert_last_changed_at_on_location_or_hash_change(
 
 #[sqlx::test]
 async fn gateway_bulk_insert_and_get(pool: PgPool) -> anyhow::Result<()> {
-    let now = now_truncated();
+    let now = Utc::now();
 
     let a1 = pk_binary();
     let a2 = pk_binary();
@@ -98,10 +98,10 @@ async fn gateway_bulk_insert_and_get(pool: PgPool) -> anyhow::Result<()> {
         let got = Gateway::get_by_address(&pool, addr)
             .await?
             .expect("row should exist");
-        assert_eq!(got.created_at, now);
-        assert_eq!(got.updated_at, now);
-        assert_eq!(got.refreshed_at, now);
-        assert_eq!(got.last_changed_at, now);
+        assert_eq!(got.created_at, common::nanos_trunc(now));
+        assert_eq!(got.updated_at, common::nanos_trunc(now));
+        assert_eq!(got.refreshed_at, common::nanos_trunc(now));
+        assert_eq!(got.last_changed_at, common::nanos_trunc(now));
         assert_eq!(got.location, Some(123));
         assert_eq!(got.hash, "h0");
     }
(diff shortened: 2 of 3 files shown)


mobile_config/src/main.rs
```diff
@@ -48,9 +47,7 @@ impl Cli {
             Cmd::PostMigrateMobileTracker => {
```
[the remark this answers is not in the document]
macpie: removed

mobile_config/src/cli/daemon.rs
```diff
@@ -0,0 +1,93 @@
...
+    hex_boosting_service::HexBoostingService,
+    key_cache::KeyCache,
+    settings::Settings,
+    sub_dao_service::SubDaoService,
+};
+
+use task_manager::TaskManager;
+
+#[derive(Debug, clap::Args)]
+pub struct Daemon;
```
michaeldjeffrey, line 17: If we're going to follow the pattern, can we rename this to `Server`?

It seems weird to have the cli command be `server` that delegates to a file called `deamon.rs`.

mobile_config/src/client/mod.rs
```diff
@@ -1,21 +1,20 @@
...
+pub use crate::gateway::client::GatewayClient;
 pub use authorization_client::AuthorizationClient;
 pub use carrier_service_client::CarrierServiceClient;
 pub use entity_client::EntityClient;
-pub use gateway_client::GatewayClient;
 pub use settings::Settings;
 pub use sub_dao_client::SubDaoClient;
 
-const CACHE_EVICTION_FREQUENCY: Duration = Duration::from_secs(60 * 60);
+pub const CACHE_EVICTION_FREQUENCY: Duration = Duration::from_secs(60 * 60);
```
michaeldjeffrey, line 17: Is this being exported up? does it need to be `pub`?

mobile_config/src/gateway/db.rs
```diff
@@ -0,0 +1,365 @@
...
+        .bind(address.as_ref())
+        .fetch_optional(db)
+        .await?;
+
+        Ok(gateway)
+    }
+
+    pub fn stream_by_addresses<'a>(
+        db: impl PgExecutor<'a> + 'a,
+        addresses: &[PublicKeyBinary],
```
michaeldjeffrey, line 270: addresses could be `Vec<PublicKeyBinary>`. 

Everywhere this function is called, the addresses are never re-used. 
Ownership could be given, instead of cloning the contents of the slice just for the source to be dropped without doing anything else.

And the first line could be
```rs
let addr_array: Vec<Vec<u8>> = addresses.into_iter().map(|a| a.into()).collect();
```

michaeldjeffrey: That ownership could flow up through `gateway::service::info::stream_by_addresses()` as well.

mobile_config/src/gateway/db.rs
```diff
@@ -0,0 +1,365 @@
...
+
+impl Gateway {
+    pub async fn insert_bulk(pool: &PgPool, rows: &[Gateway]) -> anyhow::Result<u64> {
+        if rows.is_empty() {
+            return Ok(0);
+        }
+
+        // IMPORTANT: no trailing comma in the column list
+        let mut qb = QueryBuilder::<Postgres>::new(
+            "INSERT INTO gateways (\
```
michaeldjeffrey, line 108: Are all these backslashes needed?

mobile_config/src/gateway/db.rs
```diff
@@ -0,0 +1,365 @@
...
+                    $3::timestamptz IS NULL
+                    OR (location IS NOT NULL AND location_changed_at >= $3)
+                )
+            "#,
+        )
+        .bind(types)
+        .bind(min_date)
+        .bind(min_location_changed_at)
+        .fetch(db)
+        .map_err(anyhow::Error::from)
```
michaeldjeffrey, line 338: Should we be logging here or adding some metrics to know if our db is starting to fill with rows we can't decode?

mobile_config/tests/integrations/gateway_db.rs
```diff
@@ -0,0 +1,197 @@
...
+use mobile_config::gateway::db::{Gateway, GatewayType};
+use sqlx::PgPool;
+
+use crate::common;
+
+#[sqlx::test]
+async fn gateway_insert_and_get_by_address(pool: PgPool) -> anyhow::Result<()> {
+    let addr = pk_binary();
+    let now = Utc::now()
+        .with_nanosecond(Utc::now().timestamp_subsec_micros() * 1000)
```
michaeldjeffrey, line 13: Why are we settings nanoseconds?

mobile_config/src/cli/daemon.rs
macpie: 👍🏻

mobile_config/src/client/mod.rs
macpie: it is but I'll move it

mobile_config/src/gateway/db.rs
macpie: IMO these should not happen or they would explode during insertion so I don't think so

mobile_config/src/gateway/db.rs
```diff
@@ -0,0 +1,365 @@
...
+            updated_at: row.try_get("updated_at")?,
+            refreshed_at: row.try_get("refreshed_at")?,
+            last_changed_at: row.try_get("last_changed_at")?,
+            hash: row.try_get::<String, _>("hash")?,
+            antenna: to_u32(row.try_get::<Option<i64>, _>("antenna")?),
+            elevation: to_u32(row.try_get::<Option<i64>, _>("elevation")?),
+            azimuth: to_u32(row.try_get::<Option<i64>, _>("azimuth")?),
+            location: to_u64(row.try_get::<Option<i64>, _>("location")?),
+            location_changed_at: row.try_get::<Option<DateTime<Utc>>, _>("location_changed_at")?,
+            location_asserts: to_u32(row.try_get::<Option<i64>, _>("location_asserts")?),
```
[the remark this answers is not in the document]
macpie: Yes had them at first for my sanity but looks like not needed will remove

mobile_config/src/gateway/metadata_db.rs
```diff
@@ -0,0 +1,173 @@
...
+    is_full_hotspot: Option<bool>,
+    num_location_asserts: Option<i32>,
+    is_active: Option<bool>,
+    dc_onboarding_fee_paid: Option<i64>,
+    device_type: DeviceType,
+    deployment_info: Option<DeploymentInfo>,
+}
+
+impl MobileHotspotInfo {
+    fn hash(&self) -> String {
```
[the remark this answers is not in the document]
macpie: dont know why blake3 vs Hash but I can at least update fun name right now

mobile_config/tests/integrations/gateway_db.rs
macpie: so this happens in test cause one does nano and the other micro so they dont match note this only happens on GA

mobile_config/src/gateway/service/info_v3.rs
```diff
@@ -0,0 +1,167 @@
...
+    }
+}
+
+impl TryFrom<GatewayInfoV3> for GatewayInfoProtoV3 {
+    type Error = hextree::Error;
+
+    fn try_from(info: GatewayInfoV3) -> Result<Self, Self::Error> {
+        let metadata = if let Some(metadata) = info.metadata {
+            let location_info = LocationInfoProto {
+                location: hextree::Cell::from_raw(metadata.location_info.location)?.to_string(),
```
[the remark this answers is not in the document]
macpie: done

mobile_config/src/gateway/tracker.rs
```diff
@@ -0,0 +1,84 @@
...
+    let start = Instant::now();
+
+    const BATCH_SIZE: usize = 1_000;
+
+    let (_, total): (&Pool<Postgres>, u64) = MobileHotspotInfo::stream(metadata)
+        .map_err(anyhow::Error::from)
+        .try_filter_map(|mhi| async move { mhi.to_gateway() })
+        .try_chunks(BATCH_SIZE)
+        .map_err(|TryChunksError(_gateways, err)| err)
+        .try_fold((pool, 0), |(p, total), batch| async move {
```
michaeldjeffrey, line 73: This operation doesn't outlast the function, `pool` can be used as a reference in the `async` blocks without being seeded into `try_fold`.

mobile_config/tests/integrations/gateway_db.rs
```diff
@@ -0,0 +1,309 @@
...
+    assert!(s.next().await.is_none());
+
+    Ok(())
+}
+
+fn now_truncated() -> chrono::DateTime<Utc> {
+    let n = Utc::now();
+    let micros = n.timestamp_subsec_micros();
+    n.with_nanosecond(micros * 1_000).unwrap()
+}
```
michaeldjeffrey, line 287: https://github.com/helium/oracles/blob/main/reward_index/tests/integrations/common/mod.rs#L18-L23

We're doing something like this in a few places.
Might be a good time to consolidate, or update our CI postgres to be consistent with prod.

michaeldjeffrey (approved):

mobile_config/src/cli/mod.rs
```diff
@@ -0,0 +1,61 @@
...
+}
+
+impl Cli {
+    pub async fn run(self) -> anyhow::Result<()> {
+        match self.cmd {
+            Cmd::Server(server) => {
+                let settings = Settings::new(self.config)?;
+                server.run(&settings).await
+            }
+            Cmd::GenerateKey => {
```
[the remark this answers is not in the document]
macpie: Sure, this is useful for tests

PR #1048, opened 2025-09-15
Remove FileStore struct from file_store

bbalser:
An issue was identified when there serveral instances of aws_sdk_s3::Client created when the credentials are refreshed.  
When trying to refresh credentials it will sometimes error saying no credentials found.  The best fix identified was to create a single aws_sdk_s3::Client and clone it rather than creating new ones per bucket.  

To address this, the FileStore struct was deleted and all low level methods from it have been moved to free functions on the file_store crate.  These low level functions now take a file_store::Client and bucket as arguments.


file_store/src/cli/bucket.rs
```diff
@@ -85,14 +87,16 @@ impl FileFilter {
...
+    bucket: String,
     #[clap(flatten)]
     filter: FileFilter,
 }
 
 impl List {
     pub async fn run(&self, settings: &Settings) -> Result {
-        let store = FileStore::from_settings(settings).await?;
-        let mut file_infos = self.filter.list(&store);
+        let client = settings.connect().await;
```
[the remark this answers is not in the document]
bbalser: I was following the pattern we have established with `db_store`, I thought it would feel good for both to follow the same pattern.

file_store/src/file_info_poller.rs
```diff
@@ -130,11 +130,19 @@ pub struct FileInfoPollerConfig<Message, State, Store, Parser> {
     p: PhantomData<Message>,
 }
 
+impl<Message, State, Parser>
+    FileInfoPollerConfigBuilder<Message, State, FileStoreInfoPollerStore, Parser>
+{
+    pub fn file_store(self, client: crate::Client, bucket: String) -> Self {
```
michaeldjeffrey, line 136: Might be an opportunity to be nice here and accept anything that can be `.to_string()`.

```rs
pub fn file_store(self, client: crate::Client, bucket: impl Into<String>) -> Self { /* ... */ }
```

This would mean downstream could pass a `&str`, or decide to manage their buckets through an enum.

file_store/src/file_info_poller.rs
```diff
@@ -179,7 +187,8 @@ where
     }
 }
 
-impl<Message, State, Parser> ManagedTask for FileInfoPollerServer<Message, State, FileStore, Parser>
+impl<Message, State, Parser> ManagedTask
```
michaeldjeffrey, line 190: I don't remember why I hardcoded `FileStore` in here. But maybe we want to make this generic over `Store` like the impl block below.

That way someone could still use `task-manager` if they decide to implement `FileInfoPollerStore` themselves.

file_store/src/file_upload.rs
```diff
@@ -75,14 +67,13 @@ impl ManagedTask for FileUploadServer {
...
     pub async fn run(self, shutdown: triggered::Listener) -> Result {
-        tracing::info!("starting file uploader {}", self.store.bucket);
+        tracing::info!("starting file uploader {}", self.bucket);
 
         let uploads = self
             .messages
-            .map(|msg| (self.store.clone(), msg))
-            .for_each_concurrent(5, |(store, path)| async move {
+            .map(|msg| (self.client.clone(), self.bucket.clone(), msg))
+            .for_each_concurrent(5, |(client, bucket, path)| async move {
```
michaeldjeffrey, line 75: Potential for future updates.

This function takes `self`, we can grab references to instead of cloning.

```rs
let client = &self.client;
let bucket = &self.bucket;
```

That would allow us to remove the `.map()` call.

file_store/src/settings.rs
```diff
@@ -35,4 +25,13 @@ impl Settings {
             .and_then(|config| config.try_deserialize())
             .map_err(Error::from)
     }
+
+    pub async fn connect(&self) -> crate::Client {
+        crate::new_client(
+            self.endpoint.clone(),
+            self.access_key_id.clone(),
+            self.secret_access_key.clone(),
+        )
```
michaeldjeffrey, line 34: Our production happy path being to call this function where we expect all the settings to not be provided makes me feel like we're missing an abstraction. I'm not sure what. But I'll be thinking.

iot_verifier/src/settings.rs
```diff
@@ -4,6 +4,14 @@ use humantime_serde::re::humantime;
 use serde::Deserialize;
 use std::{path::Path, time::Duration};
 
+#[derive(Debug, Deserialize, Clone)]
+pub struct Buckets {
```
michaeldjeffrey, line 8: This is nice.

price/src/main.rs
```diff
@@ -80,9 +80,11 @@ impl Server {
         // Install the prometheus metrics exporter
         poc_metrics::start_metrics(&settings.metrics)?;
 
+        let s3_client = settings.file_store.connect().await;
```
michaeldjeffrey, line 83: Everywhere else this is `file_store_client`.

michaeldjeffrey (approved):

file_store/src/lib.rs
```diff
@@ -40,8 +44,213 @@ pub use iot_valid_poc::SCALING_PRECISION;
...
+    let bucket = bucket.to_string();
+    let client = client.clone();
+    infos
+        .map_ok(move |info| get_byte_stream(client.clone(), bucket.clone(), info.key))
+        .try_buffer_unordered(workers)
+        .flat_map(|stream| match stream {
+            Ok(stream) => stream_source(stream),
+            Err(err) => stream::once(async move { Err(err) }).boxed(),
+        })
+        // .fuse()
```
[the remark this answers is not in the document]
bbalser: mistake, was testing something and forgot to uncomment

file_store/src/settings.rs
bbalser: it is a fair point, in our production settings we inherit all settings to connect to s3 from the environment.  These properties are only necessary when doing local testing.

PR #1055, opened 2025-09-29
in gateway tracker, update location_changed_at for new records being …

bbalser:
…inserted


mobile_config/src/gateway/db.rs
```diff
@@ -349,31 +350,41 @@ impl Gateway {
...
             }
-        }
 
-        qb.push(
-            ") AS v(address, location_changed_at) \
-            WHERE g.address = v.address \
-              AND g.location_changed_at IS NULL",
-        );
+            qb.push(
+                ") AS v(address, location_changed_at, location) \
```
michaeldjeffrey, line 377: Do we need these backslashes?

bbalser: I thought we always preferred the multi line string syntax for larger text blocks?

mobile_config/src/gateway/db.rs
```diff
@@ -335,6 +341,51 @@ impl Gateway {
...
+
+        for chunk in updates.chunks(MAX_ROWS) {
+            let mut qb = QueryBuilder::<Postgres>::new(
+                "UPDATE gateways AS g \
+                SET location_changed_at = v.location_changed_at \
+                FROM (VALUES ",
+            );
+
+            {
+                let mut rows = qb.separated(", ");
```
bbalser, line 364: Why not use `push_values` to stay consistent?

Something like:
```
        let mut q = QueryBuilder::<Postgres>::new(
            r#"
            UPDATE gateways as g
            SET location_changed_at = v.location_changed_at
            FROM (
        "#,
        );

        q.push_values(vec![(1, 1, 1), (2, 2, 2)], |mut b, (one, two, three)| {
            b.push_bind(one).push_bind(two).push_bind(three);
        });

        q.push(
            r#"
            ) as v(address, location_changed_at, location)
            where g.address = v.address
                and g.locaton_changed_at is null
                and g.location = v.location
        "#,
        );
```

PR #1058, opened 2025-10-02
Publish on-chain data to s3 buckets

ChewingGlass:


file_store/src/traits/file_sink_write.rs
```diff
@@ -313,3 +313,23 @@ impl_file_sink!(
     FileType::VerifiedSubscriberMappingActivityReport.to_str(),
     "verified_subscriber_mapping_activity_report"
 );
+impl_file_sink!(
+    chain_rewardable_entities::MobileHotspotChangeReportV1,
+    FileType::MobileHotspotChangeReport.to_str(),
+    "chain_rewardable_entities_mobile_hotspot_change_report"
```
dcyoung, line 319: I'm not entirely sure on naming conventions here... but should these match their proto equivalent? ie: `..._v1` in case v2s are introduced later on?

michaeldjeffrey: The last argument for this macro is the name for a metric.

For context.
There are 3 occurences of written files with a `_v{x}` in the name. 
And all of their `FileType` counterparts do not reference the version.

michaeldjeffrey (approved):

PR #1059, opened 2025-10-03
FileInfoPollerConfigBuilder add builder setter

michaeldjeffrey:
Add new setters that are prefixed with `lookback_` to make them discoverable when setting `lookback`, and it’s a nice place to hang some documentation

### DIFF
--- a/file_store/src/file_info_poller.rs
+++ b/file_store/src/file_info_poller.rs
@@ -131,7 +131,6 @@ pub struct FileInfoPollerConfig<Message, State, Store, Parser> {
     store: Store,
     prefix: String,
     parser: Parser,
-    #[builder(setter(into))]
     lookback: LookbackBehavior,
     #[builder(default = "DEFAULT_OFFSET_DURATION")]
     offset: Duration,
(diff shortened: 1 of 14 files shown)


bbalser:
> I don’t totally agree the change reduces _readability_ but it puts the onus on the caller to name that datetime variable something like “start_after” to keep the same readability, which we just happen to have done. Feels like it could get confusing if we deviated from that pattern.
> 
> Am I right that you could still use the lookbackbehavior enum if you wanted to or would that make clippy complain about an unnecessary from conversion or something?
> 
> However, maybe bigger question is: do we even need the lookback behavior? We don’t seem to use the duration based variant at all. What if we scrapped it altogether and just did lookback from a fixed datetime?

We do use the max lookback behavior in the IOT verifier

jeffgrunewald:
> > However, maybe bigger question is: do we even need the lookback behavior? We don’t seem to use the duration based variant at all. What if we scrapped it altogether and just did lookback from a fixed datetime?
> 
> We do use the max lookback behavior in the IOT verifier

Seems like that answers my other question then since the original commit didn't include the iot verifier max lookback builder which means you _could_ use the previous `From/Into` impl and just explicitly pass the lookback enum if you preferred

PR #1173, opened 2026-04-16
Update publish_branch to detect and perform a rebase

bbalser:


helium_iceberg/src/branch.rs
```diff
@@ -345,6 +451,50 @@ fn validate_branch_name(branch_name: &str) -> Result<()> {
...
+    };
+    writer
+        .add_manifests(manifests.into_iter())
+        .map_err(Error::Iceberg)?;
+    writer.close().await.map_err(Error::Iceberg)?;
+    Ok(manifest_list_path)
+}
+
+fn current_timestamp_ms() -> Result<i64> {
+    std::time::SystemTime::now()
```
michaeldjeffrey, line 492: We have `chrono` available to us, I think we should consider using it here.
Currently it's only used in testing, but I think it's a normal enough dependency to have that we can rely on it.

`Utc::now().timestamp_millis()` returns a bare `i64` which I think poses a good question of, if we can't get the SystemTime, what are we even doing?


There's also a usage of `SystemTime` in `iceberg_table.rs` I think we should hit up with this treatment.

helium_iceberg/src/branch.rs
```diff
@@ -345,6 +451,50 @@ fn validate_branch_name(branch_name: &str) -> Result<()> {
...
+/// location for the table, handling the V1/V2/V3 writer variants. Returns
+/// the path the manifest list was written to.
+async fn write_manifest_list(
+    table: &Table,
+    snapshot_id: i64,
+    commit_uuid: Uuid,
+    parent_snapshot_id: Option<i64>,
+    sequence_number: i64,
+    manifests: impl IntoIterator<Item = ManifestFile>,
+) -> Result<String> {
```
michaeldjeffrey, line 464: This function takes place solely in iceberg land as it is now, maybe we can return a `iceberg::Result<String>` so we can make better error mapping decisions if that ever changes in the future.

helium_iceberg/src/branch.rs
```diff
@@ -276,13 +253,30 @@ pub(crate) async fn publish_branch(
...
+    .await
+}
 
-        let metadata = table_guard.metadata();
-        let branch_snapshot = metadata
-            .snapshot_for_ref(branch_name)
-            .ok_or_else(|| Error::Branch(format!("branch '{branch_name}' does not exist")))?;
-        let branch_snapshot_id = branch_snapshot.snapshot_id();
-
+async fn publish_branch_once(
```
michaeldjeffrey, line 264: We should take a pass through these functions sometime soon.
I'm getting a bit lost with the difference between `commit_to_branch`, `publish_to_branch`, and `rebase_branch_onto_main`, when they all end up at `commit`. 

It'd be nice if we could clean up the amount of code that we have to read through doing iceberg things and make the top level api of this file easier to understand.

michaeldjeffrey (approved):

PR #1175, opened 2026-04-17
Simplify helium-iceberg by using set_snapshot_properties

bbalser:
## Motivation

  The branch-based Write-Audit-Publish workflow in `helium_iceberg` had a latent bug: `publish_branch` did a
  non-rebasing fast-forward that overwrote any commit to `main` made between branch creation and publish. A
  `DELETE` issued against `mobile.data_transfer.session` while `mobile-packet-verifier` was running would
  briefly disappear and then reappear after the next publish — the branch's snapshot (created from the
  pre-delete main) was promoted to main, orphaning the delete.

  The only capability the codebase actually uses WAP for is **idempotency** — "don't re-commit the same file".
  That can be provided by stamping an id onto the snapshot summary and checking for it before writing, with no
  branch state machine and no client-side rebase logic.

  This PR replaces ~520 lines of `branch.rs` plus the three-trait state machine (`BranchTransaction` /
  `BranchWriter` / `BranchPublisher`) with a single trait method implemented on top of iceberg-rust's native
  `fast_append`.

  ## API change

  **Old** — three-step state machine:

  ```rust
  let mut txn = writer.begin(wap_id).await?;
  txn.write(records).await?;
  txn.publish().await?;
  ```

  **New** — one call:

  ```rust
  writer.write_idempotent(id, records).await?;
  ```

  The `DataWriter` trait now has just two methods:

  ```rust
  async fn write(&self, records: Vec<T>) -> Result;
  async fn write_idempotent(&self, id: &str, records: Vec<T>) -> Result;
  ```

  Idempotency is keyed by an application-provided `id` stamped onto the snapshot summary under
  `helium.write_id`. Before a write, if any snapshot on main already carries that id, the call is a no-op.

  The legacy `wap.id` property is still read during the idempotency check for a migration window, so in-flight
  files written before this refactor aren't reprocessed after deploy.

  ## How concurrency is handled now

  `fast_append` (from `iceberg-rust 0.9`) already does what `publish_branch` was trying to replicate manually:

  - Each commit parents its new snapshot on **current main** (not a stale branch).
  - The commit request carries `RefSnapshotIdMatch(main == observed)` as an optimistic lock.
  - On 409, `Transaction::commit` reloads the table, rebuilds the manifest list over the new main, and retries.

  This makes the clobber scenario structurally impossible — every successful commit is a true descendant of
  main at the time it landed.

  ## Deleted

  - `helium_iceberg/src/branch.rs` (~520 lines)
  - `BranchTransaction`, `BranchWriter`, `BranchPublisher` traits and impls
  - `IcebergBranchWriter`, `IcebergBranchPublisher`, `EmptyIcebergBranchPublisher`
  - `WapState` + `detect_wap_state`, `has_wap_id`
  - `Error::CommitConflict` / `is_commit_conflict`
  - `TableDefinition::wap_enabled()` / `with_wap_enabled()` and the `write.wap.enabled` table property
  - The direct REST commit endpoint in `catalog/rest_endpoint.rs` (previously used only by `branch.rs`)

  ## Caller migration

  All callers moved to the new API (`mobile_verifier`, `mobile_packet_verifier`). `RewardTransactions` is gone;
   `RewardWriters` now has per-table `write_*(id, rows)` methods. `maybe_begin`/`maybe_publish` are replaced by
   a single `maybe_write_idempotent`. Sub-functions in `rewarder.rs` take `Option<(&RewardWriters, &str)>`
  instead of threading a mutable transaction.

  No `iot_*` or `ingest` callers touched the old API.

  ## Bug surfaced and fixed in `write_data_files`

  The initial concurrency test was weak — it cloned a single `Arc<dyn DataWriter<T>>` and ran on a
  single-threaded tokio runtime, which serialized the writes and never exercised `fast_append`'s conflict-retry
   path. Rewriting it with a multi-thread runtime, two independent `IcebergTable` instances, and a
  `tokio::sync::Barrier` immediately caught this:

  ```
  DataInvalid => Cannot add files that are already referenced by table,
  files: s3://.../data/inserted_day=2026-04-17/1776424975595-00000.parquet
  ```

  Both writers were deriving the parquet file-name prefix from `SystemTime::now().as_millis()`. Under real
  concurrency they landed on the same millisecond, produced identical paths, overwrote each other's S3 objects,
   and tripped `FastAppendAction::validate_duplicate_files` on the second commit.

  Fixed by switching the prefix to `uuid::Uuid::now_v7()` (time-ordered + random). This bug was latent with the
   old branch-based workflow too — no existing test would have caught it.

  ## New integration tests (`helium_iceberg/src/test_harness.rs`)

  Three tests, all running against the real Polaris + Trino + S3 harness:

  - **`concurrent_writes_dont_clobber`** — multi-thread runtime, two independent `IcebergTable` instances,
  barrier synchronized, 5 rounds. Validates that `fast_append`'s retry-on-conflict path actually composes
  writes end-to-end.
  - **`delete_between_writes_is_preserved`** — end-to-end reproduction of the original bug. Writes via
  `write_idempotent("a")`, runs `DELETE FROM ... WHERE name = 'Alice'` through Trino, writes again via
  `write_idempotent("b")`, asserts the delete is preserved.
  - **`write_idempotent_skips_duplicate_id_across_instances`** — drops and recreates the writer between the two
   idempotent calls so the second call has a fresh metadata cache and must hit the catalog to find the existing
   snapshot. Interleaves a plain `write` so `has_write_id` is forced to scan snapshot history rather than just
  the current tip.

  ## What reviewers should look at

  - **`helium_iceberg/src/iceberg_table.rs`** — the new `write_idempotent` implementation and `has_write_id`
  (with the legacy `wap.id` compat read).
  - **`mobile_verifier/src/rewarder.rs`** — the threading of `Option<(&RewardWriters, &str)>` through the
  reward sub-functions. Per-table idempotency is unchanged; cross-table atomicity is not provided (same as
  before).
  - **The new tests** — they're the primary safety net for the `fast_append` path.

  ## Caveats (documented in `iceberg_table.rs` module doc)

  1. Snapshot expiration invalidates idempotency. Anything that needs retention longer than Iceberg's snapshot
  history should track processed ids in Postgres.
  2. Cross-table atomicity is not provided (wasn't before either). Partial failure across multiple
  `write_idempotent` calls leaves some tables committed; retry completes the rest.
  3. Orphaned parquet files if concurrent `write_idempotent` calls race on the same id. Iceberg's orphan-file
  cleanup handles these.
  4. The legacy `wap.id` dual-read in `has_write_id` is a temporary migration shim. Safe to remove once all
  in-flight snapshots carrying the old property have expired.


helium_iceberg/src/catalog/rest_endpoint.rs
```diff
@@ -1,43 +1,23 @@
-//! Direct REST API access for the Iceberg catalog.
+//! Catalog endpoint configuration.
```
michaeldjeffrey: Should we roll whatever is left of this file into the `auth` module?

helium_iceberg/src/iceberg_table.rs
```diff
@@ -43,34 +72,35 @@ impl<T> IcebergTable<T> {
...
-            .snapshot_for_ref(MAIN_BRANCH)
-            .map(|s| s.summary().additional_properties.clone())
-            .unwrap_or_default()
-    }
-
-    async fn write_and_commit(&self, batch: RecordBatch) -> Result {
+    async fn write_and_commit(
+        &self,
+        batch: RecordBatch,
+        snapshot_properties: HashMap<String, String>,
```
michaeldjeffrey, line 78: Wondering if we want to make this more explicitly an optional write_id. We're kind of enforcing that on the caller through `write_idempotent`. My thinking behind this is let's start narrow with what we allow so we don't abuse a grab bag of settings that confuse us down the road one day.

helium_iceberg/src/iceberg_table.rs
```diff
@@ -94,129 +124,27 @@ impl<T: Serialize + Send + Sync + 'static> DataWriter<T> for IcebergTable<T> {
         let table = self.table.read().await;
         let batch = records_to_batch(&table, &records)?;
         drop(table);
-        self.write_and_commit(batch).await
+        self.write_and_commit(batch, HashMap::new()).await
     }
 
-    async fn begin(&self, wap_id: &str) -> Result<BranchTransaction<T>> {
+    async fn write_idempotent(&self, id: &str, records: Vec<T>) -> Result {
```
michaeldjeffrey, line 130: It might made sense to take `id: impl Into<String>` here? We could pass a bare timestamp `i64`...

helium_iceberg/src/iceberg_table.rs
```diff
@@ -318,44 +246,11 @@ fn build_writer_properties(
...
-        )));
-    }
-
-    Ok(meta.snapshots().any(|snapshot| {
-        snapshot
-            .summary()
-            .additional_properties
-            .get(crate::branch::WAP_ID_KEY)
-            .is_some_and(|v| v == wap_id)
+fn has_write_id(table: &Table, id: &str) -> Result<bool> {
```
michaeldjeffrey: This function cannot fail

helium_iceberg/src/table_creator.rs
```diff
@@ -718,16 +718,6 @@ impl TableDefinitionBuilder {
         )
     }
 
-    /// Enable Write-Audit-Publish (WAP) for this table.
-    pub fn wap_enabled(self) -> Self {
```
michaeldjeffrey, line 722: Not a fully fledged idea yet.

We could mark a table definition as only accepting idempotent writes, then the writer you get back gives you a write method that takes the `write_idempotent` path always so you can't accidentally call the wrong thing.

helium_iceberg/src/test_harness.rs
```diff
@@ -949,6 +949,213 @@ mod tests {
...
+    /// call) so its metadata cache is empty and `has_write_id` must hit the
+    /// catalog to find the existing snapshot. This is the production scenario
+    /// the idempotency guarantee is actually protecting against — a crash
+    /// between submit and commit-recording, followed by restart and retry.
+    ///
+    /// Also verifies an interleaved plain `write` between the two idempotent
+    /// calls doesn't break the check — `has_write_id` scans all snapshots, so
+    /// the duplicate id must still be found even when it's no longer the tip.
+    #[tokio::test]
+    async fn write_idempotent_skips_duplicate_id_across_instances() -> anyhow::Result<()> {
```
michaeldjeffrey: We're starting to validate the properties of the library, maybe it's time to move these tests out of the test_harness file and into a proper `tests` directory.

mobile_packet_verifier/src/daemon.rs
```diff
@@ -134,17 +133,18 @@ where
...
 
-        iceberg::maybe_publish(iceberg_txn).await?;
         transaction.commit().await?;
         self.verified_data_session_report_sink.commit().await?;
 
         Ok(())
     }
 }
 
+#[allow(clippy::too_many_arguments)]
```
michaeldjeffrey:
```suggestion
#[expect(clippy::too_many_arguments)]
```

michaeldjeffrey (commented):
- `mobile_packet_verifier/src/burner.rs` was missed

helium_iceberg/src/iceberg_table.rs
bbalser: good catch updated 85a661a

michaeldjeffrey (approved):

helium_iceberg/src/catalog/rest_endpoint.rs
bbalser: direct endpoint was no longer needed c5b61d0

helium_iceberg/src/test_harness.rs
bbalser: moved c4cf26c

PR #1178, opened 2026-04-23
Speedtest iceberg

michaeldjeffrey:
Start writing speedtests to iceberg.
Along with adding the backfill machinery for verified speedtests.

I went back and forth on the standalone backfill commands. Thought we could maybe use it to test out the fast append stuff.

Backfill specific settings are separate from iceberg settings, and they will be separate for upcoming backfills of other types so they can be enabled/disabled independently.

The difference in construction between the `PocWriters` and the `RewardWriters` seemed a bit odd to me at first, but what I've been telling myself is that the `PocWriters` are independent of one another and the `RewardWriters` work as a unit. So there's a difference between each field being optional versus the entire struct being optional. 

I made each backfill process_name overrideable in case we want to change the settings and start from a new place without needing to make a new release. 

### Questions
I'm starting to notice that we have some differences in how we treat public keys. For example.
|||
|---|---|
| `speedtest` | `hotspot_pubkey` |
|`heartbeat`|`hotspot_pubkey`|
| `radio_reward_covered_hex` | `hotspot_key`|
|`radio_reward`|`hotspot_key`|
|`gateway_reward`|`hotspot_key`|
|`service_provider_reward`|`rewardable_entity_key`|

Do we want to try and consolidate some of these...

(service provider reward is an outlier for many reasons, I'm mostly going after the suffix of `key` vs `pubkey` vs `pub_key` (that could potentially sneak it's way in if we're no vigilant)).

### DIFF
--- a/mobile_verifier/src/backfill.rs
+++ b/mobile_verifier/src/backfill.rs
@@ -1,0 +1,219 @@
+use anyhow::Context;
+use chrono::{DateTime, Utc};
+use file_store::{
+    file_info::FileInfo, file_info_poller::FileInfoStream, file_source, traits::MsgDecode,
+    BucketClient,
+};
+use file_store_oracles::FileType;
+use futures::StreamExt;
+use helium_iceberg::BoxedDataWriter;
+use sqlx::PgPool;
+use std::time::Duration;
+use task_manager::ManagedTask;
+use tokio::sync::mpsc::Receiver;
+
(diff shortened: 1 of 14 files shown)


michaeldjeffrey:
CI should be fixed in this PR with https://github.com/helium/oracles/pull/1179

bbalser (approved):

PR #1179, opened 2026-04-23
`FileUpload` test race

michaeldjeffrey:
There’s a chance for some of these test that we trigger a shutdown of
the daemon after a file has been processed, but before FileUpload was
able to fully upload it to s3. 

Given our pattern is to look for orphaned files and try to upload them
on startup, I didn’t want to change the contract that much for currently
running services by making FileUpload wait until it’s upload channel was
complete before shutting down. So we’ve added a small counter where you
can ask it how many files have been uploaded. In a test situation that
should be a predictable thing, if there’s a case where it’s not, I can't
predict that right now and whoever runs into that will need to come up
with a better solution.

### DIFF
--- a/file_store/src/file_upload.rs
+++ b/file_store/src/file_upload.rs
@@ -5,7 +5,11 @@ use std::{
     time::Duration,
 };
 use task_manager::ManagedTask;
-use tokio::{fs, sync::mpsc, time};
+use tokio::{
+    fs,
+    sync::{mpsc, watch},
+    time,
+};
 use tokio_stream::wrappers::UnboundedReceiverStream;
 
 pub type MessageSender = mpsc::UnboundedSender<PathBuf>;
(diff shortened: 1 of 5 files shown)


file_store/src/file_upload.rs
```diff
@@ -67,22 +104,26 @@ impl FileUploadServer {
...
                 tracing::warn!("ignoring non file {path_str}");
+                completion_tx.send_modify(|n| *n += 1);
                 return;
             }
             let mut retry = 0;
             const MAX_RETRIES: u8 = 5;
             const RETRY_WAIT: Duration = Duration::from_secs(10);
             while retry <= MAX_RETRIES {
                 tracing::debug!("storing {path_str} in {bucket} retry {retry}");
+                // tokio::time::sleep(std::time::Duration::from_secs(11)).await;
```
bbalser: did you intend for this comment to stay?

bbalser (approved):

PR #1180, opened 2026-04-24
Add dump comands for Mobile and Iot Change Reports

michaeldjeffrey:
not strictly necessary, but nice to have the coverage for when you want to dump a file you haven’t before.


bbalser:
these reports are in trino already

bbalser (approved):

PR #1192, opened 2026-05-18
Add `trino_client` crate: typed Settings, parameterized SQL via EXECUTE IMMEDIATE

bbalser:
New workspace crate at `trino_client/` that wraps `trino-rust-client` with:

- **`Settings`**: typed, serde-deserializable config — `host`, `port`, `user`, `catalog`, `schema`, `secure`, `ca_cert_path`, `insecure_skip_tls_verify`, plus `auth: Option<AuthSettings>` with `Basic { username, password }` and `Jwt { token }` variants.
- **`Statement`**: parameterized SQL builder — `Statement::new(sql).bind(name, value).render() -> Result<String>`. Named placeholders (`:id`, `:name`) render to Trino's `EXECUTE IMMEDIATE 'sql with ?' USING v1, v2, ...` form. Scanner is quote- and comment-aware (skips `:` inside `'...'`, `"..."`, `-- ...`, `/* ... */`).
- **`Param`** enum with `From` impls for: `i32`, `i64`, `u32`, `u64`, `f32`, `f64`, `bool`, `&str`, `String`, `Vec<u8>`, `&[u8]`, `chrono::NaiveDate`, `chrono::NaiveDateTime`, `chrono::DateTime<Utc>`, and a blanket `Option<T: Into<Param>>` that renders `None` as bare `NULL`. Bytes render as Trino `VARBINARY` hex literals (`X'deadbeef'`). `u64 > i64::MAX` errors at render time.
- **`SqlStatement` / `SqlQuery` traits**: a struct implements `SqlStatement` (returns a `Statement`); add `SqlQuery: SqlStatement` with `type Row` for typed row-returning queries. `Client::execute(&q)` and `Client::get_all(&q)` dispatch through these.
- **`Client`**: owns the underlying `trino_rust_client::Client`. Exposes `from_settings`, `from_inner`, `inner`, trait-based `execute`/`get_all`, plus raw `execute_raw` / `get_all_raw` for ad-hoc SQL (and for DDL, since Trino's `EXECUTE IMMEDIATE` rejects DDL).
- Re-exports `trino_rust_client` so downstream callers using `#[derive(Trino)]` and `error::Error::EmptyData` only need one dep.

Migration of existing callers (`helium_iceberg`, `mobile_verifier`, `mobile_packet_verifier`) is intentionally **out of scope** for this PR — most of those sites interpolate identifiers (`{NAMESPACE}.{TABLE_NAME}`) which the parameterized API can't bind, so migration is a separate, mechanical follow-up.

## Test plan

54 tests pass (44 unit + 10 integration):

- [x] `cargo build -p trino_client --all-targets`
- [x] `cargo clippy -p trino_client --all-targets -- -D warnings`
- [x] Unit tests cover: literal rendering for every `Param` variant, quote-/comment-aware scanner edge cases (`''`, `""`, `:` inside strings/identifiers/comments), missing-param / no-param / NaN / Inf / u64-overflow errors, `Option` blanket, `Settings` serde deserialization for basic auth, JWT auth, TLS options.
- [x] Integration tests against local Trino (`localhost:8080`, user `admin`): scalar round-trip, date round-trip, naive-/UTC-timestamp equality vs `TIMESTAMP` literals, apostrophe escaping round-trip, `SqlStatement::execute` path, `SqlQuery::get_all` path with typed `Row`, `Option<None>` → `IS NULL`, `Option<Some>` → value, `u64` → `BIGINT`, `Vec<u8>` → `X'...'` `VARBINARY`.

### DIFF
--- a/trino_client/src/builder.rs
+++ b/trino_client/src/builder.rs
@@ -1,0 +1,129 @@
+use crate::error::Result;
+use crate::settings::{AuthSettings, Settings};
+use crate::Client;
+use std::path::PathBuf;
+
+pub struct ClientBuilder {
+    host: String,
+    port: u16,
+    user: String,
+    catalog: Option<String>,
+    schema: Option<String>,
+    secure: bool,
+    ca_cert_path: Option<PathBuf>,
+    insecure_skip_tls_verify: bool,
+    auth: Option<AuthSettings>,
+}
+
+impl ClientBuilder {
+    pub fn new(host: impl Into<String>, port: u16, user: impl Into<String>) -> Self {
+        Self {
+            host: host.into(),
+            port,
+            user: user.into(),
+            catalog: None,
+            schema: None,
+            secure: false,
+            ca_cert_path: None,
+            insecure_skip_tls_verify: false,
+            auth: None,
+        }
+    }
+
+    pub fn catalog(mut self, catalog: impl Into<String>) -> Self {
+        self.catalog = Some(catalog.into());
+        self
+    }
+
+    pub fn schema(mut self, schema: impl Into<String>) -> Self {
+        self.schema = Some(schema.into());
+        self
+    }
+
+    pub fn secure(mut self, secure: bool) -> Self {
+        self.secure = secure;
+        self
+    }
+
+    pub fn ca_cert_path(mut self, path: impl Into<PathBuf>) -> Self {
+        self.ca_cert_path = Some(path.into());
+        self
+    }
+
+    pub fn insecure_skip_tls_verify(mut self, skip: bool) -> Self {
+        self.insecure_skip_tls_verify = skip;
+        self
+    }
+
+    pub fn basic_auth(mut self, username: impl Into<String>, password: impl Into<String>) -> Self {
+        self.auth = Some(AuthSettings::Basic {
+            username: username.into(),
+            password: Some(password.into()),
+        });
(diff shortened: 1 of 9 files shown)


trino_client/src/builder.rs
```diff
@@ -0,0 +1,129 @@
...
+use crate::Client;
+use std::path::PathBuf;
+
+pub struct ClientBuilder {
+    host: String,
+    port: u16,
+    user: String,
+    catalog: Option<String>,
+    schema: Option<String>,
+    secure: bool,
```
michaeldjeffrey, line 11: having `scheme` and `secure` seems a bit redundant.

trino_client/src/builder.rs
```diff
@@ -0,0 +1,129 @@
...
+    ca_cert_path: Option<PathBuf>,
```
michaeldjeffrey: Do we use, or are we planning on using ca_certs?

michaeldjeffrey (approved):

trino_client/src/builder.rs
bbalser: that is `schema` not `scheme`

trino_client/src/builder.rs
bbalser: I don't believe we do, removed 91b745d

PR #1196, opened 2026-05-19
update test-harness and all users of it to use trino-client

bbalser:


helium_iceberg/Cargo.toml
```diff
@@ -43,6 +43,7 @@ triggered = { workspace = true }
 
 # Test harness dependencies (optional)
 trino-rust-client = { version = "0.9", optional = true }
+trino-client = { path = "../trino_client" }
```
michaeldjeffrey, line 46: Should this be optional and feature gated as well?

mobile_packet_verifier/Cargo.toml
```diff
@@ -47,7 +47,8 @@ task-manager = { path = "../task_manager" }
 tls-init = { path = "../tls_init" }
 
 helium-iceberg = { path = "../helium_iceberg", features = ["test-harness"] }
-trino-rust-client = { version = "0.9" }
+trino-client = { path = "../trino_client" }
+trino-rust-client = "0.9"
```
michaeldjeffrey, line 51: Should we loosen this to `"*"` so it always picks up the version from `trino-client`?

mobile_verifier/src/iceberg/speedtest_avg.rs
```diff
@@ -62,15 +62,13 @@ pub fn table_definition() -> helium_iceberg::Result<TableDefinition> {
...
+pub async fn get_all(trino: &trino_client::Client) -> anyhow::Result<Vec<IcebergSpeedtestAvg>> {
     let all = match trino
-        .get_all(format!("SELECT * from {NAMESPACE}.{TABLE_NAME}"))
+        .get_all_raw(format!("SELECT * from {NAMESPACE}.{TABLE_NAME}"))
         .await
     {
-        Ok(all) => all.into_vec(),
-        Err(trino_rust_client::error::Error::EmptyData) => vec![],
+        Ok(all) => all,
+        Err(trino_client::Error::EmptyData) => vec![],
```
michaeldjeffrey, line 71: I'm trying to think of why it would be a bad idea to handle `EmptyData` in our wrapper around `get_all_raw`.

PR #1198, opened 2026-05-20
Add JWT file watcher to `trino_client`

macpie:
## Summary

JWTs rotate in production; without a refresh mechanism, the inner Trino client (and its containing service) would need to restart on each rotation. This ports the `JwtWatcher` pattern from glacon-rs into the `trino_client` crate.

- New `AuthSettings::JwtFile { path, refresh_interval }` variant — polls the token file via `notify::PollWatcher` and rebuilds the inner client on change.
- `Client` now holds a `watch::Receiver<Arc<...>>`, so token refreshes swap the inner client behind a single `Client` handle used by all callers.
- `Client::watch_jwt(shutdown)` exposes the watcher future. Static `Jwt`/`Basic`/anonymous auth keep working unchanged.
- Opt-in `task-manager` cargo feature provides a `ManagedTask` impl for `Client`, mirroring glacon-rs's `add_task(client.clone())` pattern.

### DIFF
--- a/trino_client/src/error.rs
+++ b/trino_client/src/error.rs
@@ -15,9 +15,6 @@ pub enum Error {
     #[error("notify error: {0}")]
     Watcher(#[from] notify::Error),
 
-    #[error("no tokio runtime available to spawn jwt watcher")]
-    NoTokioRuntime,
-
     #[error("watch channel send: {0}")]
     WatchSend(String),
 
--- a/trino_client/src/error.rs
+++ b/trino_client/src/error.rs
@@ -15,8 +15,8 @@ pub enum Error {
     #[error("notify error: {0}")]
     Watcher(#[from] notify::Error),
 
-    #[error("watch_jwt requires AuthSettings::JwtFile")]
-    JwtWatchUnsupported,
+    #[error("no tokio runtime available to spawn jwt watcher")]
+    NoTokioRuntime,
 
     #[error("watch channel send: {0}")]
     WatchSend(String),
--- a/trino_client/src/error.rs
+++ b/trino_client/src/error.rs
@@ -1,8 +1,5 @@
 #[derive(Debug, thiserror::Error)]
 pub enum Error {
-    #[error("failed to build trino client: {0}")]
-    Build(String),
-
     #[error("missing bind parameter: {0}")]
     MissingParameter(String),
 
@@ -12,11 +9,8 @@ pub enum Error {
     #[error("io error: {0}")]
     Io(#[from] std::io::Error),
 
-    #[error("notify error: {0}")]
-    Watcher(#[from] notify::Error),
-
-    #[error("watch channel send: {0}")]
-    WatchSend(String),
+    #[error(transparent)]
+    JwtWatcher(#[from] crate::jwt_watcher::JwtWatcherError),
 
     #[error(transparent)]
     Client(#[from] trino_rust_client::error::Error),
--- a/trino_client/src/error.rs
+++ b/trino_client/src/error.rs
@@ -9,6 +9,18 @@ pub enum Error {
     #[error("invalid parameter value: {0}")]
     InvalidParam(String),
 
+    #[error("io error: {0}")]
+    Io(#[from] std::io::Error),
+
+    #[error("notify error: {0}")]
+    Watcher(#[from] notify::Error),
+
+    #[error("watch_jwt requires AuthSettings::JwtFile")]
+    JwtWatchUnsupported,
+
+    #[error("watch channel send: {0}")]
+    WatchSend(String),
+
     #[error(transparent)]
     Client(#[from] trino_rust_client::error::Error),
 }
--- a/trino_client/src/jwt_watcher.rs
+++ b/trino_client/src/jwt_watcher.rs
@@ -70,8 +70,14 @@ impl notify::EventHandler for JwtWatcher {
             return;
         }
 
-        if let Err(err) = self.create_and_send_new_inner() {
-            tracing::error!(?err, "failed to refresh trino client");
+        match self.create_and_send_new_inner() {
+            Ok(()) => {}
+            Err(Error::WatchSend(_)) => {
+                // All receivers have been dropped — we're on the way down.
+                // Benign during shutdown; loud at `error!` would be noise.
+                tracing::debug!("jwt refresh skipped: no receivers");
+            }
+            Err(err) => tracing::error!(?err, "failed to refresh trino client"),
         }
     }
 }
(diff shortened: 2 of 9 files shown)


trino_client/src/lib.rs
```diff
@@ -32,46 +39,88 @@ impl<T: SqlQuery + ?Sized> SqlQuery for &T {
...
+        let (updater, inner_rx) = watch::channel(initial);
+
+        let watcher_task = match &settings.auth {
+            Some(AuthSettings::JwtFile {
+                path,
+                refresh_interval,
+            }) => {
+                // Refuse to silently no-op outside a runtime.
+                let rt =
+                    tokio::runtime::Handle::try_current().map_err(|_| Error::NoTokioRuntime)?;
```
bbalser: ### `watcher_task` is a `pending()` future — the spawn isn't load-bearing

In `from_settings`:

```rust
Some(rt.spawn(async move {
    let _watcher = poll_watcher;
    std::future::pending::<()>().await;
}))
```
Claude seems to believe this will simplify some things


The spawned task does no async work — it exists solely to own `PollWatcher` and be abortable from `Drop`. After checking `notify` 6.1.1's `src/poll.rs`:

- `PollWatcher: Send + Sync` (their own test asserts this)
- The polling thread is **detached** — the internal `JoinHandle` is dropped at spawn time
- `PollWatcher::drop` is non-blocking: it just sets an `Arc<AtomicBool>` flag. The polling thread observes it on its next tick and exits on its own

So the usual reason to offload a destructor onto a tokio task (avoid blocking the executor) doesn't apply here. The spawn is dead weight, and it costs:

- A `Handle::try_current()` gate, which is the entire reason `Error::NoTokioRuntime` and the `from_settings_errors_without_tokio_runtime` test exist
- A custom `Drop` on `ClientInner` whose only job is `handle.abort()`
- An `Option<JoinHandle<()>>` field that duplicates information already in `settings.auth`
- A `pending().await` that every future reader has to stop and decode

Suggested simplification — store the watcher directly:

```rust
struct ClientInner {
    inner_rx: watch::Receiver<InnerClient>,
    settings: Settings,
    _watcher: Option<PollWatcher>,
}
```

No `Drop` impl, no `JoinHandle`, no `pending()`. `from_settings` then works from any context (sync `main`, `current_thread` runtime, test harnesses) rather than only inside a multi-thread runtime. `Error::NoTokioRuntime` and its test delete cleanly.

Minor caveat either way: the polling thread can fire one final `handle_event` after `want_to_stop` is set but before it observes the flag. If the receivers are already gone, `updater.send(...)` returns `Err` and logs `"failed to refresh trino client"` at `error!`. Worth downgrading that specific send-failure to `debug!` so normal shutdown doesn't generate spurious error logs.

macpie: Good catch!

bbalser (approved):

trino_client/src/jwt_watcher.rs
```diff
@@ -0,0 +1,170 @@
...
+    /// Construct and start the `PollWatcher`. The returned handle owns the
+    /// internal polling thread; keep it alive for as long as you want refresh
+    /// events to be processed.
+    pub(crate) fn start(
+        self,
+        token_path: &Path,
+        refresh_interval: Duration,
+    ) -> Result<PollWatcher> {
+        let config = Config::default()
+            .with_poll_interval(refresh_interval)
```
michaeldjeffrey, line 53: We may need to update our version or `notify` from `6.1.1` to the version used in glacon-rs `8.2.0` to get back `.with_follow_symlinks(true)`.

I remember there being some weirdness around how mounts are handled in k8s wrt how the token files are swapped out.

trino_client/src/jwt_watcher.rs
```diff
@@ -0,0 +1,170 @@
...
+        // Constructor is fully synchronous: no runtime is required.
+        let dir = tempfile::tempdir().unwrap();
+        let path = dir.path().join("token.jwt");
+        std::fs::write(&path, b"some-token").unwrap();
+        let settings = base_settings(Some(AuthSettings::JwtFile {
+            path,
+            refresh_interval: Duration::from_millis(50),
+        }));
+        let client = Client::from_settings(&settings).expect("constructor must not require tokio");
+        drop(client);
```
michaeldjeffrey: This drop is weird in a test.
And do we care to make running outside a tokio context part of our api?

trino_client/src/jwt_watcher.rs
```diff
@@ -0,0 +1,170 @@
...
+    }
+
+    #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
```
michaeldjeffrey: Are we testing something specific that requires the extra args to `#[tokio::test]`? 
This still passes without declaring `flavor` or `worker_threads`.

trino_client/src/lib.rs
```diff
@@ -32,46 +39,65 @@ impl<T: SqlQuery + ?Sized> SqlQuery for &T {
...
+    /// caching the returned `Arc` long-term.
+    pub fn inner(&self) -> InnerClient {
+        self.inner.inner_rx.borrow().clone()
     }
 
-    pub fn inner(&self) -> &trino_rust_client::Client {
-        &self.inner
+    pub fn settings(&self) -> &Settings {
+        &self.inner.settings
     }
```
michaeldjeffrey: If this isn't going to be used anywhere, we can drop this handle on `Settings`.

trino_client/src/lib.rs
```diff
@@ -32,46 +39,65 @@ impl<T: SqlQuery + ?Sized> SqlQuery for &T {
...
+
+        let watcher = match &settings.auth {
+            Some(AuthSettings::JwtFile {
+                path,
+                refresh_interval,
+            }) => Some(JwtWatcher::new(settings.clone(), updater)?.start(path, *refresh_interval)?),
+            _ => {
+                // No watcher: the receiver still holds the last value even
+                // after the sender is dropped.
+                drop(updater);
```
michaeldjeffrey: `updater` will already be dropped when this function exits and it wasn't given away to anyone.

trino_client/src/lib.rs
```diff
@@ -32,46 +39,65 @@ impl<T: SqlQuery + ?Sized> SqlQuery for &T {
...
+            }),
+        })
     }
 
-    pub fn from_inner(inner: trino_rust_client::Client) -> Self {
-        Self { inner }
+    /// Returns the current inner client. JWT refreshes swap the value behind
+    /// this `Arc`, so callers should call `inner()` per request rather than
+    /// caching the returned `Arc` long-term.
+    pub fn inner(&self) -> InnerClient {
```
michaeldjeffrey: This function should probably not be public. We want to force queries to go through the channel borrow.

trino_client/src/error.rs
```diff
@@ -9,6 +9,15 @@ pub enum Error {
     #[error("invalid parameter value: {0}")]
     InvalidParam(String),
 
+    #[error("io error: {0}")]
+    Io(#[from] std::io::Error),
```
michaeldjeffrey, line 10: Given the `JwtWatcher` is fairly self contained, it would be nice to have another level of error, something like `JwtWatcherError`. 

That would keep any potential future `std::io::Error` variants from bubbling up without context.
And differentiate `trino_rust_client::error::Error`s between building a client, and query errors.

trino_client/src/lib.rs
```diff
@@ -11,9 +12,15 @@ pub use error::{Error, Result};
...
 pub use statement::{Param, Statement, TypedStatement};
 
+use jwt_watcher::JwtWatcher;
+use notify::PollWatcher;
+use std::sync::Arc;
+use tokio::sync::watch;
 use trino_rust_client::auth::Auth;
 use trino_rust_client::ClientBuilder as UpstreamClientBuilder;
 
+pub type InnerClient = Arc<trino_rust_client::Client>;
```
michaeldjeffrey: Can we come up with a different name for this alias?

- `Client`
- `InnerClient`
- `ClientInner`

Are a bit confusing to read through in this workspace.
Best guess I have so far is `TrinoClient` or `WrappedTrinoClient`. We shouldn't really be depending on that type outside of this crate.

michaeldjeffrey (approved):

PR #1199, opened 2026-05-22
Add mobile_verifier_compare crate: PG ↔ Trino reward-query parity CLI

macpie:
## Summary

New standalone binary crate `mobile-verifier-compare` that diffs every Postgres reward-input query `mobile_verifier` runs against an equivalent query over the matching Iceberg tables in Trino (`poc.heartbeats`, `poc.speedtests`, `rewards.data_transfer`), then prints a side-by-side table with per-row `MATCH` / `MISMATCH` / `PG_ONLY` / `TRINO_ONLY` status.

Lives in its own workspace member — `mobile_verifier` itself is untouched. The prod/dev boundary is enforced by the dep graph rather than a Cargo feature flag, so release builds of `mobile_verifier` contain no Trino client code by construction.

## Why

We want to verify that the Iceberg-side data the new backfillers and writers produce is byte-equivalent to what the Postgres-side reward pipeline reads, before we trust downstream consumers to use Iceberg instead of PG. This is the validation harness for that work.

### Diff subcommands
Each subcommand calls the **production** PG function so the Postgres SQL cannot drift from what the reward pipeline actually executes:

| Subcommand | PG function (reused via path dep) | Trino side |
|---|---|---|
| `heartbeats` | `HeartbeatReward::validated` (`valid_radios.sql`) | mirrored CTE over `poc.heartbeats` with `count(*) >= 12` filter |
| `speedtests` | `aggregate_epoch_speedtests` | window-function translation over `poc.speedtests` |
| `latest-speedtests` | `get_latest_speedtests_for_pubkey` | per-hotspot lookup |
| `data-sessions` | `aggregate_hotspot_data_sessions_to_dc` | aggregate `rewards.data_transfer` over the window |
| `payer-burns` | `sum_data_sessions_to_dc_by_payer` | rolled-up `rewards.data_transfer` (header notes Iceberg has no payer column) |
| `pending-burns` | `mobile_packet_verifier::pending_burns::initialize` SQL | n/a — pending is by definition not in Iceberg |
| `all` | runs every comparison sequentially | |

The diff table shows mismatches by default (`--show-all` to dump matches too) and prints totals + max delta.

### `seed` subcommand (zero-config bootstrap)
Brings the local docker-compose stack from nothing to a runnable diff in one command:

1. Drops + creates `mobile_verifier` and `mobile_packet_verifier` Postgres databases on the docker-compose PG.
2. Runs `sqlx` migrations against both (paths resolved relative to `CARGO_MANIFEST_DIR`).
3. `CREATE SCHEMA` + `CREATE TABLE` for `iceberg.poc.heartbeats`, `iceberg.poc.speedtests`, `iceberg.rewards.data_transfer` via Trino DDL.
4. Populates deterministic fixtures across four buckets (60% match / 20% mismatch / 10% pg-only / 10% trino-only) so every diff status appears on the very first run.
5. Auto-writes `mobile_verifier_compare/pkg/compare-trino-settings.local.toml` (gitignored) with docker-compose defaults, and prints the exact next `cargo run … all` command to copy.

Takes every endpoint as a CLI flag with docker-compose-friendly defaults — works with no `-c` config file.

## Reviewer notes

- The data-sessions / payer-burns comparisons diff PG **input** sessions against `rewards.data_transfer` **output** rewards — the printed table headers call out that this is an input-vs-output sanity check, not strict row equality. A future raw-sessions Iceberg writer would let us swap that SQL.
- `payer-burns` rolls Iceberg to a single `__all__` bucket because `rewards.data_transfer` has no payer column. Per-payer PG rows are shown for visibility but always tag as `PG_ONLY` — by design today.
- `pending-burns` is one-sided (no Iceberg counterpart for unprocessed sessions); reported as `PG_ONLY` with a header note.


mobile_verifier_compare/src/speedtests.rs
```diff
@@ -0,0 +1,252 @@
...
+use sqlx::{Pool, Postgres};
+use std::collections::BTreeMap;
+use trino_client::{Client as TrinoClient, Statement};
+use trino_rust_client::Trino;
+
+// Mirrors `crate::speedtests::SPEEDTEST_LAPSE` and `SPEEDTEST_AVG_MAX_DATA_POINTS`.
+// Those constants are not pub today; matching them by value keeps the Trino
+// query in lockstep with production. If they ever diverge, update both here
+// and in `speedtests.rs`.
+const SPEEDTEST_LAPSE_HOURS: i64 = 24;
```
bbalser: shouldn't this be 48 hours?

macpie: fixed by important prod const

mobile_verifier_compare/src/heartbeats.rs
```diff
@@ -0,0 +1,151 @@
...
+    WITH hourly AS (
+        SELECT
+            hotspot_pubkey,
+            device_type,
+            date_trunc('hour', heartbeat_timestamp) AS hour,
+            avg(location_trust_score_multiplier) AS trust
+        FROM poc.heartbeats
+        WHERE heartbeat_timestamp >= :start
+        AND heartbeat_timestamp < :end
+        GROUP BY hotspot_pubkey, device_type, date_trunc('hour', heartbeat_timestamp)
```
bbalser, line 61: This does not calculate the average location trust score the same way, this is averaging all heartbeats in a give hour, currently we average the mutlipliers from the first heartbeat each hour

mobile_verifier_compare/src/data_sessions.rs
```diff
@@ -0,0 +1,111 @@
...
+    hotspot_key: String,
+    rewardable_bytes: i64,
+    dc_transfer_reward: i64,
+}
+
+/// Trino-side: aggregate `rewards.data_transfer` rows whose `[start_period, end_period)`
+/// overlaps the requested window. This is the *output* of the production reward
+/// pipeline, not raw sessions, so this comparison is an input-vs-output sanity
+/// check: PG numbers should be >= Trino numbers in a normal epoch (PG includes
+/// all sessions, Trino only includes those that got rewards written).
```
bbalser: I am not sure I understand what this is validating

PR #1201, opened 2026-05-28
mobile-config: drive gateway updates from chain_rewardable_entities streams

bbalser:
## Summary

- Add two `file_info_poller` consumers in `mobile-config` for the `mobile_hotspot_change_report` and `entity_ownership_change_report` proto streams written to S3 by the ingest service, replacing the per-event portion of the metadata-DB polling tracker.
- Keep the existing `Tracker` running as a daily reconciliation pass (interval bumped from 1 hour to 1 day) to backfill `antenna`/`elevation` drift and any events the stream missed.
- For each `mobile_hotspot_change_v1`, look up `(antenna, elevation)` from `mobile_hotspot_infos.deployment_info` in the metadata DB at event time — these two fields aren't carried by the proto but everything else (`device_type`, `asserted_hex`, `azimuth`, owner via the separate stream) is.

## Why

The chain indexer now publishes change events directly. Consuming them lets us:
- Drop the cost of polling the full `mobile_hotspot_infos` join hourly.
- Bring update latency from "up to an hour" to "next report file."
- Stop depending on metadata-DB reads for the hot path — the daily tracker is now the only metadata-DB consumer here, with the stream consumers doing one indexed single-row lookup per event for the two missing fields.

## What changed

- `mobile_config/src/gateway/hotspot_change_stream.rs` *(new)* — `MobileHotspotChange` domain wrapper + `MsgDecode` impl + `HotspotChangeDaemon`. Skips CBRS, reconstructs `num_location_asserts` locally (increment when `asserted_hex` differs from prior), carries forward owner/`owner_changed_at` from the previous row, and uses change-detection via the harmonized hash.
- `mobile_config/src/gateway/ownership_change_stream.rs` *(new)* — `OwnershipChange` wrapper + `OwnershipChangeDaemon`. Updates `owner`/`owner_changed_at` for known gateways only; events for unknown entities are dropped (reconciled by the daily tracker).
- `mobile_config/src/gateway/db.rs` — `Gateway::compute_hash` moved here and harmonized to the narrower input set (`location`, `location_asserts`, `gateway_type`, `antenna`, `elevation`, `azimuth`) so tracker and streams agree. `Gateway::insert` and `Gateway::insert_bulk` now bind the hash directly from `compute_hash()` — callers can no longer accidentally insert an empty hash. `Gateway::insert` generalized to any `PgExecutor` so daemons can write through a transaction alongside `files_processed`.
- `mobile_config/src/gateway/metadata_db.rs` — new `MobileHotspotInfo::fetch_antenna_elevation` for the single-row antenna/elevation backstop. Old wide `compute_hash` removed; `is_active`/`is_full_hotspot`/`dc_onboarding_fee_paid` kept on the struct (now `#[allow(dead_code)]`) for parity with the metadata DB schema, slated for cleanup once the stream path is proven.
- `mobile_config/src/gateway/tracker.rs` — uses `gw.compute_hash()` for change detection so it agrees with the stream-written rows.
- `mobile_config/src/settings.rs` — adds `ingest: file_store::BucketSettings`, `gateway_stream_start_after: DateTime<Utc>`, and changes default `gateway_tracker_interval` from 1 hour → 1 day.
- `mobile_config/src/cli/server.rs` — wires both pollers + daemons into the task manager alongside the existing grpc server and (now-daily) Tracker.
- `mobile_config/migrations/20260526000000_files_processed.sql` *(new)* — matches the schema used elsewhere in the repo (mobile_verifier). Idempotent (`CREATE TABLE IF NOT EXISTS`).

## What the stream does NOT carry (and how we handle it)

| Field | Handling |
|---|---|
| `antenna`, `elevation` | Looked up per-event from `mobile_hotspot_infos.deployment_info`. If the metadata DB doesn't have the row yet, nulls are stored and the daily tracker backfills. |
| `num_location_asserts` | Reconstructed locally: increment when incoming `asserted_hex` differs from prior `location`. |
| `is_active`, `is_full_hotspot`, `dc_onboarding_fee_paid` | Not exposed by any `gateway_info_*` gRPC handler today. Dropped from the change-detection hash. |
| `owner` (and `owner_changed_at`) | Comes from the separate `entity_ownership_change_report` stream. |

Searched the rest of the repo to confirm no other consumer reads the three dropped fields from `gateways` — see prior conversation. Only `elevation` outside `mobile-config` lives in `file_store_oracles/src/iot/*` (IoT PoC beacon, unrelated).

## Notes for reviewers

- The plan and tradeoffs (including the "accept staleness" decision for antenna/elevation) are at [`.claude/plans/lets-accept-the-staleness-elegant-dream.md`](.claude/plans/lets-accept-the-staleness-elegant-dream.md) at this commit.
- After the cutover, the first daily Tracker run will rewrite hashes on every existing gateway because the old wide hash and the new narrow hash disagree by definition. Expect a one-time bulk insert of history rows on the first daily tick post-deploy.
- The daemon writes the gateway row and the `files_processed` marker in the same transaction, so a crash mid-file replays cleanly.
- We didn't change `Gateway::insert`'s reliance on the `inserted_at` column's `DEFAULT now()` — a couple of local-env errors hit during testing turned out to be schema-copy drift (missing default; un-dropped `refreshed_at` column). Production schema is unaffected.


bbalser:
## Follow-up: narrow the tracker to antenna/elevation reconciliation only

After cutting over to the change streams, the tracker's only structural job is filling in `antenna`/`elevation` when the metadata DB lagged at stream-event time. This commit narrows it to exactly that — and removes the daily full re-sync that caused the one-time bulk rewrite on first deploy.

### What changed

- **`metadata_db.rs`** — collapsed `MobileHotspotInfo` struct entirely. Now two free functions:
  - `fetch_antenna_and_elevation` (single-row, used by stream daemon at event time)
  - `fetch_antenna_and_elevation_batch` (used by tracker; 500-key batched lookup)
- **`tracker.rs`** — `execute` rewritten:
  - Streams latest row per address from local `gateways`.
  - Batches of 500 → one indexed lookup against `mobile_hotspot_infos` per batch.
  - Writes a new history row only when metadata DB returns a **non-null** value that differs from stored. Never regresses `Some(x) → None`.
- **`db.rs`** — added `Gateway::stream_latest_per_address` (`DISTINCT ON (address) ORDER BY address, inserted_at DESC`).
- **`hotspot_change_stream.rs`** — call site updated to the free-function form.
- **Tests** — replaced the 2 wide-tracker tests with 6 narrow ones: fix-null, update-on-change, no-op-on-match, no-op-on-missing-row, no-op-on-null-meta, partial-field-update.
- **Test helpers** — dropped now-unused `update_gateway` / `insert_asset_owner` / `update_asset_owner`.
- **Flake fix** — replaced wall-clock-vs-DB-time sandwich assertions in `gateway_db.rs` with a `±5m of now` window helper. Same bug-catching power, no flake.

### Behavioral consequences

- No more bulk-rewrite-everything on tracker tick. Only rows where antenna/elevation actually differ get a new history row.
- `MobileHotspotInfo::stream`, `to_gateway`, and the `is_active` / `is_full_hotspot` / `dc_onboarding_fee_paid` fields are gone — they only served the wide tracker path.
- Tracker is now the only metadata-DB writer; stream daemon does its own per-event lookup. Both compute identical narrow hashes since they read from the same source.

### Explicitly given up vs. the old tracker

- Ownership backstop for entities we haven't seen via the hotspot stream (drops stay dropped).
- Stream-missed-hotspot pickup (relies on the proto producer firing at least once per hotspot).
- DR rebuild (would need a separate one-shot tool).

These were nice-to-haves bolted on top of the polling design — accepting their loss in exchange for the tracker doing exactly one thing.

mobile_config/src/cli/server.rs
```diff
@@ -79,8 +86,38 @@ impl Server {
             sub_dao_svc,
         );
 
+        let ingest_bucket = settings.ingest.connect().await;
+
+        let (hotspot_change_rx, hotspot_change_server) =
+            file_source::continuous_source::<MobileHotspotChange, _, _>()
```
michaeldjeffrey: If you like, we can elide the generics as they can be populated when we pass the `change_rx` channel to the corresponding `Daemon`.

mobile_config/src/gateway/db.rs
```diff
@@ -105,6 +105,47 @@ pub struct LocationChangedAtUpdate {
...
+    /// dc_onboarding_fee_paid) so the two paths agree on whether anything
+    /// observable changed.
+    pub fn compute_hash(&self) -> String {
+        let mut hasher = blake3::Hasher::new();
+        hasher.update(
+            self.location
+                .map(|l| l.to_le_bytes())
+                .unwrap_or([0_u8; 8])
+                .as_ref(),
+        );
```
michaeldjeffrey, line 120: Probably not the time to do this, but it we change the position of the unwrap we can get a nicer condensed version of what's happening here.

```
hasher.update(&self.location.unwrap_or_default().to_le_bytes());
```

mobile_config/src/gateway/db.rs
```diff
@@ -147,7 +190,7 @@ impl Gateway {
         Ok(res.rows_affected())
     }
 
-    pub async fn insert(&self, pool: &PgPool) -> anyhow::Result<()> {
+    pub async fn insert<'a>(&self, db: impl PgExecutor<'a>) -> anyhow::Result<()> {
```
michaeldjeffrey:
```suggestion
    pub async fn insert(&self, db: impl PgExecutor<'_>) -> anyhow::Result<()> {
```

mobile_config/src/gateway/hotspot_change_stream.rs
```diff
@@ -0,0 +1,356 @@
...
+        let metadata = change
+            .metadata
+            .ok_or(MobileHotspotChangeError::MissingMetadata)?;
+        let pub_key_bytes = change
+            .pub_key
+            .ok_or(MobileHotspotChangeError::MissingPubKey)?
+            .value;
+
+        let device_type_proto = MobileHotspotDeviceType::try_from(metadata.device_type)
+            .map_err(MobileHotspotChangeError::UnsupportedDeviceType)?;
```
michaeldjeffrey: `file_store_oracles` provides a helper for this.

```rs
let device_type_proto = file_store_oracles::prost_enum(
    metadata.device_type,
    MobileHotspotChangeError::UnsupportedDeviceType,
)?;
```

mobile_config/src/gateway/hotspot_change_stream.rs
```diff
@@ -0,0 +1,356 @@
...
+        self.receiver.recv().await
+    }
+
+    async fn handle(&mut self, file: Self::Item) -> anyhow::Result<()> {
+        self.process_file(file).await
+    }
+
+    async fn on_receiver_closed(&mut self) -> Result<ControlFlow<()>, Self::Error> {
+        Ok(ControlFlow::Break(()))
+    }
```
michaeldjeffrey:
```suggestion
```

This is provided through the default impl of `ChannelConsumer`.

mobile_config/src/gateway/ownership_change_stream.rs
```diff
@@ -0,0 +1,230 @@
...
+    #[error("missing report")]
+    MissingReport,
+    #[error("missing change")]
+    MissingChange,
+    #[error("missing entity pub_key")]
+    MissingEntityPubKey,
+    #[error("missing owner")]
+    MissingOwner,
+    #[error("missing owner wallet")]
+    MissingOwnerWallet,
```
michaeldjeffrey, line 36: There's nothing wrong with this pattern. Just pointing out there's an alternative pattern in `file-store-oracles` thats used in a [few places](https://github.com/helium/oracles/blob/24cc6eb10209bb1d282840203dacde17cbf5cd49/file_store_oracles/src/mobile/mobile_session.rs#L23).

```rs
    #[error("missing field: {0}")]
    MissingField(&'static str)
```

mobile_config/src/settings.rs
```diff
@@ -29,6 +29,14 @@ pub struct Settings {
...
     /// the database for Solana on-chain data
     pub metadata: db_store::Settings,
+    /// S3 bucket holding chain_rewardable_entities change reports
+    /// (mobile_hotspot_change_report and entity_ownership_change_report).
+    pub ingest: file_store::BucketSettings,
+    /// Cold-start timestamp for both change-report pollers. Once
+    /// `files_processed` is populated, that table drives the offset
+    /// and this field is only consulted on a fresh DB.
+    #[serde(default = "default_gateway_stream_start_after")]
+    pub gateway_stream_start_after: DateTime<Utc>,
```
michaeldjeffrey, line 39: The default for `DateTime<Utc>` is already `UNIX_EPOCH`.
I'm fine leaving this as it's more explicit.

mobile_config/tests/integrations/gateway_tracker.rs
```diff
@@ -1,198 +1,186 @@
...
+    )
+    .await?;
+    Ok(())
+}
+
+fn wifi_deployment_info(antenna: u32, elevation: u32) -> String {
+    format!(
+        r#"{{"wifiInfoV0":{{"antenna":{},"elevation":{},"azimuth":0,"mechanicalDownTilt":0,"electricalDownTilt":0}}}}"#,
+        antenna, elevation
+    )
```
michaeldjeffrey:
```suggestion
    serde_json::json!({
        "wifiInfoV0": {
            "antenna": antenna,
            "elevation": elevation,
            "azimuth": 0,
            "mechanicalDownTilt": 0,
            "electricalDownTilt": 0
        }
    })
    .to_string()
```

We could use `serde_json` here to help with the eye-crossing while reading escaped `{{` brackets.

michaeldjeffrey (approved):

macpie (approved):

PR #1208, opened 2026-06-11
Create iceberg stream poller similar to file info poller

bbalser:
## Add an Iceberg table snapshot stream API

Adds a reusable streaming reader to `helium_iceberg` that turns each new Iceberg **snapshot** into an event — the read-side counterpart to the crate's existing writers, and the Iceberg analogue of `file_store`'s `file_info_poller` (which treats each new S3 file as an event). Conceptually it's how Spark Structured Streaming reads an Iceberg table snapshot-by-snapshot.

### Motivation

`helium_iceberg` could write Iceberg tables but had no way to *consume* them as a stream. This lets a service react to table updates with the same ergonomics as the file poller: a `ManagedTask` that emits decoded rows over an mpsc channel, with at-least-once delivery via a transactional watermark.

### What's included

- **Snapshot-as-event poller** (`stream/poller.rs`) — a `task_manager::ManagedTask` with a builder mirroring `file_source::continuous_source()`: `poll_duration`, `queue_size`, `process_name`, `LookbackBehavior` (`StartAfter`/`Max`), `stop_after`/`stop_at_now`/`stop_never`, `idle_timeout`, and `skip_non_append`.
- **Incremental reader** (`stream/reader.rs`) — reads only the rows a snapshot *appended*, by walking its manifest list and keeping entries with `status == Added && snapshot_id == target`, then handing the resulting `FileScanTask`s to iceberg's `ArrowReader`. (iceberg-rust 0.9 has no native append-between-snapshots scan, so this reuses the reader without the full-state planner.)
- **Typed parser** (`stream/parser.rs`) — `IcebergStreamParser<T>` + default `JsonIcebergStreamParser` decoding Arrow `RecordBatch` → `Vec<T>` for any `T: DeserializeOwned` (the inverse of the writer's `records_to_batch`).
- **Watermark state** (`stream/state.rs`) — `IcebergStreamState` / `IcebergStreamStateRecorder` traits with `sqlx::Pool`/`Transaction` impls behind a new optional **`sqlx-postgres`** feature. Progress is a single monotonic `sequence_number` per `(process_name, table_name)`; `into_stream(&mut txn)` co-commits the watermark with the consumer's own work.
- **Expired-snapshot gap detection** (`stream/poller.rs`) — if snapshots are expired before being processed (consumer lag > retention), the poller resumes at the next available snapshot and logs a `WARN` identifying the gap (`watermark`, `next_available_sequence_number`, `skipped_snapshots`), once per watermark.
- **Runnable example** (`examples/stream_heartbeats.rs`) — streams `iceberg.poc.heartbeats`, decoding rows into a typed `Heartbeat` struct and printing one JSON document per event (snapshot metadata, running totals, sample row) on stdout; status to stderr so it pipes to `jq`.

### Design notes / trade-offs

- **Non-append snapshots:** `skip_non_append` (default `true`) skips `Replace`/`Overwrite`/`Delete` snapshots — notably a full-table optimize/compaction (a `Replace`) is skipped rather than re-streaming the whole table, since compaction marks every rewritten file as `Added`. With `false`, a non-append snapshot is a hard error rather than risking duplicate emission.
- **Delivery:** at-least-once. A crash between channel-send and consumer-commit re-emits the snapshot on restart; `record` is idempotent (`ON CONFLICT DO NOTHING`).
- **Catalog load per snapshot:** each emitted snapshot triggers one full `load_table`; there's no batching/queue of pending snapshots across calls (a known, documented divergence from `file_info_poller`'s `file_queue`). Negligible at steady state; only matters for backlog catch-up on snapshot-heavy tables. Deferred.
- **Migration ownership:** the `iceberg_snapshots_processed` table is owned by the *consuming* crate's migrations (same convention as `files_processed`); no consumer is wired up in this PR.


helium_iceberg/src/stream/state.rs
```diff
@@ -0,0 +1,132 @@
+//! Offset/watermark persistence for the Iceberg snapshot stream poller.
+//!
+//! This mirrors `file_store`'s `FileInfoPollerState` / `FileInfoPollerStateRecorder`
+//! split, but the unit of progress is an Iceberg *snapshot* rather than an S3
+//! file. Because an Iceberg table's snapshot log is strictly ordered by
+//! `sequence_number`, a single high-watermark per `(process_name, table_name)`
+//! is sufficient — there is no out-of-order arrival to guard against, so none
+//! of the file poller's `exists`/`clean`/cache machinery is needed here.
+//!
+//! The `sqlx::Pool<Postgres>` / `sqlx::Transaction` impls live behind the
```
michaeldjeffrey: Could we use this opportunity to try and provide the migration with the crate rather than needing to inline it in every project that uses it?

helium_iceberg/src/error.rs
```diff
@@ -18,6 +18,18 @@ pub enum Error {
...
     Writer(String),
+
+    #[error("reader error: {0}")]
+    Reader(String),
+
+    #[error("state error: {0}")]
+    State(String),
+
+    #[error("channel error: {0}")]
+    Channel(String),
```
michaeldjeffrey: If we could get away from `(String)` error for these types, that would be lovely.

helium_iceberg/src/lib.rs
```diff
@@ -5,6 +5,7 @@ mod catalog;
 mod error;
 mod iceberg_table;
 mod settings;
+pub mod stream;
```
michaeldjeffrey: For style, can we separate public modules from private modules, similar to file-store?

helium_iceberg/src/lib.rs
```diff
@@ -22,8 +23,13 @@ pub use table_creator::{
...
 
+pub use stream::{
+    continuous, IcebergStream, IcebergStreamParser, IcebergStreamPollerServer, IcebergStreamState,
+    IcebergStreamStateRecorder, JsonIcebergStreamParser, LookbackBehavior, SnapshotMeta,
+};
+
 // Re-export iceberg types for ergonomic API usage
-pub use iceberg::spec::{NullOrder, PrimitiveType, SortDirection, Transform, Type};
+pub use iceberg::spec::{NullOrder, Operation, PrimitiveType, SortDirection, Transform, Type};
 pub use iceberg::{NamespaceIdent, TableIdent};
```
michaeldjeffrey: These aren't used anywhere.
Can we substitute for a top-level `iceberg` re-export?

helium_iceberg/src/stream/poller.rs
```diff
@@ -0,0 +1,491 @@
...
+pub enum LookbackBehavior {
+    /// Only emit snapshots committed strictly after this instant.
+    StartAfter(DateTime<Utc>),
+    /// Only emit snapshots committed within `now - max_lookback`.
+    Max(std::time::Duration),
+}
+
+impl Default for LookbackBehavior {
+    fn default() -> Self {
+        // From the beginning of the table's history.
```
michaeldjeffrey: I was thinking about this the other night. This comment is kind of a lie. 
We can't go back to the beginning of the tables history, only the oldest snapshot.
If someone has a case where they need the beginning of time, there needs to be a different sort of backfill process.

Not sure what that for here/now though...

helium_iceberg/src/stream/poller.rs
```diff
@@ -0,0 +1,491 @@
...
+    }
+}
+
+impl LookbackBehavior {
+    fn cutoff(&self, now: DateTime<Utc>) -> DateTime<Utc> {
+        match self {
+            LookbackBehavior::StartAfter(start_after) => *start_after,
+            LookbackBehavior::Max(max_lookback) => {
+                now - ChronoDuration::from_std(*max_lookback).unwrap_or(ChronoDuration::MAX)
+            }
```
michaeldjeffrey: Subtracting a `std::time::Duration` from `DateTime<Utc>` is supported.
```
LookbackBehavior::Max(max_lookback) => now - *max_lookback,
```

helium_iceberg/src/stream/poller.rs
```diff
@@ -0,0 +1,491 @@
...
+#[derive(Clone, Builder)]
+#[builder(pattern = "owned")]
+pub struct IcebergStreamPollerConfig<Message, State, Parser> {
+    catalog: Catalog,
+    #[builder(setter(into))]
+    namespace: String,
+    #[builder(setter(into, name = "table"))]
+    table_name: String,
+    state: State,
+    parser: Parser,
```
michaeldjeffrey: Can we explore cleaning up some of these generics?

The parser would be the closest thing to a hot path, but we parse in batches. 
I'm willing to be convinced it's not worth it.

helium_iceberg/src/stream/poller.rs
```diff
@@ -0,0 +1,491 @@
...
+    /// Watermark value we last emitted an expired-gap warning for, so we don't
+    /// re-warn on every poll while sitting at the same watermark.
+    warned_gap_at: Option<i64>,
+}
+
+/// A run of snapshots that were expired before the poller could process them.
+#[derive(Debug, Clone, PartialEq, Eq)]
+struct ExpiredGap {
+    /// The watermark we resumed from.
+    watermark: i64,
```
michaeldjeffrey, line 246: I understand this makes it easier to work with the `iceberg` crate, but the Typist in me is annoyed that these are signed values for values that can never be negative.

helium_iceberg/src/stream/poller.rs
```diff
@@ -0,0 +1,491 @@
...
+{
+    /// Pick the lowest-`sequence_number` snapshot that is past both the
+    /// watermark and the lookback cutoff. Returns the chosen snapshot's
+    /// metadata, or `None` if there's nothing new.
+    fn select_next(&self, table: &Table) -> Result<Option<SnapshotMeta>> {
+        let cutoff = self.config.lookback.cutoff(Utc::now());
+        let watermark = self.latest_sequence_number.unwrap_or(i64::MIN);
+
+        let mut best: Option<SnapshotMeta> = None;
+        for snapshot in table.metadata().snapshots() {
```
michaeldjeffrey, line 312: If we sort by `snapshot.sequence_number()` we should be able to return the first item that passes the watermark and timestamp check.

Probably a micro-optimization at this point though.

helium_iceberg/src/stream/poller.rs
```diff
@@ -0,0 +1,491 @@
...
+        }
+
+        Ok(best)
+    }
+
+    /// Warn (once per watermark) if snapshots were expired before we processed
+    /// them. The stream resumes at the next available snapshot regardless —
+    /// the gapped rows are simply never emitted — so this is observability, not
+    /// a recoverable condition.
+    fn warn_on_expired_gap(&mut self, table: &Table) {
```
michaeldjeffrey, line 337: Should this emit a metric? That's more likely to be noticed than a log message. 
Or we could add a ExpiredGapBehavior enum.

The worry here is we set something up, it goes and goes, we realize we missed data and have to figure out how to backprocess up to a random date that may be lost to time/k8s logs.

helium_iceberg/src/error.rs
bbalser: More structure to the errors added 03b1f11

helium_iceberg/src/lib.rs
bbalser: Addressed a43158e

helium_iceberg/src/lib.rs
bbalser: Address fdce4f4

helium_iceberg/src/stream/poller.rs
bbalser: Address 3e3d0866

helium_iceberg/src/stream/poller.rs
bbalser: Added ccd556d

helium_iceberg/src/stream/poller.rs
bbalser: Addressed ccd556d

helium_iceberg/src/stream/poller.rs
bbalser: good call, removed the Parser trait 87b7de4

helium_iceberg/src/stream/poller.rs
```diff
@@ -0,0 +1,539 @@
...
+    catalog: Catalog,
+    #[builder(setter(into))]
+    namespace: String,
+    #[builder(setter(into, name = "table"))]
+    table_name: String,
+    /// Pre-created SQLite pool for the watermark store. When set, the poller
+    /// migrates and uses it, and [`db_dir`](Self::db_dir) is ignored. When
+    /// unset, the poller opens a db under `db_dir`.
+    #[builder(default, setter(strip_option))]
+    pool: Option<SqlitePool>,
```
macpie, line 110: Should we bundle pool and db_dir in one struct? Seems like we don't need one without the other
ex: 
```rust
pub struct WatermarkStore {
    pool: SqlitePool,
    #[builder(default = "default_db_dir()", setter(into))]
    db_dir: PathBuf,
}
pub struct IcebergStreamPollerConfig<Message>  {
    watermark_store: Option<WatermarkStore>,
    ...
}
```

michaeldjeffrey (approved):

PR #1214, opened 2026-06-30
HIP-149: size data-transfer rewards as residual of hnt_rewards_issued

michaeldjeffrey:
The 3x cap and backstop shift HNT between `hnt_rewards_issued` and `delegation_rewards_issued`, so a fixed 70% data transfer share would over-allocate (over the cap) and under-allocate (under the backstop) the HNT the rewarder is handed. Carry both `hnt_rewards_issued` and `delegation_rewards_issued` through `EpochRewardInfo` (the latter for completeness; the rewarder doesn't distribute it) and split `hnt_rewards_issued` as:

  service providers = flat 24% of total emissions
  data transfer     = hnt_rewards_issued - service providers (residual)

so `service_provider + data_transfer == hnt_rewards_issued` exactly, every epoch. New reward_shares::emissions_split holds the split, with proptests pinning that invariant across the cap / backstop / baseline range.

Cleanup:
- centralize `floor_to_u64` so every reward pool rounds the same way (ToZero)
- drop now-unused `get_scheduled_tokens_for_{data_transfer,service_providers}`

### DIFF
--- a/mobile_config/src/sub_dao_epoch_reward_info.rs
+++ b/mobile_config/src/sub_dao_epoch_reward_info.rs
@@ -11,7 +11,21 @@ pub struct EpochRewardInfo {
     pub epoch_address: String,
     pub sub_dao_address: String,
     pub epoch_period: Range<DateTime<Utc>>,
+    /// Total mobile sub-DAO emissions for the epoch (100%): the HNT this rewarder
+    /// distributes (`hnt_rewards_issued`) plus the delegation rewards already paid
+    /// to veHNT holders on-chain.
     pub epoch_emissions: Decimal,
+    /// The slice of `epoch_emissions` this rewarder is responsible for
+    /// distributing (service providers + data transfer). Under the HIP-149 3×
+    /// cap / backstop the chain shifts HNT between this and the delegation
+    /// rewards, so it is no longer a fixed 94% of `epoch_emissions`; the rewarder
+    /// must distribute exactly this amount — no more, no less.
+    pub hnt_rewards_issued: Decimal,
+    /// The portion of `epoch_emissions` paid to veHNT delegators on-chain
+    /// (`epoch_emissions - hnt_rewards_issued`). The rewarder does not distribute
+    /// this; it is carried so the full on-chain split is available without another
+    /// mobile-config change.
+    pub delegation_rewards_issued: Decimal,
     pub rewards_issued_at: DateTime<Utc>,
 }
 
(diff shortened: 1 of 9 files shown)


bbalser (approved):

madninja (commented):
Reviewed the reward allocator across the HIP-149 stack (#1211 + this PR) against the HIP-149 bucket model and the on-chain cap/floor side that shipped in helium-program-library #1197. The allocator is correct and consistent with the on-chain mechanism, no blockers:

- Residual sizing (`emissions_split.rs` `split`): `service_provider = floor(total * 0.24).min(hnt)`, `data_transfer = hnt - service_provider`. Floor-only can't over-allocate the escrow, the `.min(hnt)` clamp keeps the residual non-negative, and the dropped sub-bone lands in the residual so `service_provider + data_transfer == hnt_rewards_issued` exactly. The data pool shrinks under the cap and grows under the floor, as intended.
- PoC is fully retired from the live path (`reward()` calls only `reward_dc_hip_149` + `reward_service_providers`).
- SP is a flat 24% of total emissions; the delegators' 6% is correctly not paid here (it's paid on-chain).

Two test-quality items worth tightening (not correctness):

1. Proptest `unallocated_never_exceeds_individual_gw_rewards` (`data_transfer.rs` ~425) asserts `unallocated < reward.reward` for every reward, but a zero-DC gateway has `reward == 0`, so `unallocated < 0` is false. It passes only because `any::<u64>()` effectively never samples `dc == 0`. Suggest guarding `dc >= 1` (like the sibling test) or asserting `reward == 0 || unallocated < reward`. Production dust handling is fine (covered by `unallocated_dust_never_exceeds_gateway_count`); this is just false confidence in the test as written.

2. No integration-level cap/floor coverage. Every `reward_dc_hip_149` integration test uses the baseline 6%-delegation `reward_info`, so the residual shrink/grow (the whole point of this PR) is only exercised at the unit layer. An integration test each for a capped epoch (larger `delegation_rewards_issued`, smaller `hnt_rewards_issued`) and a floored epoch (`hnt_rewards_issued` above 94%), asserting data shrinks/grows while SP stays flat and `dc_sum + unallocated == data_transfer`, would lock in the end-to-end wiring.

Minor:

- Data-transfer rounding dust is emitted as `UnallocatedRewardType::Poc` (`rewarder.rs` ~414), a harmless legacy label that now misnames data dust as PoC.
- `cli/reward_from_db.rs` still uses the old 70%-data/PoC helpers and would print HIP-149-inconsistent numbers (`println!` only, so it can't emit on-chain), which also keeps the retired PoC symbols alive despite the "dead code" comments elsewhere.

PR #1218, opened 2026-07-07
Bound iceberg stream poller watermark db with timed pruning

bbalser:
## What

The iceberg stream info poller records its progress in a per-poller SQLite watermark db (`iceberg_snapshots_processed`). It appended one row per processed snapshot and **never removed any**, so the db grew without bound — even though only `MAX(sequence_number)` is ever read back.

This adds a bounded-history maintenance routine:

- **`StreamState::prune_history`** (`state.rs`) — deletes all but the most recent 100 rows per `(process_name, table_name)`, keeping the watermark plus a small window of history for troubleshooting.
- **Timer-driven from the poller run loop** (`poller.rs`) — a new `prune_duration` config (default hourly) adds a `tokio::select!` branch that prunes on a timer, rather than on every `record`. This keeps maintenance off the per-snapshot commit hot path.

## Why

Left unbounded, the watermark db grows one row per snapshot forever. On a long-running poller against a frequently-committing table (30s default poll interval), that table accumulates rows indefinitely, all of them dead weight since only the max is read.

## Notes for reviewers

- The prune branch operates on a **clone of the store** (a ref-counted `SqlitePool` handle) so it doesn't borrow `self` while the poll branch holds it mutably — same pattern already used for `sender`.
- The interval's first tick fires immediately, so a restart also trims any pre-existing bloat on startup, then settles into the hourly cadence.
- A prune failure logs a warning rather than failing the poller — it's housekeeping, not correctness-critical.
- Between prunes the table can grow to ~100 + (snapshots per interval); still bounded and small.
- Rows are deleted but disk isn't reclaimed (no `VACUUM`); negligible at this steady-state size.
- Covered by `prune_history_keeps_recent_window_and_watermark` in `state.rs`.

### DIFF
--- a/helium_iceberg/src/stream/poller.rs
+++ b/helium_iceberg/src/stream/poller.rs
@@ -40,7 +40,7 @@ const DEFAULT_QUEUE_SIZE: usize = 5;
 /// window. Kept well above `poll_duration` since it's pure housekeeping — the
 /// table only accrues rows between prunes, and the first tick fires
 /// immediately so a restart also trims any pre-existing bloat.
-const DEFAULT_PRUNE_DURATION: std::time::Duration = std::time::Duration::from_secs(60 * 60);
+const DEFAULT_PRUNE_DURATION: std::time::Duration = std::time::Duration::from_mins(60);
 
 /// Env var supplying the default watermark-db directory when `db_dir` isn't set
 /// explicitly (and no `pool` is provided).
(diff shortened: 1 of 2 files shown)


helium_iceberg/src/stream/poller.rs
```diff
@@ -36,6 +36,12 @@ use crate::{Error, Result};
 const DEFAULT_POLL_DURATION: std::time::Duration = std::time::Duration::from_secs(30);
 const DEFAULT_QUEUE_SIZE: usize = 5;
 
+/// How often the poller prunes the watermark db down to its retained history
+/// window. Kept well above `poll_duration` since it's pure housekeeping — the
+/// table only accrues rows between prunes, and the first tick fires
+/// immediately so a restart also trims any pre-existing bloat.
+const DEFAULT_PRUNE_DURATION: std::time::Duration = std::time::Duration::from_secs(60 * 60);
```
michaeldjeffrey: `std::time::Duration::from_mins()` was stabilized in `1.91.0`.

michaeldjeffrey (approved):

PR #1222, opened 2026-07-14
Remove `mobile-config` from `mobie-packet-verifier`

michaeldjeffrey:
- Remove unused `CarrierService` from `mobile-config`
- Use trino for querying mobile hotspot existence
  - fronted by a `retainer::Cache` similar to the `GatewayClient` provided by `mobile-config`
- All tests go through instead of mocking


bbalser:
Currently this will do N queries per hour against trino, we could just do 1 query, pull into sqlite (or memory) and query there? thoughts?

mobile_packet_verifier/src/lib.rs
```diff
@@ -29,52 +27,111 @@ pub fn dc_to_bytes(dcs: u64) -> u64 {
...
+            is_known: bool,
         }
+
+        let stmt = trino_client::Statement::new(format!(
+            r#"
+            SELECT EXISTS (
+                SELECT 1
+                FROM {}
+                WHERE pub_key = :address
+                  AND received_timestamp <= :timestamp
```
bbalser: not sure the received_timestamp part of this query makes sense, that is received_timestamp of the last record ingested for that hotspot

michaeldjeffrey: The original call looked at the `inserted_at` column.
https://github.com/helium/oracles/blob/main/mobile_config/src/gateway/db.rs#L318

But we don't have that in iceberg. Not totally sure how to reconcile that right now.

mobile_packet_verifier/src/gateway.rs
```diff
@@ -0,0 +1,268 @@
...
+/// inventory queried by [`GatewayResolver::is_gateway_known`].
+pub const MOBILE_HOTSPOT_INVENTORY_TABLE: &str = "network.chain.mobile_hotspot_inventory";
+
+/// How often the fallback cache sweeps expired entries.
+const CACHE_EVICTION_FREQUENCY: Duration = Duration::from_secs(60 * 60);
+
+/// How soon to retry the inventory snapshot load after a failure, rather than
+/// waiting the full `refresh_interval`. Keeps a Trino outage from leaving the
+/// snapshot degraded for up to an hour.
+const GATEWAY_REFRESH_RETRY_INTERVAL: Duration = Duration::from_secs(60);
```
bbalser, line 29: is this supposed to be from_mins?

mobile_packet_verifier/src/gateway.rs
```diff
@@ -0,0 +1,268 @@
...
+    known_gateways: Arc<RwLock<HashSet<PublicKeyBinary>>>,
+    refresh_interval: Duration,
+    last_refresh: Option<Instant>,
+}
+
+impl GatewaySnapshotRefresher {
+    // Periodic tasks don't allow for changing schedules. So we check at the
+    // minimum rate and early return when there's nothing to do. Should we add a
+    // ScheduledTask?
+    async fn should_run(&self) -> bool {
```
bbalser, line 205: I don't follow why we need this check as opposed to just always refreshing on tick

mobile_packet_verifier/src/gateway.rs
michaeldjeffrey: For historical purposes...

If we cannot reach trino for some reason, we don't want to wait for another hour to retry.
In a failure case, we will attempt to refresh from trino every minute.

mobile_packet_verifier/src/gateway.rs
michaeldjeffrey: For historical purposes...

Task Managers `PeriodicTask` does not have a way to change the amount of time to wait for the next trigger.
To support the failure case of not being able to reach trino and not wanting to wait another hour for a chance, we trigger the task on the retry interval and determine if it's been an hour since the last refresh.

bbalser (approved):

PR #1230, opened 2026-08-04
Fix timestamp-aware gateway resolution in snapshots and caches

### LINKED ISSUE #1229: Gateway snapshots and fallback caches ignore timestamp-dependent onboarding checks
## Summary

The Trino-backed gateway resolvers in:

- `mobile_packet_verifier/src/gateway.rs`
- `mobile_verifier/src/gateway.rs`

do not consistently enforce the timestamp-dependent gateway onboarding check.

Both per-gateway Trino queries contain:

```sql
AND inserted_at <= :timestamp
```

This means gateway existence is intended to be evaluated relative to the report timestamp.

However, both resolvers have an in-memory snapshot fast path that does not retain or check inserted_at.

In mobile_packet_verifier:

```rust
if self.known_gateways.read().unwrap().contains(public_key) {
    return true;
}
```
In mobile_verifier:

```rust

if let Some(meta) = self.known_gateways.read().unwrap().get(public_key).copied() {
    return Ok(Some(meta));
}

```

The snapshot contains only the current gateway set or current GatewayMeta. Once a gateway is present in the snapshot, the supplied query timestamp is ignored.

The timestamp is part of report validation:
- data transfer sessions use received_timestamp;
- heartbeats use the heartbeat's timestamp;
- speedtests use received_timestamp.

## Reproduction

Given a gateway with:
`inserted_at = 2026-08-01 10:00 UTC`

and a historical report with:

`gateway query timestamp = 2026-07-31 10:00 UTC`

the expected result is that the gateway is not found because it had not yet been onboarded.

If the resolver's startup or refreshed snapshot already contains the gateway, the snapshot fast path returns it immediately, bypassing the `inserted_at <= query timestamp` condition.

## Fallback cache issue

The fallback caches have the same correctness problem.

They are keyed only by public key:

`Cache<PublicKeyBinary, bool>`

and:

`Cache<PublicKeyBinary, Option<GatewayMeta>>`

However, the cached query result depends on both:

`(public_key, query_timestamp)`

This makes results dependent on lookup order:

1. A pre-onboarding query caches false/None; a later post-onboarding query is incorrectly rejected.
2. A post-onboarding query caches true/metadata; a later pre-onboarding query is incorrectly accepted.

The default `gateway_refresh_interval`, which is also used as the fallback cache TTL, is one hour.

### mobile_packet_verifier

- Data transfer reports predating gateway onboarding can be accepted when the gateway is in the snapshot.
- Valid post-onboarding reports can be rejected after a pre-onboarding fallback result has been cached.

### mobile_verifier

- Heartbeats and speedtests predating gateway onboarding can be resolved using a gateway from the current snapshot.
- At minimum, this breaks the onboarding-time boundary and can apply current gateway metadata to a report from before the gateway existed.

This issue is specifically about the inserted_at onboarding cutoff. Full historical versioning of device type or asserted location would be a separate concern.

## Proposed fix

Store inserted_at together with the snapshot data.

For mobile_packet_verifier, the snapshot could be:

```rust
HashMap<PublicKeyBinary, DateTime<Utc>>
```

For mobile_verifier, it could use:

```rust
struct GatewaySnapshotEntry {
    inserted_at: DateTime<Utc>,
    meta: GatewayMeta,
}
```

A snapshot hit should only resolve the gateway when:
`entry.inserted_at <= *gateway_query_timestamp`

The fallback cache should cache timestamp-independent source data rather than the final timestamp-dependent result:

`Cache<PublicKeyBinary, Option<GatewaySnapshotEntry>>`

The fallback query would fetch the gateway and its inserted_at without applying the report timestamp cutoff. Each lookup would then evaluate the cutoff against the supplied timestamp.

## Tests

Please add coverage for both resolvers.
Snapshot path:

```shell
query_time < inserted_at  => not found
query_time >= inserted_at => found
```

Fallback cache path:

```shell
pre-onboarding query followed by post-onboarding query
post-onboarding query followed by pre-onboarding query
```

Both lookup orders must return distinct and correct results.

## Context
Timestamp semantics were discussed during PR #1222. The fallback query retained the original inserted_at cutoff, but the snapshot and cache paths do not preserve that behavior: https://github.com/helium/oracles/pull/1222

grandpig:
## Summary

Fix gateway resolution so the `inserted_at` onboarding boundary is consistently enforced by both the snapshot and fallback-cache paths.

Closes https://github.com/helium/oracles/issues/1229

## Problem

The per-gateway Trino queries evaluate gateway existence using:

```sql
inserted_at <= :timestamp
```

but the in-memory snapshots do not retain inserted_at, so snapshot hits ignore the supplied query timestamp.

Fallback results are also cached only by public key even though the result depends on both public key and timestamp. This can cause a lookup for one report timestamp to incorrectly determine the result for another timestamp.

## Changes

- Include inserted_at in gateway snapshot entries.
- Check inserted_at against the supplied query timestamp on every snapshot lookup.
- Cache timestamp-independent gateway entries instead of final boolean/resolution results.
- Apply the timestamp cutoff after reading from the fallback cache.
- Preserve the existing fail-closed behavior for Trino query failures.

## Tests

Added regression coverage for:
- snapshot lookup before onboarding;
- snapshot lookup at or after onboarding;
- fallback lookup before onboarding followed by one after onboarding;
- fallback lookup after onboarding followed by one before onboarding.

These cases are covered for both mobile_packet_verifier and mobile_verifier.

### DIFF
--- a/mobile_packet_verifier/src/gateway.rs
+++ b/mobile_packet_verifier/src/gateway.rs
@@ -6,12 +6,12 @@
 //! per-pubkey Trino fallback.
 
 use std::{
-    collections::HashSet,
+    collections::HashMap,
     sync::{Arc, RwLock},
     time::{Duration, Instant},
 };
 
-use chrono::{DateTime, Utc};
+use chrono::{DateTime, FixedOffset, Utc};
 use helium_crypto::PublicKeyBinary;
 use retainer::Cache;
 use task_manager::Periodic;
@@ -39,8 +39,8 @@ const GATEWAY_REFRESH_RETRY_INTERVAL: Duration = Duration::from_secs(60);
 pub struct GatewayResolver {
     trino_client: trino_client::Client,
     inventory_table: String,
-    known_gateways: Arc<RwLock<HashSet<PublicKeyBinary>>>,
-    fallback_cache: Arc<Cache<PublicKeyBinary, bool>>,
+    known_gateways: Arc<RwLock<HashMap<PublicKeyBinary, GatewayEntry>>>,
+    fallback_cache: Arc<Cache<PublicKeyBinary, Option<GatewayEntry>>>,
     refresh_interval: Duration,
     /// When the startup snapshot was loaded (`None` if the load failed). Seeds
     /// the refresher's first-refresh timing via [`GatewayResolver::refresher`].
@@ -85,7 +85,7 @@ impl GatewayResolver {
                         ?err,
                         "failed to load gateway inventory at startup; starting empty, will retry"
                     );
-                    (HashSet::new(), None)
+                    (HashMap::new(), None)
                 }
             };
         let known_gateways = Arc::new(RwLock::new(initial));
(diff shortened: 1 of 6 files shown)


grandpig:
@michaeldjeffrey Just following up on this in case it got buried. The main semantic change is to retain each gateway's inserted_at in both the snapshot and fallback cache, so resolution remains timestamp-aware regardless of cache order. The regression tests cover both before→after and after→before lookup orderings. Would appreciate a look when you have time. Thanks!

bbalser (approved):

michaeldjeffrey (approved):

PR #1238, opened 2026-08-18
Stop writing heartbeats/speedtests/bans to postgres

michaeldjeffrey:
This is quite the overhaul of the reward pipeline, here we go.

- Settings changes:
  - `iceberg_settings`, REQUIRED
  - `rewarder_state_file` for meta information transition, REQUIRED
  - `location_cache_refresh_interval` (default 1 hour)

## Wifi Heartbeats, Speedtests, and Bans no longer written to postgres

Being no longer written to postgres, they are no longer "cleared" at the end of a rewards cycle. They are still being written to s3 and iceberg.

The speedtest window for calulating averages is now warmed from trino.

These changes don't touch the rewarding pipeline, but until we split apart ingesting and rewarding, they live here.

The tables are still in place, and will be removed with a future migration.

## Iceberg is hard required

NOTE: Trino has been made a requirement, if the LocationCache cannot warm on boot, it is treated as a hard failure so we don't mark heartbeats with a bad location trust score.

## Rewarder state `meta` table

In an attempt to further reduce the dependency on postgres, reward meta information is being persisted to a file on disk, this will later become the source of truth for:
  - `next_reward_epoch`
  - `disable_complete_data_checks_until`

## More POC code removal

Unused code from `SpeedtestAverages`, removed.
Unused db code from `banning.rs`, removed.

## We're still ingesting Bans

Just thought it should be noted.


mobile_verifier/src/speedtests.rs
```diff
@@ -289,90 +276,287 @@ where
...
+/// anymore. This keeps the same window the old query produced — up to
+/// `SPEEDTEST_AVG_MAX_DATA_POINTS` samples within [`SPEEDTEST_LAPSE`] hours —
+/// per hotspot instead, warmed once at startup from `poc.speedtests` in Trino so
+/// a restart doesn't reset every hotspot to a single-sample average.
+///
+/// Entries expire on their own after [`SPEEDTEST_LAPSE`] hours of inactivity, so
+/// hotspots that stop reporting fall out without an explicit purge.
+#[derive(Clone)]
+pub struct RecentSpeedtests {
+    windows: Arc<Cache<PublicKeyBinary, Vec<Speedtest>>>,
```
bbalser, line 292: This data structure will store the PublicKeyBinary for the hotspot up to 7 times, could be a decent chunk of memory.  Could we restructure to only store once?

bbalser (approved):

PR #1243, opened 2026-08-26
HIP-150: ingest data transfer multiplier tickets in the packet verifier

michaeldjeffrey:
This is only the ingestion of multipliers in mobile-packet-verifier, applying them to the burn process is coming in the next PR.

---- 

Reads ticket files from s3, rules on each ticket, and writes the verdict to
a verified report and to data_transfer.multiplier_ticket_history.

No behaviour change: nothing applies a multiplier yet. The burn path is
untouched, and get_multipliers has no caller outside tests. A ticket can be
issued, verified and recorded, and it affects nothing until the burn is
wired up in a following change.

Ruling here rather than at ingest means a refused ticket is recorded: every
ticket produces a verified report and a history row, so the record shows why
a hotspot is not multiplied as well as why it is.

Nothing is stored in Postgres. The history table is append-only, and current
state is a dbt model over it — the same shape as enabled_carriers_inventory
over enabled_carriers_history — so no service owns a mutable copy.

Notes:

- The burn will read the history table rather than the inventory mart. It
  needs the multiplier in force at the timestamp of the file being
  processed, which merged current state cannot answer, and reading the log
  this service wrote keeps burning off the dbt cadence.
- Latest-per-hotspot is ordered on the issuer's signed timestamp, not on
  arrival. A ticket is a correctly signed message that never expires, so a
  captured one can be resubmitted after the grant it carries was revoked;
  ordered by arrival it would win. Ordered by signed timestamp it sorts below
  the ticket that superseded it, which a replay cannot change without the
  signing key.
- received_timestamp bounds which files a ticket applies to, so a ticket
  never applies retroactively.
- Signer and freshness are checked again even though ingest checked them.
  The two services are configured separately, and this is the copy that
  decides whether a hotspot's rewards are multiplied. Freshness is measured
  against when ingest received the ticket, so replaying a backlog of files
  does not reject every ticket in it.
- The multiplier is stored as decimal(9,6), never a float: values are
  negotiated per venue and are not always binary-representable.
  IcebergDecimal bridges serde (the write path) and the Trino trait (the
  read path), which no existing type spans.
- Ticket signers may be empty and default to empty, matching ingest.

### DIFF
--- a/mobile_packet_verifier/src/multiplier/ingestor.rs
+++ b/mobile_packet_verifier/src/multiplier/ingestor.rs
@@ -1,0 +1,218 @@
+//! Reads ticket files from s3, rules on each ticket, records the verdict.
+//!
+//! Every ticket produces a verified report — accepted or rejected — so the
+//! public record shows refusals as well as grants. Both also land in the
+//! append-only history table, tagged with the verdict.
+//!
+//! Nothing here holds mutable state: this module only appends. What is
+//! *currently* in force is derived from those rows separately, by
+//! [`super::inventory`].
+
+use std::{ops::ControlFlow, time::Duration};
+
+use chrono::Utc;
+use file_store::file_info_poller::FileInfoStream;
+use file_store_oracles::mobile::data_transfer_multiplier::{
+    proto::VerifiedDataTransferMultiplierTicketReportV1, DataTransferMultiplier,
+    DataTransferMultiplierTicketReport, VerifiedDataTransferMultiplierTicketReport,
+    VerifiedDataTransferMultiplierTicketStatus as Status, MAX_CLOCK_DRIFT,
+};
+use futures::StreamExt;
+use sqlx::PgPool;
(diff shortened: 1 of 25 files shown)


mobile_packet_verifier/src/multiplier/ingestor.rs
```diff
@@ -0,0 +1,212 @@
...
+///
+/// A free function rather than a method: this is the rule that decides whether a
+/// hotspot's rewards get multiplied, and it should be testable without a
+/// channel, a file poller or a sink.
+///
+/// The multiplier itself is not checked here. It was parsed and range-checked on
+/// the way out of the proto, so a report that exists at all cannot carry an
+/// out-of-range value — what is left is who signed it, when, and whether the
+/// hotspot is real.
+pub async fn ticket_status(
```
bbalser, line 163: shouldn't this validate the mutliplier is between 1 and 5?

PR #1245, opened 2026-08-28
HIP-150

michaeldjeffrey:
In draft until the last PR is merged into this tracking branch.
https://github.com/helium/oracles/pull/1244

---

- [Proto changes](#org112c447)
- [Service Provider Allocation](#org5ae8dc1)
- [Storing multipliers](#orgb64738d)
- [Applying multipliers](#org258265e)
- [Postgres Migrations](#orgd63eae0)
- [Iceberg Migrations](#org9ca57d7)
- [Iceberg multipliers table](#org8f0345e)
- [New Settings](#org8adea2e)
- [New Metrics](#org849ed98)

<a id="org112c447"></a>

# Proto changes

Still need to be merged before this branch is merged to main.

<https://github.com/helium/proto/pull/483>

<a id="org5ae8dc1"></a>

# Service Provider Allocation

The Service Provider allocation is being contributed to the data pool.

<a id="orgb64738d"></a>

# Storing multipliers

The history is going to trino, along with s3. The inventory table is stored in postgres. Until we figure out a more effecient way to cleanup with inventory table or move more things into trino, that table has no maintenance process.

<a id="org258265e"></a>

# Applying multipliers

Data Transfer Sessions are stored indexed by `(pub_key, payer, file_timestamp)`. This allows us to make sure we apply the multiplier that was active for a given gateway at the time we processed the file.

Sessions are collected, and per HIP-150, the multiplier is applied to "data credits derived from that Hotspot's rewardable bytes".

Pay special attention to `group_by_multiplier()`, it's the mechanism that let's us avoid paying the `bytes_to_dc()` penalty per session when there are rewardable bytes right around the single DC threshold. This was previously achieved by accumulating by pubkey into a single row.

Think 10 sessions with 10,000 bytes with 1.5x multiplier. Applied per session would come to 10 DC becuause of \`ceil\`. Grouped, is 1 DC.

<a id="orgd63eae0"></a>

# Postgres Migrations

Three migrations run on start:

-   **`10_data_transfer_multipliers.sql`:** new table
-   **`11_data_transfer_session_buckets.sql`:** `data_transfer_sessions p_key -> (pub_key, payer, last_timestamp)`
-   **`12_pending_session_multiplier.sql`:** `pending_data_transfer_sessions`: multiplier column, same key change

Both key changes widen an existing key so they cannot collide, but they take an `ACCESS EXCLUSIVE` lock and rebuild the index. Deploy just after a burn, when `data_transfer_sessions` is smallest.

Rolling the binary back after migration 11 will not work. The previous release's `ON CONFLICT (pub_key, payer)` has no matching unique constraint once the key is widened, and every accumulate fails. Roll forward.

<a id="org9ca57d7"></a>

# Iceberg Migrations

Adding multipliers to BurnedDataTransferSessions in Iceberg.

If we don't run this query in iceberg, the multiplier fields will be ignored while we're writing new data transfer sessions.

```sql
ALTER TABLE data_transfer.burned_sessions ADD COLUMN multiplier DECIMAL(9, 6)
```

<a id="org8f0345e"></a>

# Iceberg multipliers table

Currently, we're choosing partition this table by buckets on `hotspot_pubkey`.

Given what the HIP says about locations that are likely to obtain data transfer multipliers, ~<10%. That means we will have a sparse table where partitioning by day spreads out the little information we have into many small files.

<a id="org8adea2e"></a>

# New Settings

```sh
MPV__MULTIPLIER__INPUT_BUCKET__BUCKET # Where to find data_transfer_multiplier ticket reports.
MPV__MULTIPLIER__START_AFTER          # Start ingesting from...
MPV__MULTIPLIER__AUTHORIZED_KEYS      # Comma separated list of b58 keys that can sign multiplier tickets
MPV__MULTIPLIER__TICKET_MAX_AGE       # How long are tickets valid after being signed?
```

<a id="org849ed98"></a>

# New Metrics

```
burned_dc_by_multiplier{payer="", multiplier=""}
```


file_store_oracles/src/mobile/data_transfer_multiplier.rs
```diff
@@ -0,0 +1,614 @@
...
+    /// this, and an absent multiplier on a *burned session* resolves to it.
+    ///
+    /// An absent multiplier on a *ticket* is a different thing — a grant that
+    /// grants nothing — and is refused rather than defaulted. See
+    /// `mobile_packet_verifier::multiplier::ingestor::ticket_status`.
+    pub const DEFAULT: Self = Self(Decimal::ONE);
+
+    /// Validate and normalize a multiplier.
+    ///
+    /// Normalizing means `1.5`, `1.50` and `1.5e0` all become the same value, so
```
macpie, line 149:
```suggestion
    /// Normalizing means `1.5`, `1.50` and `1.5e0` all become the same value (up to `MAX_SCALE`), so
```

macpie (approved):

PR #1248, opened 2026-09-09
Write ingest output to multiple buckets

macpie:
Ingest can now write every rolled file to more than one bucket: byte for byte, under the same key, to an S3 primary and a Cloudflare R2 mirror that share no credentials, region or endpoint.

Along the way the upload path was reworked enough that it's worth describing the end state rather than the diff.

## Config

Every service that uploads now uses one settings type, `file_upload::Settings`:

```toml
[file_upload]
root = "/opt/ingest/data"

[file_upload.buckets.primary]
bucket = "ingest-bucket"
region = "us-west-2"

[file_upload.buckets.r2]
bucket = "ingest-mirror"
endpoint = "https://<account_id>.r2.cloudflarestorage.com"
region = "auto"
access_key_id = "<r2 access key id>"
secret_access_key = "<r2 secret access key>"
```

One bucket or several — the sinks can't tell the difference. Each entry carries its own connection settings and inherits nothing from the others, which is what makes a second provider possible: lending an R2 bucket the S3 account's region or key pair only builds a client that authenticates against neither.

**Buckets are keyed by a label, not by bucket name.** Bucket names are hyphenated and hyphens can't appear in an environment variable, so keying on them would put `INGEST__FILE_UPLOAD__BUCKETS__R2__SECRET_ACCESS_KEY` out of reach — and a mirror's credentials belong in the environment, not the settings file. A `BTreeMap`, so connect/log/report order is stable across runs.

`validate()` runs before anything connects and rejects an empty map, a blank bucket name, and an entry with only one half of a credential pair — the S3 client installs static credentials only when it has both, and otherwise silently falls back to the ambient AWS chain, which can't authenticate against R2. That's a typo that would otherwise pass startup and fail every upload.

### Breaking config changes

- **`cache` is gone** from ingest, mobile-verifier and mobile-packet-verifier — it only ever fed the sinks and the uploader, and is now `file_upload.root`. **price keeps it**: it also roots the iceberg spool and the latest-price file, which aren't uploads.
- **`root` has no default.** Each service's `default_cache()` differed, so a shared default would silently point somewhere other than where a deployment's files already are. Every deployment must set it.
- ingest's `output_bucket` / `output_bucket_mirror` / `file_store` / `file_store_mirror` are replaced by the block above; mobile-verifier's `buckets.output`, mobile-packet-verifier's `output_bucket` and price's `output` likewise.

## How a file reaches two buckets

No bytes are ever copied — it's rename, hardlink and unlink throughout.

```
cache/tmp/report.gz            sink writes here
   ↓ rename                    sink rolls, deposits
cache/report.gz                nlink=1
   ↓ hard_link ×N              one link per bucket, no bytes copied
cache/ingest-bucket/report.gz  ┐ nlink=3, one inode
cache/ingest-mirror/report.gz  ┘
   ↓ remove cache/report.gz    caller's link released → nlink=2
```

Each bucket then owns its own link: it uploads on its own schedule and unlinks when done, and the bytes survive until the last one finishes. So **N buckets cost one copy of the bytes on disk**, and a failing bucket can't delete a file another still needs, can't hold up another's uploads, and can't lose its own — whatever it hasn't stored is re-queued from its own directory on the next start.

A single bucket takes the identical path with one link, so there's no separate "uploads in place" mode to reason about.

## Shape

`FileUpload` is what a sink holds — one `InnerFileUpload` per bucket behind it. `Settings::connect()` returns the uploader plus a nested `TaskManager` (it implements `ManagedTask`), so callers add one task however many buckets are configured:

```rust
let (file_upload, file_upload_tasks) = settings.file_upload.connect().await?;

TaskManager::builder()
    .add_task(file_upload_tasks)
    .add_task(wifi_heartbeat_report_sink_server)
```

Registered before the sinks, so LIFO shutdown still stops the uploaders *after* them — a sink's last files are handed over before its uploader goes. Tasks are named `file_upload:<bucket>` rather than taking `add_task`'s `type_name`, which would log every bucket under one identical string.

## Observability

`file_store_upload{bucket,status}` — one increment per file per bucket at its terminal outcome, so `status="error"` counts files that bucket actually dropped:

```
rate(file_store_upload{status="error"}[5m]) > 0
```

Both series are seeded to zero per bucket at startup, so an alert can tell "no failures" from "not reporting at all". Without it a bucket that quietly stops accepting files is invisible: the uploader keeps its copy, logs, and the service looks healthy.

## Two fixes that fell out

**A `roll_time` under 60s was silently ignored.** Time-based rolling happens only on a rollover timer whose period was a flat 60s; the write path checks size alone. The interval is now `min(roll_time, 60s)`. Production is unchanged (3 min default, 15 for ingest), and `file_store`'s `#[cfg(test)]` override of the constant is gone. `mobile_packet_verifier`'s tests set `roll_time(100ms)` with a comment claiming it "ensures the file is committed promptly" — that only became true with this change.

**`AwsLocal::delete_bucket` couldn't delete an empty bucket.** It always sent `DeleteObjects`, which S3 rejects with no keys (`MalformedXML`), so a bucket created but never written to couldn't be torn down.

## Tests

End to end against the RustFS in `docker-compose.yml`, on buckets `AwsLocal` creates and deletes per run — no `init.sh` changes, no cross-run collisions.

- `ingest/tests/multi_output_bucket.rs` drives the real `grpc_server` from a real settings file through the real sinks and uploader, asserting both buckets hold the same key with identical bytes that decode back to the submitted report. Unlike `mobile_ingest.rs`, nothing is stubbed.
- `file_store/tests/file_upload.rs` — 8 tests: the fan-out, `hardlinks_into_each_bucket_rather_than_copying` (asserts one inode, `nlink == 3`), `a_single_bucket_stages_in_its_own_directory`, `a_failing_bucket_keeps_only_its_own_copy`, `resumes_files_left_in_its_directory`, the two config guards, and the metric's label shape.
- `ingest/tests/settings_env.rs` pins that a bucket's credentials can be injected purely from the environment — the reason buckets are keyed by label.
- `ingest` unit tests cover the config shape and that a settings dump omits bucket credentials (`main` logs the whole struct as JSON at startup, and `skip_serializing` now sits behind a `serde(flatten)`).

Each assertion was mutation-checked to confirm it fails when the behaviour breaks.


file_store/src/file_upload.rs
```diff
@@ -32,13 +33,62 @@ pub struct FileUpload {
...
+
+    /// Uploads every file to `bucket` and, byte for byte and under the same
+    /// key, to each of `additional_buckets`.
+    ///
+    /// A local file is only deleted once every bucket has it, so a bucket that
+    /// is failing never costs the others their copy. Nothing re-queues a file
+    /// whose upload was abandoned, so a file left behind stays in the cache
+    /// directory until the process restarts — [`crate::file_sink::FileSink`]
+    /// rescans that directory on startup and queues what it finds.
+    pub async fn with_additional_buckets(
```
macpie: This does not need to be async

ingest/src/settings.rs
```diff
@@ -146,6 +168,94 @@ impl Settings {
...
+    /// Builds the uploader every file sink writes through, plus a server task
+    /// per output bucket for the caller to manage.
+    ///
+    /// One [`FileUpload`] per bucket, gathered behind a [`MultiFileUpload`].
+    /// Each bucket uploads from its own directory under `cache`, so a rolled
+    /// file is hardlinked once per bucket rather than copied, and the buckets
+    /// make progress independently: a mirror that is failing cannot delete the
+    /// primary's copy, hold up its uploads, or lose its own — whatever it has
+    /// not stored is picked up again on the next start.
+    pub async fn file_uploaders(&self) -> anyhow::Result<(MultiFileUpload, Vec<FileUploadServer>)> {
```
bbalser, line 180: rather than returning a Vec<FileUploadServer>, should this function setup a TaskManager and add the FileUploadServers to that TaskManager and return it?

ingest/src/settings.rs
```diff
@@ -38,7 +42,25 @@ pub struct Settings {
...
+    /// credentials.
+    ///
+    /// The mirror inherits nothing from `[file_store]`, which is what lets it
+    /// live on another provider — an R2 bucket alongside an S3 primary. Lending
+    /// it the S3 account's region or key pair would only build a client that
+    /// authenticates against neither.
+    ///
+    /// Each bucket uploads from its own directory under `cache` and so makes
+    /// progress independently: neither can hold up or delete the other's copy.
+    pub output_bucket_mirror: Option<String>,
```
bbalser, line 60: should we centralize FileUpload settings into a new struct in file upload and use BucketSettings??

file_store/src/file_upload.rs
macpie: they do now as in I`nnerFileUpload:new/2` we `fs::create_dir_all(&dir).await?;`

bbalser (approved):

file_store/src/file_sink.rs
```diff
@@ -20,10 +20,14 @@ use tokio::{
...
-#[cfg(test)]
-pub const SINK_CHECK_MILLIS: u64 = 50;
+/// Longest a sink waits between checks for a file that is due to roll. The
+/// check interval is the configured roll time capped at this, so a sink told to
+/// roll every few seconds actually does — before this was a flat 60s, a
+/// `roll_time` shorter than the interval was silently rounded up to it, and the
+/// tests had to override the constant under `cfg(test)` to observe a roll at
+/// all. At or above 60s (every deployment: the default roll time is 3 minutes,
+/// ingest configures 15) the interval is unchanged.
+pub const MAX_SINK_CHECK_MILLIS: u64 = 60_000;
```
michaeldjeffrey, line 30: This PR introduces some other `Duration`s as `const`. 

Maybe we do that here as well.

file_store/Cargo.toml
```diff
@@ -37,6 +37,10 @@ aws-sdk-s3 = { workspace = true }
 aws-smithy-types-convert = { workspace = true }
 
 [dev-dependencies]
+metrics-util = { version = "0", features = ["debugging"] }
+tempfile = { workspace = true }
+tokio = { workspace = true }
+triggered = { workspace = true }
```
michaeldjeffrey, line 43: These 3 deps already exist.
They're good here, just wanted to point out.

michaeldjeffrey (approved):

macpie:
This PR is code ready but as we do not need this functionality right we do not want to merge and have to update config for all the Oracles services

